PR #38, opened 2021-01-13
feat: enable selections with mouse click

### LINKED ISSUE #33: Send selections to engine and update data
By clicking dimension cells that value should be selected. When selections are confirmed, the data should be updated. when selections are cleared, all rows should be returned from engine. When selections are active, only selections in that column should be possible. If selections are made on other than the first page, should go back to first pave

haxxmaxx:
fixes #33 
Think I got most of it covered. See some comments below.
The test decrease is a bit unfortunate, but will probably fluctuate when adding spec files at this stage.


.eslintrc.json
```diff
@@ -15,6 +15,7 @@
     "no-bitwise": 0,
     "prettier/prettier": 2,
     "no-unused-expressions": 0,
+    "no-nested-ternary": 0,
```
haxxmaxx, line 18: I'd say this is rather clean if used correctly

src/selections-factory.js
```diff
@@ -0,0 +1,101 @@
+import { useEffect, useState, useSelections } from '@nebula.js/stardust';
+
+export function selectCell(selections, cell) {
```
haxxmaxx, line 3: getting `Cognitive Complexity of 10 (exceeds 5 allowed)` here. Sure, there is a lot of logic in this func. Could split this up, but not sure it adds any value? opinions?

src/__test__/selections-factory.spec.js
```diff
@@ -0,0 +1,126 @@
+import { getCellStyle, selectCell } from '../selections-factory';
+
+describe('selections-factory', () => {
+  // describe('initSelections', () => {
```
haxxmaxx, line 4: I can't run this since the nebula hooks running in `initSelections` needs a visualization component. Worth it to mock one just to test? Would be one of those tests that look good but doesn't really test much.

src/selections-factory.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  setSelected(newSelected);
+}
+
+export default function initSelections() {
+  const api = useSelections();
+  const [selections] = useState({
+    api,
+    getCellStyle: (cell, isDim) => getCellStyle(selections.selected, cell, isDim),
```
haxxmaxx, line 53: Tried messing with `this` here, but couldn't get it to work + this is easier to test IMO

src/selections-factory.js
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    if (!selections.api) {
+      return () => {};
+    }
+
+    selections.api = api;
+    selections.api.on('deactivated', resetSelections);
+    selections.api.on('canceled', resetSelections);
+    selections.api.on('confirmed', resetSelections);
+    selections.api.on('cleared', resetSelections);
```
haxxmaxx, line 74: one call here would be awsm, don't think it is possible.

src/selections-factory.js
haxxmaxx: I changed the rules so 8 is fine. But i think we shouldn't get more complex than this.

src/selections-factory.js
Caele: nope, doesn't seem like it.
I mean you could
`types.forEach(t=>selections.api.on(t,resetSelections)`
but that just makes it shorter and less readable.

src/selections-factory.js
Caele: think this is fine

src/selections-factory.js
```diff
@@ -0,0 +1,85 @@
...
+
+export function selectCell(selections, cell) {
+  const { api, selected, setSelected } = selections;
+  const { rowIdx, colIdx, qElemNumber } = cell;
+  let newSelected = [];
+
+  // prevent selection in other column
+  if ((selected.length && selected[0].colIdx !== colIdx) || !api) {
+    return;
+  }
```
Caele, line 18: Is this done in the ui as well? I mean after you clicked in one column, we should not listen to events in the other?
I guess it can be something like click outside confirms, but still think you should need two clicks to start selecting in the next column.

src/index.js
```diff
@@ -22,6 +23,8 @@ export default function supernova(env) {
       const [pageInfo, setPageInfo] = useState({ top: 0, height: 100 });
       const [tableData, setTableData] = useState();
 
+      const selections = initSelections();
+
```
Caele, line 27: Can you make selections dependant on options? So that it can be disabled from the getgo. Both for edit mode cases but also snapshots and directly from a mashup emebd.
Think it is useOptions?

src/selections-factory.js
haxxmaxx: No, this function is always called when clicking a cell. I guess we need to handle clicking outside, I just checked with the client and as you say, one click confirms, another starts new selections

src/selections-factory.js
haxxmaxx: agree

src/index.js
haxxmaxx: that should work

Caele (approved):

PR #50, opened 2021-01-18
feat: clicking outside confirms selections + refactor

### LINKED ISSUE #47: Confirm selections when clicking outside active column
When clicking outside of the column where there are active selections, the selections should be confirmed. This goes both for excluded cells and other areas in the chart

haxxmaxx:
fixes #47


package.json
```diff
@@ -74,7 +75,6 @@
     "rollup": "^2.36.1",
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-peer-deps-external": "^2.2.4",
-    "rollup-plugin-postcss": "^4.0.0",
```
haxxmaxx, line 77: This was complaining, it needed a newer version of postCSS after I added less or somthing. Are we using it or just copied from another packege.json?

src/selections-factory.js
```diff
@@ -81,5 +77,22 @@ export default function initSelections() {
...
+      if (selections.api.isActive() && !isSelectableCell) {
+        e.stopPropagation();
+        selections.api.confirm();
+      }
+    };
+
+    el.addEventListener('click', onClick, true);
+    return () => {
+      el.removeEventListener('click', onClick, true);
+    };
```
haxxmaxx, line 94: Coverage here is a problem here... Need to figure out a way to test this logic.

henrikgildeus: You can add RTL tests that tests logic inside hooks

henrikgildeus: Checkout out `renderHook` from RTL

haxxmaxx: Nice, setting up RTL is on the todo list 👍

package.json
Caele: Don't think you should use less/post-css here, instead rely in css-in js like makeStyles or however the Theme thing works.

Caele: Postcss is used in nebula, so probably where the warning comes from.

src/table/root.jsx
```diff
@@ -22,9 +22,11 @@ export function render(element, props) {
...
 
   ReactDOM.render(
-    <ThemeProvider generateClassName={generateClassName} theme={theme}>
-      <TableWrapper {...props} columnRenderers={columnRenderers} />
-    </ThemeProvider>,
+    <StylesProvider generateClassName={generateClassName}>
+      <ThemeProvider theme={theme}>
+        <TableWrapper {...props} columnRenderers={columnRenderers} />
+      </ThemeProvider>
+    </StylesProvider>,
```
haxxmaxx, line 29: This fixes the mui collisions we had with nebula

src/table/cells/selections.jsx
```diff
@@ -1,17 +1,28 @@
...
+
+const useStyles = makeStyles({
+  selectionStyle: (props) => props,
+});
 
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
     const { cell, selections } = props;
-    const { selectCell, getCellStyle } = selections;
+    const { style, identifyerClass } = selections.getSelectionStyle(cell);
```
Caele, line 12: This feels like it needs to run quite a lot of code to just get the styling for each render.
Each row has to run the 
`  const hasSelectedElemNumber = !!selected.rows?.find((r) => r.qElemNumber === cell.qElemNumber);` 
for example, which could be pre-generated instead?

haxxmaxx:
I've done some digging on the update performance. The main issue is that when clicking on a cell, we have to update the styling for each cell individually. One way around this could possibly be only updating the active column, but since the data is stored in rows, as well as the HTML table, It will most likely complicate things. So what I have done instead:

- Optimized `getSelectionClass` for speed
- Create `handleMouseUp`, apparently if you use a unnamed function, it is created every render
- Make sure useStyles don't take any arguments, this felt faster when testing

Possibly there is more react trickery that could optimize this better, but I think it is decent now compared to the old table.

Caele (approved):

PR #53, opened 2021-01-21
feat: disable selections in edit mode etc

### LINKED ISSUE #52: Diisable selections when not applicable
When in a context when you can't select (api missing, in edit mode etc) nothing should happen when trying to select

haxxmaxx:
fixes #52 

Looked at the org-chart. They store this separate from selections, since it impacts other things. Since this is the only difference between different views/environments so far I suggest we store it on the selections state


src/table/tableBody.jsx
```diff
@@ -1,19 +1,29 @@
...
     colIdx: -1,
+    isEnabled: !!selectionsAPI && !constraints.active,
   });
 
+  useEffect(() => {
+    console.log('setting columns and isEnabled');
+    const isEnabled = !!selectionsAPI && !constraints.active;
+    selDispatch({ type: 'set-enabled', payload: { isEnabled } });
+    setColumnRenderers(tableData.columns.map((c) => getCellRenderer(c, isEnabled)));
+  }, [selectionsAPI, constraints]);
```
haxxmaxx, line 25: I can't think of a case where the selection api is suddenly becomes unavailable, without the layout changing and the entire extension rerenders. If that is the case, `selectionsAPI` is not needed in the dependency array

src/table/cells/__tests__/renderer.spec.js
```diff
@@ -0,0 +1,31 @@
+import TableCell from '@material-ui/core/TableCell';
+import getCellRenderer from '../renderer';
+import selections from '../selections';
```
haxxmaxx, line 3: Importing this is why code coverage is dropped

Caele (approved):

PR #58, opened 2021-01-26
disable no console in eslint

LiKang6688:


haxxmaxx:
Why do you need this? We generally try to avoid it

LiKang6688:
> Why do you need this? We generally try to avoid it

When I use `console.log();` in the project, the eslint will give a warning "Unexpected console statement."
Shall we avoid having the console.log in the codebase?

LiKang6688:
@haxxmaxx I get it. Thank you, Max. It works as intended so this PR can be closed.

PR #63, opened 2021-01-28
chore: make sprout optional

Caele:


src/index.js
```diff
@@ -43,6 +45,7 @@ export default function supernova(env) {
       useEffect(() => {
         if (layout && tableData) {
           render(el, { tableData, setPageInfo, pageInfo, constraints, selections, muiParameters });
+          render(el, { tableData, setPageInfo, pageInfo, constraints, selections });
```
haxxmaxx, line 48: Should there be two render calls here?

src/index.js
```diff
@@ -6,6 +6,8 @@ import ext from './ext';
 import muiSetup from './mui-setup';
 import { render, teardown } from './table/root';
 import manageData from './table/handle-data';
+// This line is replaced by rollback with an import for internal builds
```
haxxmaxx, line 9: rollup?

src/index.js
```diff
@@ -6,6 +6,8 @@ import ext from './ext';
...
+const __OPIONAL_THEME_DEPS__ = {}; // eslint-disable-line no-underscore-dangle
```
haxxmaxx, line 10: I don't know why you use __ before/after the variable name, does it mean it has to do with building or something?

src/index.js
Caele: double the fun?
No, merge problem

src/index.js
Caele: yes

src/index.js
Caele: yea, more or less like a warning "DON*T TOUCH!".

haxxmaxx (approved):

PR #67, opened 2021-02-02
refactor: use useReducer and only run render when data is actually changed

### LINKED ISSUE #62: Improve selections performance
Every time you click a cell, a re-render is triggered all the way from `index`. This also means that cells need to recalculate their styling. Additionally, since `layout` also is changed, this fetches new data and in turn renders the tree. The last one seems to not be affecting the speed of the UI.

We need to investigate whether the cells can store the styling in their state, and only change it when needed. This could reduce calculations significantly

haxxmaxx:
fixes #62

Was forced to rethink how we store the selections. By keeping them in `tableBody` (the only component effected when selections are active) there is no delay between clicking and the cell styling updating. This also enables using `useReducer` which is  much cleaner IMO when when we have a bunch of stuff on one state. An `isEnabled` property will be added later, and corresponding reducer case.


src/table/cells/selections.jsx
```diff
@@ -15,10 +16,10 @@ const useStyles = makeStyles({
 
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
-    const { cell, selections } = props;
+    const { cell, selState, selDispatch } = props;
     const classes = useStyles();
-    const handleMouseUp = () => cell.isDim && selections.selectCell(cell);
-    const selectionClass = selections.getSelectionClass(cell);
+    const handleMouseUp = () => cell.isDim && selectCell(cell, selState, selDispatch);
+    const selectionClass = getSelectionClass(cell, selState);
```
haxxmaxx, line 22: Before these functions where stored on the `selections` state and we only needed to pass `cell`. Tried this, but then it was using a stale state, so decided to import them directly from selection-utils.

src/table/selections-utils.js
```diff
@@ -0,0 +1,79 @@
+export function addSelectionListeners(api, selDispatch) {
```
haxxmaxx, line 1: This file isn't all new, just renamed. `addSelectionListeners` and `reducer` are basically what is added

src/index-helper.js
```diff
@@ -18,3 +18,7 @@ export default function muiSetup(isActive, sproutBase) {
 
   return { theme, generateClassName };
 }
+
+export function shouldUpdateData(layout, tableData) {
+  return !tableData || JSON.stringify(layout.qHyperCube.qSize) !== JSON.stringify(tableData.size);
+}
```
haxxmaxx, line 24: This works, but probably needs to be extended when there are other things that would need to reload the data

Caele: yea, can't know for sure that the size changes just because the data changes.

Caele (commented):
Looks good!
I'm still hoping for a full separation between selections and the rest, but it can wait.

src/index-helper.js
haxxmaxx: True... A different approach would be actually looking at what `manageData` returns and compare it to the old data. This would be slower, but covers what we need I think

Caele (approved):

src/index-helper.js
Caele: In general you should update if the layout hook fires, not entirely sure what difference this makes.

haxxmaxx: The layout keeps track of what the active selections are, so layout is updated every time a cell is clicked, thus renders the whole tree

LiKang6688:
> Well, it should be triggered for clicking anywhere on the extension basically. If selections are active and we press outside the that column, selections should be canceled. This is the behaviour of the old table. I have noticed though, that you can't press the scroll bar when selecting (which should be possible). But I'd say that is out of scope for this PR, this is code is just moved from `selections-factory`

https://user-images.githubusercontent.com/4471489/106646366-f4653680-658d-11eb-8af7-b26d3bb42ee9.mov
Yes, you are right. I found the reason why I feel that does not seem correct, the current table behaves as the video shows. Personally, I think only when people click that confirmation button, the selection should work.  I agree with you that this selection behavior and the scroll bar thing are out of scope for this PR.

PR #71, opened 2021-02-03
feat:toggle select one value with cmd/ctrl

LiKang6688:


src/table/selections-utils.js
```diff
@@ -58,10 +58,15 @@ export function selectCell(cell, selState, selDispatch) {
...
-  if (alreadySelectedIdx > -1) {
+  if (alreadySelectedIdx > -1 && !evt.ctrlKey && !evt.metaKey) {
+    // if the ctrl key or the ⌘ Command key (On Macintosh keyboards) or the ⊞ Windows key is pressed
+    // do not remove the clicked item
     newRows.splice(alreadySelectedIdx, 1);
   } else {
     newRows.push({ qElemNumber, rowIdx });
+    // if the ctrl key or the ⌘ Command key (On Macintosh keyboards) or the ⊞ Windows key is pressed
+    // get the last clicked item
+    (evt.ctrlKey || evt.metaKey) && (newRows = newRows.slice(-1));
```
haxxmaxx, line 69: No need to involve the other cases here, just start with checking if either of the keys are pressed -> set newRows to `[{ qElemNumber, rowIdx }]`

haxxmaxx, line 69: No need to involve the other cases here, just start with checking if either of the keys are pressed -> set newRows to `[{ qElemNumber, rowIdx }]`

haxxmaxx: so
```
if (isCtrl) {
} else if (alreadySelectedIdx) {
} else {
}
```

src/table/selections-utils.js
```diff
@@ -38,37 +38,51 @@ export function getSelectionClass(cell, selState) {
...
     case 'reset':
       return state.rows.length ? { ...state, rows: [], colIdx: -1 } : state;
     default:
       throw new Error('reducer called with invalid action type');
   }
 }
 
-export function selectCell(cell, selState, selDispatch) {
+export const getSelectedRows = (selectedRows, qElemNumber, rowIdx, rows, evt) => {
+  const alreadySelectedIdx = rows.findIndex((r) => r.qElemNumber === qElemNumber);
```
haxxmaxx, line 50: This is actually a typo, it should be `selectedRows.findIndex`. So you don't have to sent `rows` as an argument

.codeclimate.yml
```diff
@@ -5,6 +5,6 @@ checks:
       threshold: 60
   method-complexity:
     config:
-      threshold: 8
+      threshold: 9
```
haxxmaxx, line 8: I don't think this is needed any more, since you broke out some logic from `selectCell`

src/table/__tests__/selections-util.spec.js
```diff
@@ -181,13 +183,30 @@ describe('selections-utils', () => {
...
       };
+      const event = { ctrlKey: false, metaKey: false };
 
-      selectCell(cell, selState, selDispatch);
+      selectCell(cell, selState, selDispatch, event);
+      expect(selState.api.begin).to.not.have.been.called;
+      expect(selState.api.select).to.have.been.calledWith({ method: 'selectHyperCubeCells', params });
+      expect(selDispatch).to.have.been.calledWith({ type: 'select', payload });
+    });
+    it('should not call begin, ctrl/command keyboard to add to selected and call selectHyperCubeCells when selecting new qElemNumber', () => {
```
haxxmaxx, line 198: Since we refactored, I think we should test `getSelectedRows` separately. Then make some simpler tests for `selectCell`, which basically tests the different cases, not that that `selectedRows` is containing the right stuff. We can merge this as is, and I can fix it later, just me being picky about unit test structure

haxxmaxx, line 198: Since we refactored, I think we should test `getSelectedRows` separately. Then make some simpler tests for `selectCell`, which basically tests the different cases, not that that `selectedRows` is containing the right stuff. We can merge this as is, and I can fix it later, just me being picky about unit test structure

src/table/selections-utils.js
LiKang6688: will fix that. Thanks.

.codeclimate.yml
LiKang6688: Okay. Gonna try.

src/table/__tests__/selections-util.spec.js
LiKang6688: Sounds good to me. Thanks.

haxxmaxx (approved):

PR #75, opened 2021-02-04
fix: Disable view data option  in the agnostic table context menu

LiKang6688:


package.json
```diff
@@ -33,7 +33,8 @@
     "sense": "nebula sense",
     "spec": "scriptappy-from-jsdoc -c ./spec-configs/props.conf.js",
     "test:unit": "aw -c aw.config.js",
-    "test:integration": "aw puppet --testExt '*.int.js' --glob 'test/integration/**/*.int.js'"
+    "test:integration": "aw puppet --testExt '*.int.js' --glob 'test/integration/**/*.int.js'",
+    "extension": "yarn build && cp -r ./dist ~/Qlik/Sense/Extensions"
```
haxxmaxx, line 37: Don't want to push this I think, it's very specific to developing against sense-client

LiKang6688: @haxxmaxx  I imagine that we will develop against sense-client quite often. Do you think so?

haxxmaxx: We will, but we should keep the references to private code to a minimum. i don't think this command  is applicable if you don't have access to sense-client

LiKang6688: Good point. Did not consider that. I will remove it.

haxxmaxx (approved):

PR #78, opened 2021-02-05
fix: disable table pagination in edit mode

LiKang6688:


src/table/table.jsx
```diff
@@ -20,6 +20,9 @@ const useStyles = makeStyles({
   containerOverflowHidden: {
     overflow: 'hidden',
   },
+  EditMode: {
+    display: 'none',
+  },
```
Caele, line 25: This is a very general classname for a specific attribute. It is fine for now, but I'm sure there will be more css changes for EditMode, and those probably not including hiding the element.

Caele (approved):

src/table/table.jsx
LiKang6688: Good point. Thanks. I will keep it in mind.

haxxmaxx: yeah it is a bit misleading though, since it is not only in edit mode where `constraints.active` is true. I'd say call it what it is, maybe `paginationHidden` or something similar

LiKang6688: @haxxmaxx Great. Fixed that.

haxxmaxx (approved):
👌 👌

PR #89, opened 2021-02-09
test: first rtl test

haxxmaxx:
Setting up RTL and adding test for `TabelWrapper`. The big question mark is if the `test-setup` approach is solid of if there should be more/less in the script. WIll cover other components in next PR -> get coverage back up

Also some renaming and small refactor of `withSlections`


src/table/__tests__/TableWrapper.spec.jsx
```diff
@@ -0,0 +1,86 @@
...
+    rowsPerPage = 100;
+  });
+
+  beforeEach(() => {
+    sandbox.verifyAndRestore();
+    sandbox.resetHistory();
+  });
+
+  it('should render table', () => {
+    mockContent();
```
haxxmaxx, line 39: @niekvanstaveren Can't seem to run this in `beforeEach`, don't see why? Ideas?

src/table/__tests__/TableWrapper.spec.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+    const { queryByLabelText, queryByText } = render(
+      <TableWrapper tableData={tableData} setPageInfo={setPageInfo} constraints={constraints} />
+    );
+
+    expect(queryByLabelText('sticky table')).to.be.visible;
+    expect(queryByText(`1-${rowsPerPage} of ${tableData.size.qcy}`)).to.be.visible;
+    expect(queryByText(rowsPerPage)).to.be.visible;
```
haxxmaxx, line 47: There are a couple of other strings to look at, but this kind of what we control, the rest is Mui. Enough?

src/table/__tests__/TableWrapper.spec.jsx
niekvanstaveren: You have two beforeEach defined, shouldn't the second one be an afterEach? Not sure if that fixes your issue.....

haxxmaxx: of course...

src/table/TableHeadWrapper.jsx
```diff
@@ -0,0 +1,24 @@
...
+          <TableCell key={column.id} align={column.align} style={{ minWidth: column.minWidth }}>
+            {column.label}
+          </TableCell>
+        ))}
+      </TableRow>
+    </TableHead>);
+}
+
+TableHeadWrapper.propTypes = {
+  columns: PropTypes.array.isRequired,
```
niekvanstaveren, line 23: Is this correct? Seems columns is not a direct property

src/table/TableWrapper.jsx
```diff
@@ -3,12 +3,10 @@ import PropTypes from 'prop-types';
...
 import Paper from '@material-ui/core/Paper';
 import Table from '@material-ui/core/Table';
-import TableCell from '@material-ui/core/TableCell';
 import TableContainer from '@material-ui/core/TableContainer';
-import TableHead from '@material-ui/core/TableHead';
 import TablePagination from '@material-ui/core/TablePagination';
-import TableRow from '@material-ui/core/TableRow';
-import TableBodyWrapper from './tableBody';
+import TableBodyWrapper from './TableHeadWrapper';
+import TableHeadWrapper from './TableBodyWrapper';
```
niekvanstaveren, line 9: ????

src/table/__tests__/TableWrapper.spec.jsx
```diff
@@ -0,0 +1,79 @@
...
+  it('should change back to first page when not on first page and no rows', async () => {
+    tableData.rows = [];
+
+    const { findByTitle } = render(
+      <TableWrapper tableData={tableData} setPageInfo={setPageInfo} constraints={constraints} />
+    );
+    fireEvent.click(await findByTitle('Next page'));
+
+    // Called when pressing the button
+    expect(setPageInfo).to.have.been.calledWith({ top: rowsPerPage, height: rowsPerPage });
```
niekvanstaveren, line 65: Interesting that this is still called, should the button not be inactive?

niekvanstaveren (approved):
Looks good

src/table/__tests__/TableWrapper.spec.jsx
haxxmaxx: Yeah this pattern is weird. i just looked at it. the pagination does not look at `rows`, only `size.qcy`. I think I should do a somewhat realistic flow here

src/table/TableWrapper.jsx
haxxmaxx: hahaha

src/table/TableHeadWrapper.jsx
haxxmaxx: This is incorrect, I'll fix

src/table/__tests__/TableWrapper.spec.jsx
haxxmaxx: Actually, I think it is fine. It is a hacky way to simulate the real use case: that you confirm selections on any page past the first page and you don't select more rows than rows per page. The data paging then only fetches rows 101-200 or something when there are less than 100 rows, which returns 0 rows.

PR #90, opened 2021-02-10
Crl/component rtl

haxxmaxx:


src/table/handle-data.js
```diff
@@ -30,7 +29,7 @@ export default async function manageData(model, layout, pageInfo) {
 
   const columns = columnorder.map((c) => getColumnInfo(layout, c));
   const rows = matrix.map((r, rowIdx) => {
-    const row = {};
+    const row = { key: rowIdx };
```
haxxmaxx, line 32: Using index as keys should generally be avoided. But here we are adding a key (that happens to be an index) before rendering starts, not using the index of the array when looping over the array. So this should be fine. (see https://reactjs.org/docs/lists-and-keys.html#keys)

src/table/handle-data.js
```diff
@@ -15,7 +15,6 @@ export function getColumnInfo(layout, colIndex) {
     isDim,
     width: 200,
     label: info.qFallbackTitle,
-    dataKey: info.cId,
```
haxxmaxx, line 18: This wasn't used anywhere, so removing it

haxxmaxx:
Actually, I'll do the `selections` component as well, putting back to draft

PR #91, opened 2021-02-10
test: add RTL tests for remaining components

### LINKED ISSUE #22: chore: RTL testing
setup RTL testing. For now only some initial tests for the table component is needed, the selection and icon parts are most likely to change.

haxxmaxx:
fixes #22


src/table/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -0,0 +1,65 @@
...
+    const { queryByText } = render(
+      <TableBodyWrapper tableData={tableData} constraints={constraints} selectionsAPI={selectionsAPI} />
+    );
+
+    expect(queryByText(tableData.rows[0]['id-0'].qText)).to.be.visible;
+    expect(queryByText(tableData.rows[0]['id-1'].qText)).to.be.visible;
+    expect(queryByText(tableData.rows[1]['id-0'].qText)).to.be.visible;
+    expect(queryByText(tableData.rows[1]['id-1'].qText)).to.be.visible;
+  });
+  it('should render 2x2 table body and call CellRenderer', () => {
```
niekvanstaveren, line 47: I'm a bit confused here on what you exactly want to test between these two.

niekvanstaveren (approved):

src/table/__tests__/TableBodyWrapper.spec.jsx
haxxmaxx: If you look at the logic in `TableBodyWrapper`, there are two cases for rendering cells. If there is no CellRenderer for that column, it uses the mui TableCell, hence two tests. Both test should render a table body, but using different components

niekvanstaveren: Hmm now I see that the sandbox replace is different indeed. However from getCellRenderer you will always get back something right? Default seems to be the TableCell as well.

haxxmaxx: Hmm yeah, getCellRenderer used to only return something for dimensions, but now it always returns something. So might be able to remove the logic all together. I'll keep it for now, but will have it in mind when we add column styling etc

PR #92, opened 2021-02-10
fix: next page button UX/UI in mobile

### LINKED ISSUE #48: Next page button in agnostic table is not fully displayed in portrait mode on a mobile
Next page button in agnostic table is not fully displayed in portrait mode on a mobile

LiKang6688:
fixes #48

Also switching to `isModal` to check if selections are active. Needs an update to work in sense, it will be tested at the same time as this.


src/table/TableWrapper.jsx
```diff
@@ -21,16 +21,28 @@ const useStyles = makeStyles({
...
     display: 'none',
   },
+  containerMobileLayout: {
+    height: 'calc(100% - 120px)',
+  },
+  paginationMobileLayout: {
+    marginTop: '-15px',
+    '& div:nth-of-type(1)': {
+      display: 'initial',
+    },
```
haxxmaxx, line 31: What does this do?

src/table/TableWrapper.jsx
```diff
@@ -21,16 +21,28 @@ const useStyles = makeStyles({
...
 export default function TableWrapper(props) {
   const { tableData, setPageInfo, constraints } = props;
   const { size, rows } = tableData;
+  const [tableWidth, setTableWidth] = useState();
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(100);
   const classes = useStyles();
   const containerMode = constraints.active ? 'containerOverflowHidden' : 'containerOverflowAuto';
   const clientMode = constraints.active && 'paginationHidden';
+  const containerLayout = tableWidth < 380 && 'containerMobileLayout';
```
haxxmaxx, line 44: is 380 where sense-client goes into mobile mode?

src/table/TableWrapper.jsx
```diff
@@ -49,16 +61,23 @@ export default function TableWrapper(props) {
     return null;
   }
 
+  const tableContainer = useRef();
+
+  useEffect(() => {
+    window.console.log(tableContainer.current.clientWidth, 'container');
```
haxxmaxx, line 67: I think it's easier to pass `element` from index. It updates as soon as the size changes, so I think it would work as a dependency for useEffect

haxxmaxx, line 67: I think it's easier to pass `element` from index. It updates as soon as the size changes, so I think it would work as a dependency for useEffect

src/table/TableWrapper.jsx
LiKang6688: it is 380px that the next page button starts to become invisible

src/table/TableWrapper.jsx
LiKang6688: good idea. I will try it.

src/table/TableWrapper.jsx
LiKang6688: it is a trick that to get the first child element of the TablePagination component that is a div.

src/table/TableWrapper.jsx
haxxmaxx: But why 380?

src/table/TableWrapper.jsx
```diff
@@ -49,16 +61,22 @@ export default function TableWrapper(props) {
     return null;
   }
 
+  useEffect(() => {
+    const updateSize = () => setTableWidth(el.clientWidth);
+    window.addEventListener('resize', updateSize);
+    return () => window.removeEventListener('resize', updateSize);
```
haxxmaxx, line 67: No need to add another listener since `el.clientWidth` updates on resize. So adding that the dependency array and just call `setTableWidth` should work fine

src/table/TableWrapper.jsx
LiKang6688: I just change the window size and when the window size is 380px, the next page button starts to become invisible

src/table/TableWrapper.jsx
LiKang6688: no, not really. I tried the way that you described but it does not work. I belive that the reason is that **const el = useElement();** and **render(el, {})** way to render UI.

src/table/TableWrapper.jsx
haxxmaxx: okok

haxxmaxx: Changed to 400 since it gets a bit tight if you hav lik 100000 rows, the `1-100 of 100000` string becomes long

PR #93, opened 2021-02-12
fix: only confirm selections when clicking confirm or outside

### LINKED ISSUE #86: Selections are confirmed when moving to another page in the agnostic table
Selections are confirmed when moving to another page in the agnostic table. This makes it impossible to make selections on other dimension values then on the ones displayed on the page where we start making selections on.

haxxmaxx:
fixes #86

Since we don't have to cancel clicks all over the chart, we only need one listener on the table body and `selectCell` checks if we are pressing a valid cell


src/table/selections-utils.js
```diff
@@ -74,10 +74,12 @@ export function selectCell(cell, selState, selDispatch, evt) {
   const { rowIdx, colIdx, qElemNumber } = cell;
   let selectedRows = [];
 
-  if (!api.isActive()) {
+  if (selState.colIdx === -1) {
```
haxxmaxx, line 77: This is always up to date, while there is an bug on isActive (returning false positives)

LiKang6688 (approved):
looks good to me. tried in the sense-client and did not find any issue.

src/table/selections-utils.js
LiKang6688: Could you help me understand the colIdx?

haxxmaxx:
> looks good to me. tried in the sense-client and did not find any issue.

Sense works fine, But there is a bug running in nebula https://github.com/qlik-oss/nebula.js/issues/576. I'll merge this for now, Tobias is working on the nebula bug

PR #96, opened 2021-02-15
feat: header styling

haxxmaxx:
Initial structure for header styling, similar work needs to be done for the table body

### DIFF
--- a/src/table/styling-utils.js
+++ b/src/table/styling-utils.js
@@ -1,0 +1,11 @@
+export function getColor(color, defaultColor, theme) {
+  const resolvedColor = theme.getColorPickerColor(color);
+  return !resolvedColor || resolvedColor === 'none' ? defaultColor : resolvedColor;
+}
+
+export function getHeadStyle(layout, theme) {
+  return {
+    fontColor: getColor(layout.components[0].header.fontColor, '#000', theme),
+    fontSize: layout.components[0].header.fontSize,
+  };
+}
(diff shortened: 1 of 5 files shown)


src/table/styling-utils.js
```diff
@@ -0,0 +1,14 @@
+export function getColor(color, defaultColor, theme) {
+  const resolvedColor = theme.getColorPickerColor(color);
+  return !resolvedColor || resolvedColor === 'none' ? defaultColor : resolvedColor;
+}
```
LiKang6688, line 4: when I set a color, it gives me another one. My gut feeling is that it is not correct.

haxxmaxx: `theme.getColorPickerColor` works fine for me, are you sure you have formatted the properties correctly?

Good point. Well, for old tables `fontColor` will never be empty. If you never have opened the styling panel for that object, `component` array wont exist. Once you do, it is populated with default values. That being said, we should probably do more sanity checking than this. It's a bit unclear what we need at this point, only support properties form old table in sense or other cases as well. I'll ask Tobias

haxxmaxx: I think I covered it below. Layout should always exist, otherwise you are in all kinds of trouble. Then checking all steps except `fontColor`, but that is handled by `getColorPickerColor`. I've tried sending in a bunch of stuff, including `""` and it just returns undefined, which we cover with the default color.

LiKang6688 (approved):

PR #97, opened 2021-02-15
add GenericObjectProperties

LiKang6688:


haxxmaxx:
As Tobias said, we shouldn't put anything in object-propreties yet. So you can close this :)

PR #98, opened 2021-02-15
feat: add table body props

LiKang6688:


src/table/styling-utils.js
```diff
@@ -1,14 +1,24 @@
 export function getColor(color, defaultColor, theme) {
-  const resolvedColor = theme.getColorPickerColor(color);
-  return !resolvedColor || resolvedColor === 'none' ? defaultColor : resolvedColor;
+  const resolvedColor = color && theme.getColorPickerColor(color);
+  return !resolvedColor || resolvedColor === 'none' ? color || defaultColor : resolvedColor;
```
haxxmaxx, line 3: As I commented on the other PR, no need to sanity check `color`, `getColorPickerColor` just returns undefined. Also, this we always want to fall back to `defaultColor` if `resolvedColor` is not valid

src/table/cells/selections.jsx
```diff
@@ -17,17 +17,22 @@ const useStyles = makeStyles({
...
+    const { cell, selState, selDispatch, className } = props;
     const classes = useStyles();
     const handleMouseUp = (evt) => cell.isDim && selectCell(cell, selState, selDispatch, evt);
     const selectionClass = getSelectionClass(cell, selState);
 
     return (
-      <CellComponent {...props} className={`${selectionClass} ${classes[selectionClass]}`} onMouseUp={handleMouseUp} />
+      <CellComponent
+        {...props}
+        className={`${className} ${selectionClass} ${classes[selectionClass]}`}
```
haxxmaxx, line 28: This is not applied to measures, i.e. cells that is not rendered with `withSelections`. We need a new HOC called `withStyling` (wrapping `withSelections` that applies styling independent of selections. Still have to do what you do in here: reset `className` with the new selection classes.

haxxmaxx: Correction, it is applied to meassures, but it still should work without `withSelections`

src/table/TableBodyWrapper.jsx
```diff
@@ -48,7 +59,8 @@ export default function TableBodyWrapper({ tableData, constraints, selectionsAPI
                 {value}
               </CellRenderer>
             ) : (
-              <TableCell key={column.id} align={column.align}>
+              // the new added column in the edit mode, both dimension and measure
+              <TableCell className={classes.body} key={column.id} align={column.align}>
```
haxxmaxx, line 63: I think we should remove this. We always get a component from getCellRenderer, so this never runs

src/table/cells/renderer.js
```diff
@@ -6,9 +6,11 @@ import { withSelections } from './selections';
 
 // column will be needed in the future when column styling is implemented
 export function getCellRenderer(column, selectionsEnabled) {
+  // the column in the analysis mode
   if (selectionsEnabled) {
     return withSelections(TableCell);
   }
 
+  // the old column in the edit mode
```
haxxmaxx, line 14: old column? We shouldn't reference edit/analysis mode yet, since it will be a while since the table is used in sense.

haxxmaxx:
Nice that you did not need another HOC, this is much cleaner. I still see the wrong font when selecting though. It should override so it is always white when selecting.
Screenshot 2021-02-18 at 10 57 19

LiKang6688:
> Nice that you did not need another HOC, this is much cleaner. I still see the wrong font when selecting though. It should override so it is always white when selecting.
> Screenshot 2021-02-18 at 10 57 19

@haxxmaxx okay, thanks, now I know that and will fix it.

src/table/cells/withoutSelections.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
haxxmaxx, line 1: I dont think this is a good idea. There will be another HOC soon for column styling, so we should only have `with...` HOCs. It should still work if you set `className` in `tableBodyWrapper` and then return just `TableCell` from `getCellRenderer`, the way it was before

LiKang6688: yeah, sure

LiKang6688: Done

src/table/cells/selections.jsx
```diff
@@ -9,23 +8,30 @@ const useStyles = makeStyles({
...
   selected: {
-    color: '#fff',
+    color: '#fff !important',
     backgroundColor: '#009845',
   },
   possible: {},
 });
 
-export function withSelections(CellComponent) {
+export default function withSelections(CellComponent) {
```
haxxmaxx, line 17: The reason there was no `default` here we to be able to mock it for testing. If you can make it work with default, we can keep it, otherwise it shouldn't be there

src/table/cells/selections.jsx
```diff
@@ -9,23 +8,30 @@ const useStyles = makeStyles({
     background: 'repeating-linear-gradient(-45deg, #fafafa, #fafafa 2px, #eee 2.5px, #fafafa 3px, #fafafa 5px)',
   },
   selected: {
-    color: '#fff',
+    color: '#fff !important',
```
haxxmaxx, line 11: Hoping it wouldn't come to this, but it maybe there is no other way around it. Since there will be another layer of styling soon, I think it is fine for now, then maybe will be refactored

src/table/TableBodyWrapper.jsx
```diff
@@ -35,8 +44,10 @@ export default function TableBodyWrapper({ tableData, constraints, selectionsAPI
             const cell = row[column.id];
             const value = cell.qText;
             const CellRenderer = columnRenderers[i];
-            return CellRenderer ? (
+            return (
               <CellRenderer
+                className={classes.tableCell}
+                stylingClassName={classes.tableCell}
```
haxxmaxx, line 50: This is just a preference, but I think it was cleaner with only passing `className`, then using that in `withStyling`.

LiKang6688: i agree. will fix.

src/table/cells/selections.jsx
LiKang6688: now it works with default

src/table/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -33,10 +39,17 @@ describe('<TableBodyWrapper />', async () => {
   });
 
   it('should render 2x2 table body without CellRenderer', () => {
-    sandbox.replace(renderer, 'getCellRenderer', () => {});
+    // eslint-disable-next-line react/prop-types
```
haxxmaxx, line 42: This entire test should be removed. It was testing the case you removed from `tableBodyWrapper`

src/table/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -47,16 +60,22 @@ describe('<TableBodyWrapper />', async () => {
   it('should render 2x2 table body and call CellRenderer', () => {
     const cellRendererSpy = sinon.spy();
     // eslint-disable-next-line react/prop-types
-    sandbox.replace(renderer, 'getCellRenderer', () => ({ value }) => {
+    sandbox.replace(getCellRenderer, 'default', () => ({ value }) => {
```
haxxmaxx, line 63: NIce! I never got this to work, god job :)

haxxmaxx (approved):
Remove the test from `TableBodyWrapper.spec.jsx` and it's all done

PR #100, opened 2021-02-16
feat: styling panel hover

LiKang6688:


src/table/TableBodyWrapper.jsx
```diff
@@ -9,15 +9,24 @@ import { getCellRenderer } from './cells/renderer';
...
+    '&&:hover': {
+      backgroundColor: props.hoverBackGroundColor,
+    },
   }),
 });
 
 export default function TableBodyWrapper({ tableData, constraints, selectionsAPI, layout, theme }) {
   const { rows, columns } = tableData;
   const classes = useStyles(getBodyStyle(layout, theme));
+  const hoverEffect = layout?.components?.[0]?.hoverEffect;
```
haxxmaxx, line 29:
```suggestion
  const hoverEffect = layout.components?.[0]?.hoverEffect;
```
layout will always exist, we check that in index before rendering

src/table/TableBodyWrapper.jsx
```diff
@@ -39,14 +48,14 @@ export default function TableBodyWrapper({ tableData, constraints, selectionsAPI
...
-        <TableRow hover role="checkbox" tabIndex={-1} key={row.key}>
+        <TableRow hover={hoverEffect} role="checkbox" tabIndex={-1} key={row.key} className={classes.tableRow}>
           {columns.map((column, i) => {
             const cell = row[column.id];
             const value = cell.qText;
             const CellRenderer = columnRenderers[i];
             return CellRenderer ? (
               <CellRenderer
-                className={classes.body}
+                classCellName={classes.tableCell}
```
haxxmaxx, line 58: I think this could have a bit more descriptive name, since there are other classes on the cell related to selections. Maybe `stylingClass`?

src/table/cells/selections.jsx
```diff
@@ -17,22 +17,24 @@ const useStyles = makeStyles({
 
 export function withSelections(CellComponent) {
   const HOC = (props) => {
-    const { cell, selState, selDispatch, className } = props;
+    const { cell, selState, selDispatch, classCellName } = props;
+    window.console.log(classCellName, 'classCellName');
```
haxxmaxx, line 21: remove log

src/table/cells/selections.jsx
```diff
@@ -17,22 +17,24 @@ const useStyles = makeStyles({
...
+
     const classes = useStyles();
     const handleMouseUp = (evt) => cell.isDim && selectCell(cell, selState, selDispatch, evt);
     const selectionClass = getSelectionClass(cell, selState);
 
     return (
       <CellComponent
         {...props}
-        className={`${className} ${selectionClass} ${classes[selectionClass]}`}
+        className={`${classCellName} ${selectionClass} ${classes[selectionClass]}`}
```
haxxmaxx, line 30: what happens in the case of hovering when making selections? Does it look the same as in the client?

src/table/styling-utils.js
```diff
@@ -15,10 +15,14 @@ export function getHeadStyle(layout, theme) {
 
 export function getBodyStyle(layout, theme) {
   const content = layout?.components?.[0]?.content;
+  const hoverColor = layout?.components?.[0]?.hoverColor;
+  const hoverFontColor = layout?.components?.[0]?.hoverFontColor;
```
haxxmaxx, line 19: Same thing with sanity checking here, layout will always be an object

src/table/TableBodyWrapper.jsx
LiKang6688: okay, gonna fix

src/table/cells/selections.jsx
LiKang6688: it looks the same as the old table and the hover effect is for each row.

src/table/TableBodyWrapper.jsx
LiKang6688: hoverEffect does not work. will fix it.

src/table/TableBodyWrapper.jsx
LiKang6688: sure. stylingClassCellName?

src/table/cells/selections.jsx
LiKang6688: no, I am wrong. it does not look correct in sense-client. Look into it.

src/index.js
```diff
@@ -56,9 +57,10 @@ export default function supernova(env) {
             selectionsAPI,
             muiParameters,
             theme,
+            cellLockedImage,
           });
         }
-      }, [tableData, constraints, selectionsAPI.isModal()]);
+      }, [tableData, constraints, selectionsAPI.isActive()]);
```
haxxmaxx, line 63: This should be isModal. Currently it's not working in sense, but I have a fix for that in sense-client on the way in

src/table/TableBodyWrapper.jsx
```diff
@@ -12,12 +12,22 @@ const useStyles = makeStyles({
...
 });
 
-export default function TableBodyWrapper({ tableData, constraints, selectionsAPI, layout, theme }) {
+export default function TableBodyWrapper({ tableData, constraints, selectionsAPI, layout, theme, cellLockedImage }) {
   const { rows, columns } = tableData;
-  const classes = useStyles(getBodyStyle(layout, theme));
+  const hoverEffect = layout.components?.[0]?.hoverEffect;
+  const classes = useStyles(getBodyStyle(layout, theme, hoverEffect));
   const getColumnRenderers = (selectionsEnabled) => tableData.columns.map((c) => getCellRenderer(c, selectionsEnabled));
+  const [selectionsEnabled, setSelectionsEnabled] = useState(false);
```
haxxmaxx, line 30: Why do you need an extra state to store this? it's already a property on `cellState`.

src/table/TableBodyWrapper.jsx
```diff
@@ -12,12 +12,22 @@ const useStyles = makeStyles({
...
+      },
+    },
+  }),
 });
 
-export default function TableBodyWrapper({ tableData, constraints, selectionsAPI, layout, theme }) {
+export default function TableBodyWrapper({ tableData, constraints, selectionsAPI, layout, theme, cellLockedImage }) {
   const { rows, columns } = tableData;
-  const classes = useStyles(getBodyStyle(layout, theme));
+  const hoverEffect = layout.components?.[0]?.hoverEffect;
```
haxxmaxx, line 27: It's actually:

```suggestion
  const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
```

haxxmaxx, line 27: It's actually:

```suggestion
  const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
```

src/table/TableBodyWrapper.jsx
```diff
@@ -39,24 +50,33 @@ export default function TableBodyWrapper({ tableData, constraints, selectionsAPI
...
-              </CellRenderer>
+              CellRenderer && (
+                <CellRenderer
+                  className={classes.tableCell}
+                  cell={cell}
+                  column={column}
+                  value={value}
+                  key={column.id}
+                  align={column.align}
+                  cellLockedImage={cellLockedImage}
```
haxxmaxx, line 73: I'm not sure about this. Have you seen any difference in selection performance? Not sure if react handles that, but it would need to pass that image to every cell, every time you make a selection. Did you try mocking the image when testing?

src/table/styling-utils.js
```diff
@@ -15,10 +15,14 @@ export function getHeadStyle(layout, theme) {
...
 export function getBodyStyle(layout, theme) {
   const content = layout.components?.[0]?.content;
+  const hoverColor = layout.components?.[0]?.hoverColor;
+  const hoverFontColor = layout.components?.[0]?.hoverFontColor;
   return content
     ? {
         fontColor: getColor(content.fontColor, '#404040', theme),
         fontSize: content.fontSize || '14px',
+        hoverBackGroundColor: `${getColor(hoverColor, 'blue', theme)} !important`,
+        hoverFontColor: getColor(hoverFontColor, 'black', theme),
```
haxxmaxx, line 25: Need more appropriate defaults here. Some kind of light gray as background, and no change to the font color

src/index.js
LiKang6688: okay, when that fix is done, please let me know. I will change it to isModal.

src/table/TableBodyWrapper.jsx
LiKang6688: Screenshot 2021-02-22 at 10 22 45
From what I see in sense-client, the hoverEffect is not within the content.

haxxmaxx: it is, that } is closing the fontColor property.

src/table/TableBodyWrapper.jsx
LiKang6688: I missing that. Will fix it. Thanks.

src/table/TableBodyWrapper.jsx
LiKang6688: yes, you are right!

src/table/styling-utils.js
LiKang6688: thanks fixed

src/table/styling-utils.js
```diff
@@ -15,14 +15,12 @@ export function getHeadStyle(layout, theme) {
...
-  const hoverColor = layout.components?.[0]?.hoverColor;
-  const hoverFontColor = layout.components?.[0]?.hoverFontColor;
   return content
     ? {
         fontColor: getColor(content.fontColor, '#404040', theme),
         fontSize: content.fontSize || '14px',
-        hoverBackGroundColor: `${getColor(hoverColor, 'blue', theme)} !important`,
-        hoverFontColor: getColor(hoverFontColor, 'black', theme),
+        hoverBackGroundColor: `${getColor(content.hoverColor, '#D3D3D3', theme)} !important`,
+        hoverFontColor: getColor(content.hoverFontColor, 'black', theme),
```
haxxmaxx, line 23: This still overrides `fontColor`, so `this.fontColor` (not sure you can use `this` here, but something similar) should be used as default value

LiKang6688: What do you mean? If I understand correctly, the hoverFontColor should override fontColor in tableCell, if the hover is taking effect.

haxxmaxx: Only if there is a `hoverFontColor` set. We want the same behaviour as the old table. I just looked at it and it is a bit more complex than I thought:
- If no font color is set (hover or non-hover) and a hover background color is set -> the font color becomes white when the background color is a darker shade. There is logic for this inside the sn-action-button we could reuse.
- If you set a hover font color but no hover background color -> only a font color is applied when hovering
- If hovering is turned on but no hover colors are set -> use the defualt light gray background color and don't change the font color

I think that was all. I'd say have a look in sense in the old table that you get all the behaviours. This should have been specified in the issue, sorry about finding out about this now.

LiKang6688: no worries, good points really. Not thinking of all these before. Will look into all those cases.

src/index.js
haxxmaxx: You can change back now

src/table/styling-utils.js
```diff
@@ -19,6 +19,12 @@ export function getBodyStyle(layout, theme) {
     ? {
         fontColor: getColor(content.fontColor, '#404040', theme),
         fontSize: content.fontSize || '14px',
+        hoverBackGroundColor: `${
+          content.hoverFontColor && !content.hoverColor
+            ? 'rgba(0, 0, 0, 0);'
+            : getColor(content.hoverColor, '#E5E5E5', theme)
+        } !important`,
```
haxxmaxx, line 26: I important needed here? What happens without it?

src/table/TableBodyWrapper.jsx
```diff
@@ -8,15 +8,28 @@ import getCellRenderer from './cells/renderer';
 import { getBodyStyle } from './styling-utils';
 
 const useStyles = makeStyles({
+  // 1. hoverEffect is true, there is no font color (hover or non-hover) and a hover background color, the font color does not change when the background color is set.
+  // 2. hoverEffect is true, there is a hover font color but no hover background color, only a hover font color is applied when hovering
+  // 3. hoverEffect is true, there is a font color and a hover font colors and no hover background color
+  // 4. hoverEffect is true, there is a font color and no hover font colors and no hover background color
```
haxxmaxx, line 14: I think it's good to have some comments on this, since it is rather complex. But I'd say keep it in `styling-utils` where the logic actually is

src/table/TableBodyWrapper.jsx
```diff
@@ -39,24 +48,33 @@ export default function TableBodyWrapper({ tableData, constraints, selectionsAPI
   return (
     <TableBody>
       {rows.map((row) => (
-        <TableRow hover role="checkbox" tabIndex={-1} key={row.key}>
+        <TableRow
+          hover={selState.isEnabled && hoverEffect}
+          role="checkbox"
+          tabIndex={-1}
+          key={row.key}
+          className={selState.isEnabled && hoverEffect && classes.hoverTableRow}
```
haxxmaxx, line 56: I'd say we would still want the hover effect when selections are disabled

src/table/styling-utils.js
```diff
@@ -15,10 +15,32 @@ export function getHeadStyle(layout, theme) {
...
+  //   "index": -1,
+  //   "color": null
+  // } means the hover fornt color is unset.
+  const unsetHoverFontColor =
+    !content.hoverFontColor || JSON.stringify(content.hoverFontColor) === JSON.stringify({ index: -1, color: null });
+  const unsetHoverColor =
+    !content.hoverColor || JSON.stringify(content.hoverColor) === JSON.stringify({ index: -1, color: null });
+  const unsetProps = unsetHoverFontColor && unsetHoverColor;
+
+  // 1. hoverEffect is true, there is no font color (hover or non-hover) and a hover background color, the hover font color becomes white and the hover background color is set.
```
haxxmaxx, line 28: Close, but not quite. If the hover background color is "dark", we set the font color to white, otherwise it is ontouched

src/table/TableBodyWrapper.jsx
LiKang6688: sure

src/table/styling-utils.js
```diff
@@ -13,12 +14,84 @@ export function getHeadStyle(layout, theme) {
...
+    ? ''
+    : unsetFontColor
+    ? isDarkColor(getHoverBackgroundColor)
+      ? getColor(content.hoverFontColor, '#ffffff', theme)
+      : getColor(content.hoverFontColor, '#000000', theme)
+    : unsetHoverFontColor
+    ? isDarkColor(getHoverBackgroundColor)
+      ? '#ffffff'
+      : ''
+    : getColor(content.hoverFontColor, '', theme);
```
haxxmaxx, line 89:
```suggestion
  const fontColor = getColor(content.fontColor, '#404040', theme);
  const getHoverFontColor = unsetHoverFontandBackgroundColor
    ? fontColor
    : getColor(content.hoverFontColor, isDarkColor(getHoverBackgroundColor) ? '#ffffff' : fontColor, theme);
```

Did som fiddeling. All you need is this actually. You need to set `fontColor` before, so you know what it is. setting '' wont override the old font color if there was one the you remove it.

haxxmaxx: So what happens here is:
neither hover background nor font -> hover font color the same as whetever font color is.
hover background or font color is set -> call getColor: if hoverFontColor has a valid value, use that. Otherwise use the default, which is either white or whatever the normal font color is

src/table/styling-utils.js
```diff
@@ -13,12 +14,84 @@ export function getHeadStyle(layout, theme) {
...
+  // or the hover font color becomes white if the hover background color is dark.
+
+  // 3. hoverEffect is true, there is a hover font color but no hover background color,
+  // when hovering, only a hover font color is applied when hovering and and the hover background color disappears.
+
+  // 4. hoverEffect is true, there is a font color but no hover font color and no hover background color, when hovering, the defalut hover effect
+
+  // 4. hoverEffect is true, there are all colors, when hovering, the set hover font color and the the hover background color take effect.
+
+  const getHoverBackgroundColor = unsetHoverFontandBackgroundColor
```
haxxmaxx, line 73: These are not getters, so just call them `hoverBackgroundColor` and `hoverFontColor`

src/table/styling-utils.js
```diff
@@ -13,12 +14,84 @@ export function getHeadStyle(layout, theme) {
...
+    b = parseInt(matches[3], 16);
+  }
+  // Using the HSP (Highly Sensitive Poo) value, determine whether the color is light or dark
+  // HSP < 125, the color is dark, otherwise, the color is light
+  return 0.299 * r + 0.587 * g + 0.114 * b < 125;
+}
+
+function isUnset(prop) {
+  return !prop || JSON.stringify(prop) === JSON.stringify({ index: -1, color: null });
+}
```
haxxmaxx, line 43: Nice

src/table/styling-utils.js
```diff
@@ -13,12 +14,84 @@ export function getHeadStyle(layout, theme) {
...
+    : unsetHoverFontColor
+    ? isDarkColor(getHoverBackgroundColor)
+      ? '#ffffff'
+      : ''
+    : getColor(content.hoverFontColor, '', theme);
+
+  return {
+    fontColor: getColor(content.fontColor, '#404040', theme),
+    fontSize: content.fontSize || '14px',
+    hoverBackgroundColor: `${getHoverBackgroundColor} !important`,
```
haxxmaxx, line 94: important is not needed. The selection background will still override the hover background. This is the way it is in the old table, so let's keep it like that for now at least

src/table/styling-utils.js
LiKang6688: sure

src/table/styling-utils.js
```diff
@@ -13,12 +14,84 @@ export function getHeadStyle(layout, theme) {
...
+    ? ''
+    : unsetFontColor
+    ? isDarkColor(hoverBackgroundColor)
+      ? getColor(content.hoverFontColor, '#ffffff', theme)
+      : getColor(content.hoverFontColor, '#000000', theme)
+    : unsetHoverFontColor
+    ? isDarkColor(hoverBackgroundColor)
+      ? '#ffffff'
+      : ''
+    : getColor(content.hoverFontColor, '', theme);
```
haxxmaxx, line 89:
```suggestion
  const fontColor = getColor(content.fontColor, '#404040', theme);
  const getHoverFontColor = unsetHoverFontandBackgroundColor
    ? fontColor
    : getColor(content.hoverFontColor, isDarkColor(getHoverBackgroundColor) ? '#ffffff' : fontColor, theme);
```

Did some fiddeling and this works. You need to make sure `fontColor` is already set, since if there is a hover font  color set, then we set it to `''` it is not overridden.

src/table/styling-utils.js
LiKang6688: important is necessary. Otherwise, in sense-client, the hover effect skips every other row.

src/table/styling-utils.js
LiKang6688: I see the point. I thought when there is no font color and no hover font color but a hover background color, the hover font color becomes black when the hover background color is a light color. 
But the hover font color stays the same as the font color when the hover background color is a light color.
Thanks.

src/table/styling-utils.js
haxxmaxx: I have a fix for it in sense-client. It is a weird css rule for the old table that is applied to sn-table as well: https://github.com/qlik-trial/sense-client/pull/15161

haxxmaxx:
Logic and the descriptive comments look good. Some testing then we can merge this. No need to test isDarkColor for now

src/table/styling-utils.js
LiKang6688: Nice.

LiKang6688:
> Logic and the descriptive comments look good. Some testing then we can merge this. No need to test isDarkColor for now

Okay, I am writing some tests

test/hover-test-cases.json
```diff
@@ -0,0 +1,112 @@
+// 1. defalut
```
haxxmaxx, line 1: Planning on using this file?

LiKang6688: used for test case reference. if you are not interested in it, i will just remove it from here.

haxxmaxx: Then remove it, we don't want any unused code :)

haxxmaxx (approved):
Remove the test cases file and this is fine for now. I noticed that the selection performance is significantly worse with this added. Not sure why but I have some ideas

test/hover-test-cases.json
LiKang6688: Done

haxxmaxx: go ahead and merge :)

PR #108, opened 2021-02-22
Fix similar code checking in codeclimate

LiKang6688:
similar-code threshold describes the mass code block to be analyzed for duplication.
Refer https://github.com/codeclimate/codeclimate-duplication
for this settings.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,8 +4,8 @@ plugins:
     enabled: true
     config:
       languages:
-        javascript:
-          count_threshold: 3
+        filters:
+          - "(object (Identifier PropTypes))"
 checks:
   method-lines:
     config:
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,4 +1,11 @@
 version: "2"
+plugins:
+  duplication:
+    enabled: true
+    config:
+      languages:
+        javascript:
+          count_threshold: 3
 checks:
   method-lines:
     config:
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,11 +1,11 @@
 version: "2"
 plugins:
   duplication:
-    enabled: true
     config:
       languages:
-        filters:
-          - "(object (Identifier PropTypes))"
+        javascript:
+          filters:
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -1,4 +1,11 @@
 version: "2"
+plugins:
```
haxxmaxx, line 2: Does this need to be under `plugin`? Does it work putting it under `checks`, as the other settings?

LiKang6688: Following the Configuring in https://github.com/codeclimate/codeclimate-duplication, it needs to be under plugin. We can use Code Climate CLI (https://github.com/codeclimate/codeclimate) to do analyze locally to see if it works as we want.

haxxmaxx:
I just tried this rule on this branch https://github.com/qlik-oss/sn-table/pull/109. With the threshold at 3 it does not complain about `styling-utils`, which I think is a bit too weak. Those two functions could be replaced with one function taking one extra argument, which is something I think code climate should catch, but then you also get the false positives, like the propTypes.

.codeclimate.yml
```diff
@@ -1,4 +1,11 @@
...
+  duplication:
+    enabled: true
+    config:
+      languages:
+        filters:
+          - "(object (Identifier PropTypes))"
```
LiKang6688, line 8: Based on https://github.com/codeclimate/codeclimate-duplication/issues/89

haxxmaxx: Would be really nice if this works, but I still see 6 fixed issues on codeclimate, there should only be the two propTypes that are fixed

LiKang6688: okay, look into it

LiKang6688: now it is correct

haxxmaxx: I don't know why, but if you look at codeclimate it is saying that the propTypes duplications are fixed, then there are two new issues that are exactly the same... We can try to merge this and see what it looks like when ir runs on main

LiKang6688: sure

haxxmaxx (approved):

PR #109, opened 2021-02-23
fix: line height

### LINKED ISSUE #80: styling panel font size
should be able to consume font sizes set in styling panel, both header and content

haxxmaxx:
Fixes #80

Two more or less identical functions, but waiting with refactoring until we are done with all properties.


LiKang6688 (approved):

haxxmaxx:
I'll wait with this until we merge the hover styling, so I can set default constants for those values as well

LiKang6688 (approved):
Looks good to me.

haxxmaxx:
> I would say it is better to commit to a new PR as the commit here is not relevant to the current PR.

Hmm yeah you're right. I'll actually revert it for better tracking

LiKang6688:
> > I would say it is better to commit to a new PR as the commit here is not relevant to the current PR.
> 
> Hmm yeah you're right. I'll actually revert it for better tracking
👍

PR #111, opened 2021-02-25
table body performance

LiKang6688:
react skip render TableBodyWrapper when no  prop changes and add dependencies to use Memo otherwise a new value will be computed on every render


haxxmaxx (approved):
Really nice, I've seen that it re-renders more than needed. Just make sure the unit tests pass

LiKang6688:
> Really nice, I've seen that it re-renders more than needed. Just make sure the unit tests pass

Thanks. Fixed.

PR #112, opened 2021-02-26
feat: column coloring

### LINKED ISSUE #83: column conditional coloring
Able to read column coloring, could be expression or not. This is where we start using unique column renderers

haxxmaxx:
fixes #83

New approach to setting styling: calculating the base styling, then updating it depending on selections and column styling, then actually setting it in the end. Now there is no hassle with class priority.


src/table/cells/__tests__/withStyling.spec.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+describe('withStyling', () => {
+  const sandbox = sinon.createSandbox();
+  let HOC;
+  let styling;
+
+  beforeEach(() => {
+    // This is a bit of dummy mock. value will be on props, but this is not how the value is set. That is done in tableBodyWrapper
+    // But this enables testing that withStyling uses the CellComponent correctly
+    HOC = withStyling.default((props) => <div {...props}>{props.value}</div>);
```
haxxmaxx, line 17: Not the best test strategy... The thing is that checking if an element has a certain class using RTL is bad practice. You should only look at what the user could see. So all we don here is pass this mocked div through `withStyling` and check if it comes out the other side. For `withColumnStyling` we at least test if `getColumnStyle` is called properly, so that test makes sense. I'm 50/50 about keeping this, doesn't test anything properly, but keeps the coverage up 🤷

src/table/styling-utils.js
```diff
@@ -55,15 +69,44 @@ export function getBodyStyle(layout, theme) {
...
+    }
+
+    return SELECTION_STYLING.POSSIBLE;
+  }
+
+  return {};
+}
+
+export function getSelectionStyle(styling, cell, selState) {
+  return { ...styling, ...getSelectionColors(cell, selState) };
```
haxxmaxx, line 111: wrapping it like this lowers the complexity and makes it a bit easier to test

Caele (approved):

LiKang6688 (approved):

PR #113, opened 2021-02-26
Use arrow keys to navigate between cells

LiKang6688:


LiKang6688:
if it is okay, then I will write some tests for functions in handleKeyPress.js

src/table/TableBodyWrapper.jsx
```diff
@@ -51,29 +51,33 @@ const TableBodyWrapper = ({ tableData, constraints, selectionsAPI, layout, theme
...
     <TableBody>
-      {rows.map((row) => (
+      {rows.map((row, rowIndex) => (
         <TableRow
           hover={hoverEffect}
           role="checkbox"
           tabIndex={-1}
           key={row.key}
-          className={hoverEffect && classes.hoverTableRow}
+          className={`sn-table-row ${hoverEffect} && ${classes.hoverTableRow}`}
```
haxxmaxx, line 60: i think the string becomes `sn-table-row true && hoverTableRow`, think should be

```suggestion
          className={`sn-table-row ${hoverEffect && classes.hoverTableRow}`}
```

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
+const navigationEffect = (tableRows, nextRow, nextCol) => {
+  const nextCell = tableRows[nextRow].getElementsByClassName('sn-table-cell')[nextCol];
```
haxxmaxx, line 2: I know my prototype relied on classnames, but can it be done by using tags instead? `td` for the content and `th` for the head

src/table/TableBodyWrapper.jsx
LiKang6688: you are right. Will fix. Thanks.

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
+const navigationEffect = (tableRows, nextRow, nextCol) => {
```
haxxmaxx, line 1: should probably be called something like `moveFocus`

LiKang6688: sounds good. sure.

src/table/cells/selections.jsx
```diff
@@ -17,20 +18,30 @@ const useStyles = makeStyles({
...
     const handleMouseUp = (evt) => cell.isDim && evt.button === 0 && selectCell(cell, selState, selDispatch, evt);
     const selectionClass = getSelectionClass(cell, selState);
 
-    return <CellComponent {...props} className={`${classes[selectionClass]} ${className}`} onMouseUp={handleMouseUp} />;
+    return (
+      <CellComponent
+        {...props}
+        className={`${classes[selectionClass]} ${className}`}
+        onMouseUp={handleMouseUp}
+        onKeyDown={(evt) => handleKeyPress(evt, rootElement, rowIndex, columnIndex, cell, selState, selDispatch)}
```
haxxmaxx, line 32: So this is the part that needs some thinking. I'm not sure we want to place the listener on the cell. A lot of passing arguments and we need to make sure it works with a bunch of different cases. (selections enabled/disabled, column styling etc). If we put the listener on the entire table, we don't have that much information about the cell and have to fetch all cell using `getElementsByClassName` or similar. But I think it would be a simpler solution, then add a onKeyDown on this cell for clicking enter for selections or something. Not sure which approach to take at this point

src/table/cells/handleKeyPress.js
LiKang6688: Personally, I think both getElementsByClassName and getElementsByTagName are okay. Do you have any specific reason to prefer tag?

haxxmaxx: Not needing to add extra classes to the components. All classnames from mui are renamed to a unique code, like `1614601815238-sn-t44` (1614601815238 is a unique seed for that chart, sn-t for sn-table, then a number to identify which class it is). Not that pretty combining with something like `sn-table-cel'` since it wont be converted. That being said, I looked around and it seems fine to mix classes like this. I just think we can avoid adding the classes since they just correspond to a tag anyways

src/table/cells/selections.jsx
LiKang6688: Yeah, right. I was working on selections as well and decided to place the listener on the cell as it is a much easier solution.

haxxmaxx: If selections are enabled, I agree. But we need to consider the case where you want the keyboard navigation and selections disabled.

src/table/cells/handleKeyPress.js
LiKang6688: I do not think adding extra class names affects anything, which means avoiding adding the classes does not get any benefits.  
And adding class names and getting elements by those class names are a kind of common practice. 
I argue that it is compensated conciseness for its readability and understandability.

src/table/cells/selections.jsx
LiKang6688: Yeah, I will work on **Cancels Selections** in key selections PR as well. 
And in which case do we want the keyboard navigation enabled and selections disabled? IMO, in edit mode, we want both the keyboard navigation and selections disabled. In analysis mode, we want both enabled. Could you correct me If I am wrong?

src/table/cells/handleKeyPress.js
haxxmaxx: good points, let's stick with classes!

src/table/cells/selections.jsx
haxxmaxx: That is true, but there might be cases outside sense. @Caele what do you think about this?

Caele: hmm, I'd have to look more closely, but I think:
- only the focused cell should have a listener (which is added dynamically upon enter)
- moving with the keys also moves the listener (no react re-render involved?)
This kind of skips the entire React flow, but would save us lots of listeners and possible re-renders if we can pull it off. But I'm also not very fluent in react...

haxxmaxx: That's an interesting approach. In that case, where would you handle that? Still in selections? or should it be available when selections are disable?

Caele: In theory it would be handled in a navigation component thing instead, so separate from selections. (As you could have navigation without selections, in snapshot + screenreader mode).
I did check react, and it adds a listener to the document, not each element. (They will change to the root of the react app, but should be the same for us).

As you enter the table, the state should be row 0, col 0, so that cell recives focus. Then we can update and track that as the arrow keys are pressed.
I'm not entirely sure how to best do it, but I think we need to try and avoid re-rendering as you move around. There is really no need as all we do is change the focus().

haxxmaxx: `I did check react, and it adds a listener to the document, not each element. (They will change to the root of the react app, but should be the same for us).`
So when you add a listener for a bunch of elements, they are all bunched into one big listener?

`I'm not entirely sure how to best do it, but I think we need to try and avoid re-rendering as you move around.` I don't think it rerenders with the current solution. If we start adding states, especially on a higher level, then it might update the entire table.

src/table/TableBodyWrapper.jsx
```diff
@@ -69,6 +70,8 @@ const TableBodyWrapper = ({ tableData, constraints, selectionsAPI, layout, theme
                   styling={styling}
                   selState={selState}
                   selDispatch={selDispatch}
+                  tabIndex={-1}
+                  onKeyDown={(evt) => handleKeyPress(evt, rootElement, rowIndex + 1, columnIndex)}
```
haxxmaxx, line 74: I just looked at the old table and realised that you can always navigate around it the table, at least when you attain focus by clicking a cell. So always having it here is what we should do. If we, down the line, decide to be able to disable the navigation, that's future problems :)

haxxmaxx: No need to do that now. The whole focus in and out of the table depends on how we solve it in nebula. This is fine for now :)

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+  nextCell.setAttribute('tabIndex', '0');
+};
+
+const arrowKeysNavigation = (evt, rowAndColumn, rowIndex, colIndex) => {
+  let nextRow = rowIndex;
+  let nextCol = colIndex;
+
+  switch (evt.key) {
+    case 'ArrowDown':
+      if (rowIndex + 1 < rowAndColumn.tableRowSize) nextRow = ++nextRow;
```
haxxmaxx, line 13:
```suggestion
      if (rowIndex + 1 < rowAndColumn.tableRowSize) nextRow++;
```

This is all you need to do, it is the same thing as `nextRow = nextRow + 1`

LiKang6688: right, gonna change. thanks.

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+  switch (evt.key) {
+    // TODO page up/down etc
+    case 'ArrowUp':
+    case 'ArrowDown':
+    case 'ArrowRight':
+    case 'ArrowLeft': {
+      removeFocus(evt);
+      const rowAndColumn = tableRowAndColumn(rootElement);
+      const { tableRows, nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumn, rowIndex, colIndex);
+      moveFocus(tableRows, nextRow, nextCol);
```
haxxmaxx, line 65: You already have `tableRows` stored in `rowAndColumn` so no need to return that from `arrowKeysNavigation`.

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+const handleKeyPress = (evt, rootElement, rowIndex, colIndex) => {
+  preventDefaultBehavior(evt);
+  switch (evt.key) {
+    // TODO page up/down etc
+    case 'ArrowUp':
+    case 'ArrowDown':
+    case 'ArrowRight':
+    case 'ArrowLeft': {
+      removeFocus(evt);
+      const rowAndColumn = tableRowAndColumn(rootElement);
```
haxxmaxx, line 63: It would be really nice if these could be stored so we don't have to get them for every arrow key press. But I don't have a good idea of how to do it, so this is fine for now.

haxxmaxx: All though I think it should be renamed to `getRows` since it is a getter

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const tableRowAndColumn = (rootElement) => {
+  const tableRows = rootElement.getElementsByClassName('sn-table-row');
+  const tableRowSize = tableRows.length;
+
+  const headCells = rootElement.getElementsByClassName('sn-table-head-cell');
+  const tableColumnSize = headCells.length;
+
+  return { tableRows, tableRowSize, tableColumnSize };
```
haxxmaxx, line 41: And no need to call these `table...` since we know it is in the table.
```suggestion
  return { rows, rowCount, rowWidth };
```

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+  preventDefaultBehavior(evt);
+  switch (evt.key) {
+    // TODO page up/down etc
+    case 'ArrowUp':
+    case 'ArrowDown':
+    case 'ArrowRight':
+    case 'ArrowLeft': {
+      removeCurrentFocus(evt);
+      const rowAndColumn = getRowAndColumn(rootElement);
+      const { tableRows, nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumn, rowIndex, colIndex);
```
haxxmaxx, line 64: Still no need to return `tableRows` from `arrowKeysNavigation`

src/table/cells/handleKeyPress.js
```diff
@@ -10,16 +10,16 @@ const arrowKeysNavigation = (evt, rowAndColumn, rowIndex, colIndex) => {
 
   switch (evt.key) {
     case 'ArrowDown':
-      if (rowIndex + 1 < rowAndColumn.tableRowSize) nextRow = ++nextRow;
+      if (rowIndex + 1 < rowAndColumn.tableRowSize) nextRow++;
```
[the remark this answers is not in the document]
LiKang6688: Good point. Thanks

src/table/cells/handleKeyPress.js
LiKang6688: yeah, sure

src/table/TableBodyWrapper.jsx
```diff
@@ -34,6 +34,13 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
     colIdx: -1,
     isEnabled: !!selectionsAPI && !constraints.active,
   });
+  const handleMouseUp = (evt, cell) => {
```
haxxmaxx, line 37: Good that you started on this, but it should not be in this PR since it is not part of the issue.

src/table/cells/withSelections.jsx
```diff
@@ -6,10 +6,10 @@ import { getSelectionStyle } from '../styling-utils';
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
     const { cell, selState, selDispatch, styling } = props;
-    const handleMouseUp = (evt) => cell.isDim && evt.button === 0 && selectCell(cell, selState, selDispatch, evt);
+    const handleMouseDown = (evt) => cell.isDim && evt.button === 0 && selectCell(cell, selState, selDispatch, evt);
     const selectionStyling = useMemo(() => getSelectionStyle(styling, cell, selState), [cell, selState]);
 
-    return <CellComponent {...props} styling={selectionStyling} onMouseUp={handleMouseUp} />;
+    return <CellComponent {...props} styling={selectionStyling} onMouseDown={handleMouseDown} />;
```
haxxmaxx, line 12: This should be mouseUp

LiKang6688: because of const handleMouseUp = (evt, cell) => {}, gonna change back

haxxmaxx:
Looks good! just add tests and we can merge this

LiKang6688:
> Looks good! just add tests and we can merge this

Thanks. Gonna write tests.

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -0,0 +1,85 @@
...
+
+  describe('getRowAndColumnCount', () => {
+    let resolvedRowCount;
+    let resolvedColumnCount;
+
+    it('should return a rowElements, rowCount, and columnCount', () => {
+      resolvedRowCount = 15;
+      resolvedColumnCount = 2;
+      JSDOM.fromFile('src/table/cells/__tests__/table.html').then((dom) => {
+        const rootElement = dom.window.document;
```
haxxmaxx, line 78: I'm not sure about this. One of the perks of isolating the `getRowAndColumnCount` function is that you can pass a mocked `rootElement` with mocked functions on it, returning some number of rows/columns then checking if `getRowAndColumnCount` returns the same

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  return {
+    nextRow,
+    nextCol,
+  };
+};
+
+export const getRowAndColumnCount = (rootElement) => {
+  window.console.log(rootElement, 'rootElement');
```
haxxmaxx, line 34: remove log

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -0,0 +1,85 @@
...
+describe('handle-key-press', () => {
+  describe('moveToNextFocus', () => {});
+
+  describe('arrowKeysNavigation', () => {
+    const evt = {};
+    const rowAndColumnCount = {};
+    let rowIndex;
+    let colIndex;
+
+    it('should stay the current cell', () => {
```
haxxmaxx, line 14: Technically this can happen for all directions. If you want to you can add three more tests for this, but this is fine

haxxmaxx, line 14: Technically this can happen for all directions. If you want to you can add three more tests for this, but this is fine

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -0,0 +1,85 @@
...
+const { JSDOM } = require('jsdom');
+
+describe('handle-key-press', () => {
+  describe('moveToNextFocus', () => {});
+
+  describe('arrowKeysNavigation', () => {
+    const evt = {};
+    const rowAndColumnCount = {};
+    let rowIndex;
+    let colIndex;
```
haxxmaxx, line 12: I think you should add a `beforeEach`. That way you make sure everything is reset before each test plus you only need to change one or a few variables per test.

src/table/cells/handleKeyPress.js
```diff
@@ -0,0 +1,74 @@
...
+  evt.target.blur();
+  evt.target.setAttribute('tabIndex', '-1');
+};
+
+export const preventDefaultBehavior = (evt) => {
+  evt.stopPropagation();
+  evt.preventDefault();
+};
+
+const handleKeyPress = (evt, rootElement, rowIndex, colIndex) => {
```
haxxmaxx, line 54: The simplest approach to testing this would be to mock all the methods on `evt` and `rootElement` then just make one test for one of the arrows and making sure everything is called (`preventDefault`, `target.blur` etc). This would automatically cover `preventDefaultBehavior`, `removeCurrentFocus` and `moveToNextFocus` as well.

src/table/cells/__tests__/handle-key-press.spec.js
LiKang6688: good point. Fixed.

src/table/cells/__tests__/handle-key-press.spec.js
LiKang6688: good point. Fixed.

src/table/cells/__tests__/handle-key-press.spec.js
LiKang6688: I think 'passing a mocked rootElement with a mocked method on it, returning some number of rows/columns then checking if getRowAndColumnCount returns the same' is what I did here.

src/table/cells/handleKeyPress.js
LiKang6688: Thanks. Working on it.

src/table/cells/handleKeyPress.js
LiKang6688: fixed

src/table/cells/__tests__/handle-key-press.spec.js
haxxmaxx: Yes, but you don't need to create a whole DOM. We know what `getElementsByClassName` should return, given a certain DOM. You can just create an object with the `getElementsByClassName` mocked returning an array of "elements", the same way the real method would

LiKang6688: I think that if I create an object - rootElement, with a mocked getElementsByClassName which returns rowElements. Then it only really test the length in the getRowAndColumnCount function.

haxxmaxx: Well, not only that. The returned object should also contain the same thing as `getElementsByClassName` returns, plus both of the lengths of the arrays. Adding a whole DOM for a unit test, then it becomes a bit more of an integration test or something. Since this feature does a lot of DOM manipulations we should have integration tests for keyboard navigation to complement the unit tests

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -0,0 +1,150 @@
...
+    });
+
+    afterEach(() => {
+      sandbox.verifyAndRestore();
+    });
+
+    it('should stay current focus and focus the next cell', () => {
+      handleKeyPress.default(evt, rootElement, rowIndex, colIndex);
+      expect(handleKeyPress.default).to.have.been.calledOnce;
+      expect(handleKeyPress.preventDefaultBehavior).to.have.been.calledOnce;
```
LiKang6688, line 135: I got AssertionError: expected spy to have been called exactly once, but it was called 0 times.

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    });
+
+    afterEach(() => {
+      sandbox.verifyAndRestore();
+    });
+
+    it('should stay current focus and focus the next cell', () => {
+      handleKeyPress.default(evt, rootElement, rowIndex, colIndex);
+      expect(handleKeyPress.default).to.have.been.calledOnce;
+      expect(evt.stopPropagation).to.have.been.calledOnce;
```
LiKang6688, line 132: AssertionError: expected spy to have been called exactly once, but it was called 0 times

haxxmaxx (approved):
🎉

LiKang6688:
> 🎉

Thanks!

PR #118, opened 2021-02-28
Selections with keys

LiKang6688:
works on #9


src/table/selections-utils.js
```diff
@@ -62,6 +62,9 @@ export function selectCell(cell, selState, selDispatch, evt) {
 
   if (selState.colIdx === -1) {
     api.begin('/qHyperCubeDef');
+    setTimeout(() => {
```
haxxmaxx, line 65: What is the timeout needed for?

src/table/TableBodyWrapper.jsx
```diff
@@ -71,7 +71,9 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
                   selState={selState}
                   selDispatch={selDispatch}
                   tabIndex={-1}
-                  onKeyDown={(evt) => handleKeyPress(evt, rootElement, rowIndex + 1, columnIndex)}
+                  onKeyDown={(evt) =>
```
haxxmaxx, line 74: Need to make sure this only runs when selections are enabled. I'm a bit split on what is best here. Either we add a check on this listener or a separate listener on `withSelections`. It is nice too keep everything on the same listener, but then all selection stuff are not contained in `withSelections`. I think it is fine though, as long as we sanity check for `selectionsEnabled` in a nice way

src/table/TableBodyWrapper.jsx
```diff
@@ -71,7 +71,10 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
                   selState={selState}
                   selDispatch={selDispatch}
                   tabIndex={-1}
-                  onKeyDown={(evt) => handleKeyPress(evt, rootElement, rowIndex + 1, columnIndex)}
+                  onKeyDown={
+                    (evt) => handleKeyPress(evt, rootElement, rowIndex + 1, columnIndex, cell, selState, selDispatch)
+                    // eslint-disable-next-line react/jsx-curly-newline
```
haxxmaxx, line 76: I think it wants you to do something like this
```suggestion
                    (evt) =>
                      handleKeyPress(evt, rootElement, rowIndex + 1, columnIndex, cell, selState, selDispatch)
```

src/table/cells/handle-key-press.js
```diff
@@ -50,21 +52,37 @@ export const preventDefaultBehavior = (evt) => {
...
 
-const handleKeyPress = (evt, rootElement, rowIndex, colIndex) => {
-  preventDefaultBehavior(evt);
+const handleKeyPress = (evt, rootElement, rowIndex, colIndex, cell, selState, selDispatch) => {
   switch (evt.key) {
-    // TODO page up/down etc
-    case 'ArrowUp':
-    case 'ArrowDown':
+    case 'ArrowUp': // Up arrow:  Navigates to the row above.
+    case 'ArrowDown': // Down arrow: Navigates to the row below.
```
haxxmaxx, line 58: Id say what arrow down/up does is self-explanatory, no need for comments

src/table/selections-utils.js
```diff
@@ -80,3 +80,13 @@ export function selectCell(cell, selState, selDispatch, evt) {
...
 }
+
+export const confirmSelections = (selState) => {
+  const { api } = selState;
+  api.confirm();
+};
+
+export const cancelSelections = (selState) => {
+  const { api } = selState;
+  api.cancel();
```
haxxmaxx, line 91: If this is the only thing these functions are suppose to do, they are not needed. just call `selState.api.confirm` from `handleKeyPress`

src/table/TableBodyWrapper.jsx
LiKang6688: Screenshot 2021-03-17 at 12 46 06
it is same. dot not know why

src/table/selections-utils.js
LiKang6688: sure

src/table/TableBodyWrapper.jsx
haxxmaxx: I'm not getting it, have you run `yarn` in a while? some linting stuff might be updated. `yarn lint` is not complaining with my suggestion so let's go with that

haxxmaxx:
Is it possible to navigate left/right when selecting? That should not be possible

src/table/TableBodyWrapper.jsx
LiKang6688: fixed after running yarn. thanks

LiKang6688:
> Is it possible to navigate left/right when selecting? That should not be possible

good point. look at it now.

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -97,27 +105,71 @@ describe('handle-key-press', () => {
...
+        api: {
+          confirm: sinon.spy(),
+          cancel: sinon.spy(),
+        },
+      };
+      cell = [];
+    });
+
+    describe('ArrowDown', () => {
+      it('should prevent default behavior and remove current focus', () => {
```
haxxmaxx, line 146: ...and set it to the next cell. Test is solid, just the description is a bit misleading

src/table/cells/handle-key-press.js
```diff
@@ -50,21 +52,39 @@ export const preventDefaultBehavior = (evt) => {
...
       removeCurrentFocus(evt);
       const rowAndColumnCount = getRowAndColumnCount(rootElement);
       const { nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumnCount, rowIndex, colIndex);
       moveToNextFocus(rowAndColumnCount.rowElements, nextRow, nextCol);
       break;
     }
-    // TODO handle selections, search?, sorting...
+    // Space bar: Selects value.
+    case ' ': {
+      preventDefaultBehavior(evt);
```
haxxmaxx, line 70: tests missing on this case, on for dim, one for measure.

src/table/cells/handle-key-press.js
```diff
@@ -1,25 +1,29 @@
...
+  //  when selecting cell, disable navigating left/right and going into header
+  const isSelectedTable = selState && selState.rows.length > 0;
 
   switch (evt.key) {
     case 'ArrowDown':
       rowIndex + 1 < rowAndColumnCount.rowCount && nextRow++;
       break;
     case 'ArrowUp':
-      rowIndex > 0 && --nextRow;
+      rowIndex > 0 && (!isSelectedTable || !(rowIndex === 1)) && --nextRow;
```
haxxmaxx, line 20:
```suggestion
      rowIndex > 0 && (!isSelectedTable || rowIndex !== 1) && --nextRow;
```

src/table/cells/handle-key-press.js
```diff
@@ -50,21 +54,37 @@ export const preventDefaultBehavior = (evt) => {
...
-      const { nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumnCount, rowIndex, colIndex);
+      const { nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumnCount, rowIndex, colIndex, selState);
       moveToNextFocus(rowAndColumnCount.rowElements, nextRow, nextCol);
       break;
     }
-    // TODO handle selections, search?, sorting...
+    // Space bar: Selects value.
+    case ' ': {
+      preventDefaultBehavior(evt);
+      cell?.isDim && selectCell(cell, selState, selDispatch, evt);
```
haxxmaxx, line 73: Do we evet check if selections are enabled? `selState.isEnabled`

src/table/cells/handle-key-press.js
LiKang6688: fixed

src/table/cells/handle-key-press.js
LiKang6688: no, we don't. 
I do not think that it is needed here. As when selection is not enabled, it is impossible to get focus on a cell and use the space bar to do the selection.

haxxmaxx: Just tried it, you still can. In edit mode for example. But that is because of the hacky solution that the first cell has tabindex 0. As soon as you have focus on a cell, you can do whatever you want. I think we will start with only being able to tab into the chart in a proper way (working on it. now), and have all the functionality available when navigating. So this is fine for now

haxxmaxx (approved):

src/table/cells/handle-key-press.js
LiKang6688: I agree. Thanks.

PR #122, opened 2021-03-08
feat: auto font color

### LINKED ISSUE #120: auto color font when background is set
When the background is set but not the font color, the font color should adapt to be as easy to see as possible.
background dark -> white font
background light -> dark gray.

This goes both for hover and column styling. For consistency I think we should always override the font color, if it is not set on the same level.

haxxmaxx:
fixes #120


haxxmaxx:
Both the css colors and `resolveColor` should optimally be stored in nebula, I have added an issue for that. Because of that I think we should leave `color-utils` untested.

Caele:
> Both the css colors and `resolveColor` should optimally be stored in nebula, I have added an issue for that. Because of that I think we should leave `color-utils` untested.

Maybe, don't want to put too many specifics into Nebula. So could see this as a separate library instead. But if we want to solve for less duplicated code (size wise) then we need it in Nebula. We have theme stuff there already, so some color exposing is not out of the question.

Caele (approved):

haxxmaxx:
> > Both the css colors and `resolveColor` should optimally be stored in nebula, I have added an issue for that. Because of that I think we should leave `color-utils` untested.
> 
> Maybe, don't want to put too many specifics into Nebula. So could see this as a separate library instead. But if we want to solve for less duplicated code (size wise) then we need it in Nebula. We have theme stuff there already, so some color exposing is not out of the question.

It would make sense to me. the issue if you haven't seen it https://github.com/qlik-oss/nebula.js/issues/577

PR #123, opened 2021-03-08
feat: hide column

### LINKED ISSUE #84: column hide/show
Able to read show/ hide columns. Show is of course default

haxxmaxx:
fixes #84

### DIFF
--- a/src/table/handle-data.js
+++ b/src/table/handle-data.js
@@ -11,14 +11,17 @@ export function getColumnInfo(layout, colIndex) {
   const numDims = qDimensionInfo.length;
   const isDim = colIndex < numDims;
   const info = isDim ? qDimensionInfo[colIndex] : qMeasureInfo[colIndex - numDims];
-  return {
-    isDim,
-    width: 200,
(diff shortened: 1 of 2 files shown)


src/table/handle-data.js
```diff
@@ -28,7 +31,8 @@ export default async function manageData(model, layout, pageInfo) {
   ]);
   const matrix = dataPages[0].qMatrix;
 
-  const columns = columnorder.map((c) => getColumnInfo(layout, c));
+  // using filter to remove hidden columns (represented with false)
+  const columns = columnorder.map((c) => getColumnInfo(layout, c)).filter(Boolean);
```
LiKang6688, line 35: I look at the filter(Boolean) trick. It is interesting. It is very good that you write comments here.

LiKang6688 (approved):

src/table/handle-data.js
haxxmaxx: Agreed, it is not common syntax. Found it on stack overflow, someone else being the smart one here 🙃

PR #125, opened 2021-03-11
fix: properly wait for rendering

Caele:
The table wasn't properly returning the render promise due to how the load data is setup. Adding a promuse hook fixes this.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,9 +38,7 @@ export default function supernova(env) {
       const [pageInfo, setPageInfo] = useState({ top: 0, height: 100 });
       const [muiParameters] = useState(muiSetup(__OPIONAL_THEME_DEPS__));
 
-      const [tableData] = usePromise(() => {
-        return manageData(model, layout, pageInfo);
-      }, [layout, pageInfo]);
+      const [tableData] = usePromise(() => manageData(model, layout, pageInfo), [layout, pageInfo]);
 
       useEffect(() => {
         if (layout && tableData) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -35,14 +36,9 @@ export default function supernova(env) {
...
       const [pageInfo, setPageInfo] = useState({ top: 0, height: 100 });
-      const [tableData, setTableData] = useState();
       const [muiParameters] = useState(muiSetup(__OPIONAL_THEME_DEPS__));
 
-      useEffect(() => {
-        manageData(model, layout, pageInfo).then((d) => {
-          setTableData(d);
-        });
-      }, [layout, pageInfo]);
+      const [tableData] = usePromise(() => manageData(model, layout, pageInfo), [layout, pageInfo]);
```
haxxmaxx, line 41: Don't remember how `usePromise` works. `tableData` will update when `layout` or `pageInfo` changes?

haxxmaxx (approved):

src/index.js
Caele: hmnm, didn't acutally test, seems not, I'll check

Caele: yea it works as expected afaik

LiKang6688: I think it workss the same as before but more concisely

LiKang6688 (approved):

src/index.js
LiKang6688: https://qlik.dev/apis/javascript/nebulajs-stardust#%23%2Fentries%2FusePromise the example here is not the best I would say

PR #127, opened 2021-03-15
feat: sorting

### LINKED ISSUE #4: US: Sorting
No need for subtasks in this one since it is only one functionality we want to add:

Ability to change the sorting order by clicking column headers. Whichever column is pressed latest will be the one that we fort first on. Pressing the header again should toggle ascending/descending. Same as the old table.

This should be done with a patch back-end call (soft properties). In sense the property panel should handle that and display the discard/confirm dialog when going back to edit mode. Outside sense the changes are simply discarded after the session ends.

haxxmaxx:
fixes #4

There is weird lag for column styling that is dependent on the cell value. The values are updated before the the styling is. We should consider if it is good enough before starting to look at it IMO.


src/index.js
```diff
@@ -37,8 +39,8 @@ export default function supernova(env) {
 
       const [pageInfo, setPageInfo] = useState({ top: 0, height: 100 });
       const [muiParameters] = useState(muiSetup(__OPIONAL_THEME_DEPS__));
-
       const [tableData] = usePromise(() => manageData(model, layout, pageInfo), [layout, pageInfo]);
+      const changeSortOrder = useMemo(() => sortingFactory(model, layout), [layout]);
```
Caele, line 43: Unsure about using useMemo here, think you could just as well call it directly.

Caele (approved):
Looks good

src/index.js
Caele: Also, layout could come from the TableHeader, its only model that needs to be hoisted to the function.

haxxmaxx: Trying to avoid fetching if I don't need to, but it probably don't make any difference

PR #129, opened 2021-03-17
add Selects multiple values

LiKang6688:
works on #8

### DIFF
--- a/src/table/selections-utils.js
+++ b/src/table/selections-utils.js
@@ -80,3 +80,36 @@ export function selectCell(cell, selState, selDispatch, evt) {
     api.cancel();
   }
 }
+
+export const doSelectCell = (evt, rowValues, rowIndex, colId, selState, selDispatch) => {
+  const { api, rows } = selState;
+
+  let selectedRows = [];
+  let nextRowId;
+
+  switch (evt.key) {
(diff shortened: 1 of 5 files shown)


src/table/selections-utils.js
```diff
@@ -80,3 +80,36 @@ export function selectCell(cell, selState, selDispatch, evt) {
     api.cancel();
   }
 }
+
+export const doSelectCell = (evt, rowValues, rowIndex, colId, selState, selDispatch) => {
```
LiKang6688, line 84: not sure it is the correct way.  If yes, I will refactor it.

haxxmaxx: Is this for selecting multiple?

haxxmaxx: Ah, nevermind. This is a new PR...

LiKang6688: Yes, 
```
 if (keysPressed.Shift && (evt.key === 'ArrowUp' || evt.key === 'ArrowDown')) {
     doSelectCell(evt, rows, rowIndex, colId, selState, selDispatch);
 }
```

haxxmaxx:
This branch could be saved and be used if we implement this later, but the PR can be closed

Issue #128, opened 2021-03-15
use space/enter to change sorting


haxxmaxx:
When focusing on a header cell, pressing enter or space should update the sorting, the same way as when clicking. Need to make sure this can't be done everywhere, e.g. edit mode.

PR #132, opened 2021-03-18
feat: change sorting by pressing enter or space 

LiKang6688:
When focusing on a header cell, pressing enter or space should update the sorting #128

### DIFF
--- a/src/table/cells/__tests__/handle-key-press.spec.js
+++ b/src/table/cells/__tests__/handle-key-press.spec.js
@@ -1,9 +1,4 @@
-import handleKeyPress, {
-  getRowAndColumnCount,
-  arrowKeysNavigation,
-  moveFocus,
-  headHandleKeyPress,
-} from '../handle-key-press';
+import { getRowAndColumnCount, arrowKeysNavigation, bodyHandleKeyPress, headHandleKeyPress } from '../handle-key-press';
 
 describe('handle-key-press', () => {
   describe('arrowKeysNavigation', () => {
@@ -101,50 +96,7 @@ describe('handle-key-press', () => {
     });
   });
 
-  describe('moveFocus', () => {
-    let rowIndex;
-    let colIndex;
-    let evt = {};
-    let rootElement = {};
-    let selState = {};
-
-    beforeEach(() => {
-      rowIndex = 0;
-      colIndex = 0;
-      evt = {
-        key: 'ArrowDown',
-        stopPropagation: sinon.spy(),
-        preventDefault: sinon.spy(),
-        target: {
-          blur: sinon.spy(),
-          setAttribute: sinon.spy(),
-        },
-      };
-      rootElement = {
-        getElementsByClassName: () => [{ getElementsByClassName: () => [{ focus: () => {}, setAttribute: () => {} }] }],
-      };
-      selState = {
-        api: {
-          confirm: sinon.spy(),
-          cancel: sinon.spy(),
-          begin: sinon.spy(),
-          select: sinon.spy(),
-        },
-        rows: [],
-        colIdx: -1,
-      };
-    });
-
-    it('should prevent default behavior, remove current focus and set focus and attribute to the next cell', () => {
-      moveFocus(evt, rootElement, rowIndex, colIndex, selState);
-      expect(evt.preventDefault).to.have.been.calledOnce;
-      expect(evt.stopPropagation).to.have.been.calledOnce;
-      expect(evt.target.blur).to.have.been.calledOnce;
-      expect(evt.target.setAttribute).to.have.been.calledOnce;
-    });
-  });
-
-  describe('handleKeyPress', () => {
+  describe('bodyHandleKeyPress', () => {
     let rowIndex;
     let colIndex;
     let evt = {};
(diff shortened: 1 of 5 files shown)


haxxmaxx:
I don't see any of the selection navigation here. merge that and make sure you handle both cases :)

haxxmaxx:
My bad, they are in there 🙃

src/table/cells/handle-key-press.js
```diff
@@ -54,17 +54,47 @@ export const preventDefaultBehavior = (evt) => {
...
-      const { nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumnCount, rowIndex, colIndex, selState);
-      moveToNextFocus(rowAndColumnCount.rowElements, nextRow, nextCol);
+      changeSortOrder(layout, isDim, colIndex);
+      break;
+    }
+    // Enter: update the sorting
+    case 'Enter': {
+      preventDefaultBehavior(evt);
+      changeSortOrder(layout, isDim, colIndex);
+      break;
```
haxxmaxx, line 84: This is identical to space. Do as with the arrows and group them

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -192,4 +192,62 @@ describe('handle-key-press', () => {
...
+      changeSortOrder = sinon.spy();
+    });
+
+    describe('ArrowDown', () => {
+      it('should prevent default behavior, remove current focus and set focus and attribute to the next cell', () => {
+        headHandleKeyPress(evt, rootElement, rowIndex, colIndex);
+        expect(evt.preventDefault).to.have.been.calledOnce;
+        expect(evt.stopPropagation).to.have.been.calledOnce;
+        expect(evt.target.blur).to.have.been.calledOnce;
+        expect(evt.target.setAttribute).to.have.been.calledOnce;
```
haxxmaxx, line 229: You can just check if `moveFocus` is called here.

src/table/cells/handle-key-press.js
```diff
@@ -54,17 +54,47 @@ export const preventDefaultBehavior = (evt) => {
   evt.preventDefault();
 };
 
-const handleKeyPress = (evt, rootElement, rowIndex, colIndex, cell, selState, selDispatch) => {
+export const moveFocus = (evt, rootElement, rowIndex, colIndex, selState) => {
```
haxxmaxx, line 57: Yeah I'd say this should have seperate tests, basically the same as you have now for `handleKeyPress`. Only on tests is enough, and that gets all the smaller functions as well: `preventDefaultBehavior`, `removeCurrentFocus`, `moveToNextFocus`

src/table/cells/handle-key-press.js
LiKang6688: Right. Fixed. Thanks

src/table/cells/handle-key-press.js
LiKang6688: sure

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -192,4 +241,65 @@ describe('handle-key-press', () => {
...
+    });
+
+    afterEach(() => {
+      sandbox.verifyAndRestore();
+    });
+
+    describe('ArrowDown', () => {
+      it('should prevent default behavior, remove current focus and set focus and attribute to the next cell', () => {
+        headHandleKeyPress(evt, rootElement, rowIndex, colIndex);
+        expect(handleKeyPressObject.moveFocus).to.have.been.calledOnce;
```
LiKang6688, line 281: @haxxmaxx  I can not figure out why I got ' AssertionError: expected spy to have been called exactly once, but it was called 0 times' here. Could I have your help to have a look?

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -96,6 +101,49 @@ describe('handle-key-press', () => {
     });
   });
 
+  describe('moveFocus', () => {
```
haxxmaxx, line 104: Now this is redundant. It is both tested for `handleKeyDown` and `headHandleKeyDown` in the same way. You can remove it. Sorry for the back and forth, I thought it was easier to mock `moveFocus`

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -192,4 +240,62 @@ describe('handle-key-press', () => {
...
+          setAttribute: sinon.spy(),
+        },
+      };
+      rootElement = {
+        getElementsByClassName: () => [{ getElementsByClassName: () => [{ focus: () => {}, setAttribute: () => {} }] }],
+      };
+      changeSortOrder = sinon.spy();
+    });
+
+    describe('ArrowDown', () => {
```
haxxmaxx, line 271: No reason for a sepparate describe for every case, just add it to de `it` description

src/table/cells/handle-key-press.js
```diff
@@ -54,17 +54,42 @@ export const preventDefaultBehavior = (evt) => {
...
-      moveToNextFocus(rowAndColumnCount.rowElements, nextRow, nextCol);
+      changeSortOrder(layout, isDim, colIndex);
+      break;
+    }
+    default:
+      break;
+  }
+};
+
+const handleKeyPress = (evt, rootElement, rowIndex, colIndex, cell, selState, selDispatch) => {
```
haxxmaxx, line 86: I'd say this is not the default function anymore and should be renamed. Since the other one is `headHandleKeyPress` I suggest calling this `bodyHandleKeyPress`

src/table/cells/__tests__/handle-key-press.spec.js
LiKang6688: Fixed. Thanks

src/table/cells/__tests__/handle-key-press.spec.js
LiKang6688: Fixed. Thanks

src/table/cells/handle-key-press.js
LiKang6688: Fixed. Thanks

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -145,50 +223,22 @@ describe('handle-key-press', () => {
     });
 
     describe('SpaceBar', () => {
```
haxxmaxx, line 225: Still using described that are unnecessary

haxxmaxx (approved):
Refactor the describes like you did for the other handle function and this looks good!

PR #137, opened 2021-03-23
keyboard shortcuts to go to next and previous page

LiKang6688:


src/table/TableWrapper.jsx
```diff
@@ -51,6 +62,33 @@ export default function TableWrapper(props) {
...
+      updatePage(
+        event,
+        localState.current.totalRowSize,
+        localState.current.page,
+        localState.current.rowsPerPage,
+        setPageInfo,
+        setPage
+      )
+    );
+    window.addEventListener('keyup', handleKeyUp);
```
haxxmaxx, line 76: If you look at the keyDown event there are booleans for ctrl, meta and shift key. they are true when they are held down and pressing an other key. So no need for two listeners here.

src/table/TableWrapper.jsx
```diff
@@ -51,6 +62,33 @@ export default function TableWrapper(props) {
     return null;
   }
 
+  useEffect(() => {
+    window.addEventListener('keydown', (event) =>
```
haxxmaxx, line 66: Not sure we want to put the listener here. I was thinking it could use the one on the cells, but I guess you would still want it to work if you have focus on the rows per page dropdown for example. So i guess then it needs to be on the whole table. that being said, it should be the component, not on the window.

haxxmaxx: This also opens up the discussion if we should have the navigation and selections seperated on different listeners. We will have to have a listener on a higher level for this to work, it might be cleaner to have the navigation there as well. Let's stick to this for now, and maybe consider a refactor if we run in to trouble or something

src/table/TableWrapper.jsx
```diff
@@ -34,6 +35,16 @@ export default function TableWrapper(props) {
...
+  const localState = useRef();
+
+  useEffect(() => {
+    localState.current = {
+      totalRowSize: size.qcy,
+      page,
+      rowsPerPage,
+    };
+  }, [page, rowsPerPage]);
+
```
haxxmaxx, line 47: Why do you need this? Can't you use the `handleChangePage` function?

src/table/TableWrapper.jsx
LiKang6688: the keyup  is used to delete each element once their associated key has been released
```
const handleKeyUp = (event) => {
  delete keysPressed[event.key];
};
```

src/table/TableWrapper.jsx
LiKang6688: it is the way to get the latest component variable value in useEffect hook.

It involves the discussion where we want to put the listener as you mentioned above.

If it should be in the component, then I guess it is not needed.

src/table/TableWrapper.jsx
haxxmaxx: yeah but you don't have to store which keys are pressed. all that info is on the keydown event, e.g. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/ctrlKey

LiKang6688: Nice. Now I know that and did an update. Thanks.

haxxmaxx:
Tested this out and found some problems:

- The listener cannot be on the window, it should only catch key events from inside the table. I'd suggest using `onKeyDown` on the <Paper> element, then you are able to do stuff from the pagination footer as well. I tried it and it seems to use updated states, without adding new states or effects
- If you focus a cell, the other listener highjacks the arrow event. Need it to do nothing if ctrl and shift is pressed
- No need to redefine `handleChangePage`, just pass it from `TableWrapper`. Then it also has access to updated states

src/table/cells/handle-key-press.js
```diff
@@ -1,5 +1,25 @@
...
+  setPage(newPage);
+};
+
+export const updatePage = (event, totalRowSize, page, rowsPerPage, setPageInfo, setPage) => {
+  const pageSize = Math.floor(totalRowSize / rowsPerPage);
+  const leftRowSize = totalRowSize % rowsPerPage;
+  const isRightKeys = event.shiftKey && event.ctrlKey && event.key === 'ArrowRight';
+  if (
+    (isRightKeys && page + 1 < pageSize && pageSize > 0) ||
+    (isRightKeys && leftRowSize > 0 && page + 1 === pageSize)
```
haxxmaxx, line 14:
```suggestion
    (isRightKeys && page < pageSize)
```

src/table/cells/handle-key-press.js
```diff
@@ -1,5 +1,25 @@
...
 
+const handleChangePage = (newPage, setPageInfo, rowsPerPage, setPage) => {
+  setPageInfo({ top: newPage * rowsPerPage, height: rowsPerPage });
+  setPage(newPage);
+};
+
+export const updatePage = (event, totalRowSize, page, rowsPerPage, setPageInfo, setPage) => {
+  const pageSize = Math.floor(totalRowSize / rowsPerPage);
+  const leftRowSize = totalRowSize % rowsPerPage;
+  const isRightKeys = event.shiftKey && event.ctrlKey && event.key === 'ArrowRight';
```
haxxmaxx, line 11: As most applications, ctrl on windows corresponds to cmd on mac, which is `metaKey` on the event. So need to check that is either ctrl or meta

src/table/cells/handle-key-press.js
```diff
@@ -1,5 +1,18 @@
 import { selectCell } from '../selections-utils';
 
+const isCtrlShift = (evt) => evt.shiftKey && (evt.ctrlKey || evt.metaKey);
+
+export const updatePage = (evt, totalRowSize, page, rowsPerPage, handleChangePage) => {
+  if (isCtrlShift(evt)) {
+    const lastPage = Math.floor(totalRowSize / rowsPerPage);
+    if (evt.key === 'ArrowRight' && page < lastPage) {
```
LiKang6688, line 8: If we set 'Rows per page' to 10 and `totalRowSize` is 40, it does not work as expected.

haxxmaxx: true, fixed

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -235,5 +251,99 @@ describe('handle-key-press', () => {
...
+    });
+
+    it('handleChangePage should not run', () => {
+      evt.key = 'ArrowLeft';
+      page = 0;
+      updatePage(evt, totalRowSize, page, rowsPerPage, handleChangePage);
+      expect(handleChangePage).to.not.have.been.calledOnce;
+    });
+
+    it('should change page', () => {
```
haxxmaxx, line 316: you can remove this one, rows per page will never be 0

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -235,5 +251,99 @@ describe('handle-key-press', () => {
...
+      expect(handleChangePage).to.have.been.calledOnce;
+    });
+
+    it('should change page', () => {
+      totalRowSize = 40;
+      page = 0;
+      rowsPerPage = 30;
+      updatePage(evt, totalRowSize, page, rowsPerPage, handleChangePage);
+      expect(handleChangePage).to.have.been.calledOnce;
+    });
```
haxxmaxx, line 338: I'd say you only need one of them, they are basically testing the same thing

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -235,5 +251,99 @@ describe('handle-key-press', () => {
...
+      updatePage(evt, totalRowSize, page, rowsPerPage, handleChangePage);
+      expect(handleChangePage).to.not.have.been.calledOnce;
+    });
+
+    it('handleChangePage should not run', () => {
+      evt.key = 'ArrowLeft';
+      page = 0;
+      updatePage(evt, totalRowSize, page, rowsPerPage, handleChangePage);
+      expect(handleChangePage).to.not.have.been.calledOnce;
+    });
```
haxxmaxx, line 314: These two need unique names, describing why they should not run. Same thing for should change page below.
Also I think all arguments should be defined, totalRowSize and rowsPerPage are missing for the bottom one

haxxmaxx:
Consider adding a case for `TableWrapper.spec.js` where you send a keyboard event to see if updatePage runs

haxxmaxx:
Consider adding a case for `TableWrapper.spec.js` where you send a keyboard event to see if updatePage runs

haxxmaxx (approved):

PR #138, opened 2021-03-23
feat: focus on first cell from pressing enter

haxxmaxx:
So this is the first part of handling focus in and out of the table. When pressing enter on the nebula cell, the first cell in the table gets focus. Which cell has focus is stored as a state, so when you navigate back to the table, you get the same position.

Right now we only get an update when you press enter on the nebula cell, `a11y.focus` becomes true. We also need something that tells nebula that the table looses focus (working name `dropFocus`), which we would call on esc and shift+tab. Alternatively, nebula catches those events by it self. Additionally, there will be an enabled flag on the the `useAccessiblity` hook that tells whether we want the things in the table focusable at all times or not

Listening to `a11y` changes in `TableWrapper` is another reason to put the navigation listener in that file as well, and having seperate listeners for selections and sorting, placed on cell level. I'll look into that refactoring...

### DIFF
--- a/src/table/cells/handle-key-press.js
+++ b/src/table/cells/handle-key-press.js
@@ -73,6 +73,7 @@ const handleKeyPress = (evt, rootElement, cellCoord, setFocusedCell, cell, selSt
     case 'Enter': {
       preventDefaultBehavior(evt);
       selState.api.confirm();
+      // TODO: make sure focus is kept, or set on nebula cell
       break;
     }
     // Esc: Cancels selections if any, otherwise blur
@@ -80,14 +81,20 @@ const handleKeyPress = (evt, rootElement, cellCoord, setFocusedCell, cell, selSt
       preventDefaultBehavior(evt);
       if (selState?.rows.length) {
         selState.api.cancel();
+        // TODO: make sure focus is kept
       } else {
         removeCurrentFocus(evt);
+        // dropFocus()
       }
       break;
     }
-    // Tab: blur cell
+    // Tab: blur cell, or if shift, focus nebula cell
     case 'Tab': {
+      // dropFocus()
       removeCurrentFocus(evt);
+      if (evt.shiftKey) {
+        // go to nebula... might not be done here
+      }
       break;
     }
     default:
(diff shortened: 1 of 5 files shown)


src/table/cells/handle-key-press.js
```diff
@@ -102,12 +98,28 @@ export const bodyHandleKeyPress = (evt, rootElement, rowIndex, colIndex, cell, s
...
       break;
     }
-    // Esc: Cancels selections.
+    // Esc: Cancels selections if any, otherwise blur
     case 'Escape': {
       preventDefaultBehavior(evt);
-      selState.api.cancel();
+      if (selState?.rows.length) {
+        selState.api.cancel();
+        // TODO: make sure focus is kept
```
haxxmaxx, line 109: We shouldn't have to do anything more here really. AFAIK if you have proper keys sets for all sibling elements, react should preserve the state of the element, including focus, when updating. I'll look further into this in a different PR

src/table/cells/handle-key-press.js
```diff
@@ -54,42 +49,43 @@ export const preventDefaultBehavior = (evt) => {
...
-export const moveFocus = (evt, rootElement, rowIndex, colIndex, selState) => {
+export const moveFocus = (evt, rootElement, cellCoord, setFocusedCell, selState) => {
   preventDefaultBehavior(evt);
   removeCurrentFocus(evt);
   const rowAndColumnCount = getRowAndColumnCount(rootElement);
-  const { nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumnCount, rowIndex, colIndex, selState);
-  moveToNextFocus(rowAndColumnCount.rowElements, nextRow, nextCol);
+  const nextCellCoord = arrowKeysNavigation(evt, rowAndColumnCount, cellCoord, selState);
+  focusCell(rowAndColumnCount.rowElements, nextCellCoord);
+  setFocusedCell(nextCellCoord);
```
LiKang6688, line 58: Could we make a choice to call them cell, like` focusedCell,` `setFocusedCell`, `nextCell`,  
or cellCoord, like `focusedCellCoord`, `setFocusedCellCoord`, `nextCellCoord` 
to keep those consistent? 
Personally, I prefer cell.

haxxmaxx: Good point. I agree, since it is not the actual cell, but only the coordinates

PR #140, opened 2021-03-24
chore: update build

Caele:
I got a question on QlikBranch on this example from the dev portal:
```
// render a mekko chart

// index.html
<!doctype html>
<html lang="en">
<meta charset="utf-8">
<title>Mekko chart</title>
<script src="https://unpkg.com/enigma.js@2.7.0/enigma.min.js"></script>
<script src="https://unpkg.com/@nebula.js/stardust@0.6.0"></script>
<script src="https://unpkg.com/@nebula.js/sn-mekko-chart"></script>

...
</html>

// index.js

(async () => {
  // open app

  // configure nucleus
  const nuked = window.stardust.embed(app, {
    context: { theme: 'light' },
    types: [{
      name: 'mekko',
      load: () => Promise.resolve(window['sn-mekko-chart']),
    }],
  });

  // create a session object
  nuked.render({
    element: document.querySelector('.object'),
    type: 'mekko',
    fields: ['Region', 'Product Group Desc', '=Sum([Sales Quantity]*[Sales Price])'],
  });
})();
```

The window['sn-mekko-chart'] looked odd, but when I tested it worked fine for mekko, but not for all other charts. 
It boils down to (I think) the target format ('umd') and the package name. For the table I had for some reason set it to "A Table", which means I couold find it through `window['A Table'])`. Previously I've worked a long time to make that work without success, turns out it was simple.

I also updated some more things, like added the banner.


rollup.config.js
```diff
@@ -20,14 +21,27 @@ function checkInternals() {
...
+
+const auth = typeof author === 'object' ? `${author.name} <${author.email}>` : author || '';
+const moduleName = name.split('/').reverse()[0];
+
+const banner = `/*
+* ${name} v${version}
+* Copyright (c) ${new Date().getFullYear()} ${auth}
+* Released under the ${license} license.
+*/
+`;
```
haxxmaxx, line 35: What does this do? Is it added to the DOM? Is it visible?

Caele: nah, it just sits at the top of the dist file

haxxmaxx: ah ok

haxxmaxx (approved):

PR #145, opened 2021-03-25
fix: update styling when theme changes

### LINKED ISSUE #139: Theme colors are not atuomatically updated in the new table
Theme colors are not automatically updated in the new table. A refresh is needed for the colors to be applied or if you do changes to the properties after the theme change then the theme colors will also be applied.

haxxmaxx:
fixes #139


LiKang6688 (approved):
looks good to me

PR #147, opened 2021-03-25
remove layout as a dependency to avoid extra render

LiKang6688:


src/table/cells/handle-key-press.js
```diff
@@ -115,7 +114,12 @@ export const bodyHandleKeyPress = (evt, rootElement, rowIndex, colIndex, cell, s
     // Enter: Confirms selections.
     case 'Enter': {
       preventDefaultBehavior(evt);
-      selState.api.confirm();
+      selState.api.confirm().then(() => {
+        setTimeout(() => {
+          const { rowElements } = getRowAndColumnCount(rootElement);
+          moveToNextFocus(rowElements, 1, 0);
+        }, [400]);
+      });
```
haxxmaxx, line 122: Unfortunately, we can't rely on this. If the engine takes longer than 400ms than we lose focus anyway. Maybe we can have some sort of state that tells the chart to set focus when it rerenders, e.g. `shouldRefocus`. But then setting that fo false again might trigger a new reload and then it looses focus. Another alternative is that nebula handles that. So whenever the selections are confirmed, focus is set to the entire chart

haxxmaxx:
Just tried it out. This fixes the performance issue with sorting, and selections also look faster :) Just make sure to test everything is updated as it should, so we miss an update when it actually should update

src/table/cells/handle-key-press.js
LiKang6688: True. The engine! 
It goes like that the table receives an 'Entre' key event and calls the API. The engine confirms the selections. Then the nebula calls the component() again so the table rerenders and gets the latest table data and etc.
I don't know if we are able to let the table set focus after the table finishes rerendering.
Can we know the rerender of the table is caused by confirming selections?

haxxmaxx: If you set some state on the TableWrapper only for when confirming with enter, then it checks that state on rerender and sets focus and tabindex

LiKang6688: that state will be checked within react's lifestyle which is before the nebula calls the component() to run reactROM.render() and the table rerenders. So it is not working as we want.

haxxmaxx: hmm true. What about adding a state on the index component, then passing that into the table? If a set function is passed as well, then it should be possible to set it on confirm, and it should be true when render runs after the data is updated. It would kind of be a placeholder until the useAccessibility hook is done

LiKang6688: tried but when setState is called, it will cause the index component to run immediately.

LiKang6688: I think for 1.0, it is what we could do. I will look at the nebula for phrase 2.

LiKang6688:
> Just tried it out. This fixes the performance issue with sorting, and selections also look faster :) Just make sure to test everything is updated as it should, so we miss don't an update when it actually should update

Maybe it could be merged and tested regularly.

src/table/cells/handle-key-press.js
haxxmaxx: I'd say it's better to skip it then. Better to have it lose focus consistently than it keeping focus inconsistently

haxxmaxx:
Does the memo thing solve so doesn't cancel lose focus?

> > Just tried it out. This fixes the performance issue with sorting, and selections also look faster :) Just make sure to test everything is updated as it should, so we miss don't an update when it actually should update
> 
> Maybe it could be merged and tested regularly.

Well, the approach we try to have to manual testing is basically this: The developer is in the end responsible for quality. We should aim for minimizing relying on manual testers. So when adding a new feature or fixing a bug, make sure you spend some time on testing it yourself first. When it comes to feature work, a manual tester almost always have a look at it. They usually come up with cases the developer missed.

So in this case. Just have a quick run-through the functionality we have, so that e.g. changing a property or something still updates properly

haxxmaxx (approved):

LiKang6688:
> Does the memo thing solve so doesn't cancel lose focus?
> 
> > > Just tried it out. This fixes the performance issue with sorting, and selections also look faster :) Just make sure to test everything is updated as it should, so we miss don't an update when it actually should update
> > 
> > 
> > Maybe it could be merged and tested regularly.
> 
> Well, the approach we try to have to manual testing is basically this: The developer is in the end responsible for quality. We should aim for minimizing relying on manual testers. So when adding a new feature or fixing a bug, make sure you spend some time on testing it yourself first. When it comes to feature work, a manual tester almost always have a look at it. They usually come up with cases the developer missed.
> 
> So in this case. Just have a quick run-through the functionality we have, so that e.g. changing a property or something still updates properly

Thanks. Very good to know and I am in agreement with you.

It is removing layout in [constraints, layout] as a dependency to avoid extra render that cancels lose focus.

React.memo(TableWrapper) does not help that. It only helps with the performance. 
Yeah, I am thinking I should not include it in this PR.

PR #151, opened 2021-03-29
fix: the values for a measure are empty when added

LiKang6688:
Screenshot 2021-03-29 at 15 59 43


src/table/TableBodyWrapper.jsx
```diff
@@ -39,7 +39,7 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
     const selectionsEnabled = !!selectionsAPI && !constraints.active;
     selDispatch({ type: 'set-enabled', payload: { isEnabled: selectionsEnabled } });
     setColumnRenderers(getColumnRenderers(selectionsEnabled));
-  }, [constraints]);
+  }, [constraints, columns.length]);
```
LiKang6688, line 42: invoking setColumnRenderers when measure or dimension is updated, which means when the length of columns is changed, columnRenderers should be updated

haxxmaxx (approved):
Good find, I guess this is where we listened to layout before

PR #153, opened 2021-03-29
fix: Can't do the sorting if one measure is ordered before at least one dimension

LiKang6688:


src/table/sorting.js
```diff
@@ -1,11 +1,17 @@
 export default function sortingFactory(model) {
-  return (layout, isDim, sortIdx) => {
+  return (layout, isDim, label) => {
+    const { qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
+    const index = isDim
+      ? qDimensionInfo.findIndex((dimensionInfo) => dimensionInfo.qFallbackTitle === label)
+      : qMeasureInfo.findIndex((measureInfo) => measureInfo.qFallbackTitle === label);
+    const idx = isDim ? index : index + qDimensionInfo.length;
```
haxxmaxx, line 7: This is exactly what qColumnOrder can be used for. `qColumnOrder[sortIdx]` should give you the same thing.

LiKang6688: qColumnOrder is the order of columns
Screenshot 2021-03-30 at 09 31 22

idx is the order of sorting
Screenshot 2021-03-30 at 09 32 01

LiKang6688: but when the order of Columns is not set, like the case in nebula, qColumnOrder is []. qColumnOrder[sortIdx] will be undefined.

haxxmaxx: ah true, but then `getColumnOrder` returns a proper array

haxxmaxx: return that with `manageData` then send it into `sortingFactory`. Also I'd run `sortingFactory` inside the render `useEffect`.

LiKang6688: Right. Good idea. I Will work on it.

PR #154, opened 2021-03-29
disable sorting in edit mode

LiKang6688:


haxxmaxx:
Good find. What about sorting with the keyboard?

LiKang6688:
> Good find. What about sorting with the keyboard?

Thanks. Missing that. Fixed

haxxmaxx (approved):

PR #155, opened 2021-03-30
fix: selections when one measure before one dimension

LiKang6688:


src/table/TableBodyWrapper.jsx
```diff
@@ -57,6 +57,9 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
         >
           {columns.map((column, columnIndex) => {
             const cell = row[column.id];
+            cell.colIdx = layout.qHyperCube.qDimensionInfo.findIndex(
+              (dimensionInfo) => dimensionInfo.qFallbackTitle === column.label
+            );
```
haxxmaxx, line 62: Can you do this in `handle-data` instead?

LiKang6688: good ieda, yes

haxxmaxx (approved):

PR #164, opened 2021-04-07
set the tabindex of the previous focused cell (if there was one) to -1

LiKang6688:


src/table/selections-utils.js
```diff
@@ -55,7 +58,18 @@ export const getSelectedRows = (selectedRows, qElemNumber, rowIdx, evt) => {
   return selectedRows;
 };
 
-export function selectCell(cell, selState, selDispatch, evt) {
+const removeTabSequence = (rowElements, nextCellCoord) => {
+  const cell = rowElements[nextCellCoord[0]]?.getElementsByClassName('sn-table-cell')[nextCellCoord[1]];
+  cell.setAttribute('tabIndex', '-1');
+};
+
+export function selectCell(cell, selState, selDispatch, focusedCell, setFocusedCell, rootElement, pageInfo, page, evt) {
```
haxxmaxx, line 66: I think codeclimate has an valid point here. 9 arguments for a function usually indicates that it is doing too much. We want to keep the focus handling separate from the selection handling

LiKang6688: Right. Agree. Gonna fix.

src/table/selections-utils.js
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line import/no-cycle
```
haxxmaxx, line 1: this is not a good idea.

haxxmaxx:
Having trouble understanding the flow in general. You are keeping track of the last focused cell, and then removing that tabindex when a selection is made, based on that a selection is made when clicking a cell?

LiKang6688:
> Having trouble understanding the flow in general. You are keeping track of the last focused cell, and then removing that tabindex when a selection is made, based on that a selection is made when clicking a cell?
```
if (focusedCell.length === 0) {
    removeTabSequence(rootElement.getElementsByClassName('sn-table-row'), [0, 0]);
  } else {
    focusedCell[2] === page && removeTabSequence(rootElement.getElementsByClassName('sn-table-row'), focusedCell);
  }
setFocusedCell([rowIdx + 1 - pageInfo.height * page, colIdx, page]);
focusCell(rootElement.getElementsByClassName('sn-table-row'), [rowIdx + 1 - pageInfo.height * page, colIdx]);
```
when a cell is selected, its tabindex is set to 0. the tabindex of the previous cell is set to -1.

haxxmaxx:
Tried it out a bit and I can still get two cells with tabindex 0. The easiest way to do this is simply to reset the tabindex for all cells in the table. I tried that briefly. It works, but don't know how it affects performance/reloads. If there is no difference, I think we should do that. No need to keep track of what cell had focus last.

LiKang6688:
> Tried it out a bit and I can still get two cells with tabindex 0. The easiest way to do this is simply to reset the tabindex for all cells in the table. I tried that briefly. It works, but don't know how it affects performance/reloads. If there is no difference, I think we should do that. No need to keep track of what cell had focus last.

The new way sounds good. Could you push that in a new PR? I would like to have a look.

haxxmaxx:
> > Tried it out a bit and I can still get two cells with tabindex 0. The easiest way to do this is simply to reset the tabindex for all cells in the table. I tried that briefly. It works, but don't know how it affects performance/reloads. If there is no difference, I think we should do that. No need to keep track of what cell had focus last.
> 
> The new way sounds good. Could you push that in a new PR? I would like to have a look.

Haven't really cleaned it up. I basically do
```
  const cells = rootElement.getElementsByClassName('sn-table-cell')

  for (let i = 0; i < cells.length; i++) {
    cells[i].setAttribute('tabIndex', '-1');
  }
  ```
 inside handle cell focus.

**Comment by LiKang6688** [comment]
> > > Tried it out a bit and I can still get two cells with tabindex 0. The easiest way to do this is simply to reset the tabindex for all cells in the table. I tried that briefly. It works, but don't know how it affects performance/reloads. If there is no difference, I think we should do that. No need to keep track of what cell had focus last.
> > 
> > 
> > The new way sounds good. Could you push that in a new PR? I would like to have a look.
> 
> Haven't really cleaned it up. I basically do
> 
> ```
>   const cells = rootElement.getElementsByClassName('sn-table-cell')
> 
>   for (let i = 0; i < cells.length; i++) {
>     cells[i].setAttribute('tabIndex', '-1');
>   }
> ```
> 
> inside handle cell focus.

When the Rows per page are 100, I do not think the performance is good enough.  I can significantly feel that it is laggy.

**Comment by LiKang6688** [comment]
> > > > Tried it out a bit and I can still get two cells with tabindex 0. The easiest way to do this is simply to reset the tabindex for all cells in the table. I tried that briefly. It works, but don't know how it affects performance/reloads. If there is no difference, I think we should do that. No need to keep track of what cell had focus last.
> > > 
> > > 
> > > The new way sounds good. Could you push that in a new PR? I would like to have a look.
> > 
> > 
> > Haven't really cleaned it up. I basically do
> > ```
> >   const cells = rootElement.getElementsByClassName('sn-table-cell')
> > 
> >   for (let i = 0; i < cells.length; i++) {
> >     cells[i].setAttribute('tabIndex', '-1');
> >   }
> > ```
> > 
> > 
> > inside handle cell focus.
> 
> When the Rows per page are 100 and the size of columns is 8, the number of cells is more than 800. I can really feel laggy in sense-client.

File: src/table/cells/handle-cell-focus.js [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+};
+
+const handleCellFocus = (columnIndex, cell, focusedCell, setFocusedCell, rootElement, pageInfo, page) => {
+  const { rowIdx } = cell;
+
+  focusedCell.length === 0
+    ? removeTabSequence(rootElement.getElementsByClassName('sn-table-row'), [0, 0])
+    : focusedCell[2] === page && removeTabSequence(rootElement.getElementsByClassName('sn-table-row'), focusedCell);
+
+  setFocusedCell([rowIdx + 1 - pageInfo.height * page, columnIndex, page]);
```
LiKang6688, line 15: did a fix here. should be columnIndex instead of colIdx from cell

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,27 @@
...
+
+  // const cells = rootElement.getElementsByClassName('sn-table-cell');
+  // for (let i = 0; i < cells.length; i++) {
+  //   cells[i].setAttribute('tabIndex', '-1');
+  // }
+
+  focusCell(rootElement.getElementsByClassName('sn-table-row'), [rowIdx + 1 - pageInfo.height * page, columnIndex]);
```
LiKang6688, line 22: did a fix here. should be columnIndex instead of colIdx from cell

src/table/cells/handle-key-press.js
```diff
@@ -67,48 +57,71 @@ export const preventDefaultBehavior = (evt) => {
...
-      !isCtrlShift(evt) && moveFocus(evt, rootElement, rowIndex, colIndex, selState);
+      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, setFocusedCell, selState, page);
       break;
     }
     // Space bar: Selects value.
     case ' ': {
       preventDefaultBehavior(evt);
-      cell?.isDim && selectCell(cell, selState, selDispatch, evt);
+      cell?.isDim &&
+        handleCellFocus(cell, focusedCell, setFocusedCell, rootElement, page) &&
```
LiKang6688, line 123: fixed the case before clicking,  first tabing into the chart and move a couple of cells

haxxmaxx: why do you need `page`?

LiKang6688: `focusedCell[2] === page && removeTabSequence(rootElement.getElementsByClassName('sn-table-row'), focusedCell)`

to check the page that a user is selecting by clicking or space key is the same as the page of the focused cell or not.
If it is the same, set tabIndex to -1. Otherwise, no need to do that.

haxxmaxx: But it is still the same cell, on main at least. If I navigate to a cell with the keyboard, then go to the next page, that last focus cell still has tabindex 0. In any case, passing around `page` this way is not worth it. Setting tabindex -1 one to many times shouldn't be that performance heavy.

LiKang6688: You are right. And I came up with a way to address the issue when removing `page` by `cell && cell.setAttribute('tabIndex', '-1');`

LiKang6688: When we change the` rows per page` and sometimes `page`, we will lose focus. it should be created in a new issue.

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,25 @@
+export const focusCell = (rowElements, nextCellCoord) => {
+  const nextCell = rowElements[nextCellCoord[0]].getElementsByClassName('sn-table-cell')[nextCellCoord[1]];
+  nextCell.focus();
+  nextCell.setAttribute('tabIndex', '0');
+};
+
+const removeTabSequence = (rowElements, nextCellCoord) => {
```
haxxmaxx, line 7: This name is a bit misleading. Use `removeTabStop` or `resetTabIndex` or something

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,25 @@
...
+  const cell = rowElements[nextCellCoord[0]]?.getElementsByClassName('sn-table-cell')[nextCellCoord[1]];
+  cell && cell.setAttribute('tabIndex', '-1');
+};
+
+export const handleCellFocus = (cell, focusedCell, setFocusedCell, rootElement) => {
+  const { rawRowIdx, rawColIdx } = cell;
+
+  removeTabSequence(
+    rootElement.getElementsByClassName('sn-table-row'),
```
haxxmaxx, line 16: This is fetched twice, better to do it once and save it as a `const`

src/table/cells/handle-cell-focus.js
LiKang6688: sure. I will go for resetTabIndex

src/table/cells/handle-cell-focus.js
LiKang6688: good point. gonna fix.

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,25 @@
...
+    focusedCell.length === 0 ? [0, 0] : focusedCell
+  );
+
+  setFocusedCell([rawRowIdx + 1, rawColIdx]);
+
+  focusCell(rootElement.getElementsByClassName('sn-table-row'), [rawRowIdx + 1, rawColIdx]);
+
+  return true;
```
haxxmaxx, line 24: I think it's better if there is no return from this function.

src/table/cells/handle-key-press.js
```diff
@@ -67,48 +57,60 @@ export const preventDefaultBehavior = (evt) => {
...
-      !isCtrlShift(evt) && moveFocus(evt, rootElement, rowIndex, colIndex, selState);
+      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, setFocusedCell, selState);
       break;
     }
     // Space bar: Selects value.
     case ' ': {
       preventDefaultBehavior(evt);
-      cell?.isDim && selectCell(cell, selState, selDispatch, evt);
+      cell?.isDim &&
+        handleCellFocus(cell, focusedCell, setFocusedCell, rootElement) &&
```
haxxmaxx, line 112: Does it need to be called here?

src/table/cells/withSelections.jsx
```diff
@@ -1,12 +1,19 @@
...
 
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
-    const { cell, selState, selDispatch, styling } = props;
-    const handleMouseUp = (evt) => cell.isDim && evt.button === 0 && selectCell(cell, selState, selDispatch, evt);
+    const { cell, selState, selDispatch, styling, focusedCell, setFocusedCell, rootElement } = props;
+    const handleMouseUp = (evt) => {
+      cell.isDim &&
+        evt.button === 0 &&
+        handleCellFocus(cell, focusedCell, setFocusedCell, rootElement) &&
```
haxxmaxx, line 13: I might have mentioned this already, but we might need to move this outside selections later. So you can click to focus but not make selections. This is fine for now though :)

src/table/cells/handle-key-press.js
LiKang6688: clicking is the same as the space key, right?

src/table/cells/withSelections.jsx
LiKang6688: click to focus but not make selections, in the real scenario, does such a case exist?

haxxmaxx: It does, try the old table in edit mode and click a cell

src/table/cells/handle-key-press.js
haxxmaxx: well, almost. Clicking can both move focus and make selection. Pressing space should only select, since you already have navigated to that cell. The focus state should be correct before pressing space

LiKang6688: The focus state is not correct before pressing space. Pressing space should also resetTabIndex.

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,20 @@
...
+const resetTabIndex = (rowElements, nextCellCoord) => {
+  const cell = rowElements[nextCellCoord[0]]?.getElementsByClassName('sn-table-cell')[nextCellCoord[1]];
+  cell && cell.setAttribute('tabIndex', '-1');
+};
+
+export const handleCellFocus = (cell, focusedCell, setFocusedCell, rootElement) => {
+  const { rawRowIdx, rawColIdx } = cell;
+  const tableRow = rootElement.getElementsByClassName('sn-table-row');
+  resetTabIndex(tableRow, focusedCell.length === 0 ? [0, 0] : focusedCell);
+  setFocusedCell([rawRowIdx + 1, rawColIdx]);
```
LiKang6688, line 16: it is not fast enough to set array or object

src/table/cells/handle-key-press.js
```diff
@@ -67,60 +57,79 @@ export const preventDefaultBehavior = (evt) => {
...
-export const moveFocus = (evt, rootElement, rowIndex, colIndex, selState) => {
+export const moveFocus = (evt, rootElement, cellCoord, setFocusedCell, selState) => {
   preventDefaultBehavior(evt);
   removeCurrentFocus(evt);
   const rowAndColumnCount = getRowAndColumnCount(rootElement);
-  const { nextRow, nextCol } = arrowKeysNavigation(evt, rowAndColumnCount, rowIndex, colIndex, selState);
-  moveToNextFocus(rowAndColumnCount.rowElements, nextRow, nextCol);
+  const nextCellCoord = arrowKeysNavigation(evt, rowAndColumnCount, cellCoord, selState);
+  focusCell(rowAndColumnCount.rowElements, nextCellCoord);
+  setFocusedCell(nextCellCoord[1]);
```
LiKang6688, line 66: to set number instead of an array to fix performance issue

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,20 @@
...
+  const nextCell = rowElements[nextCellCoord[0]].getElementsByClassName('sn-table-cell')[nextCellCoord[1]];
+  nextCell.focus();
+  nextCell.setAttribute('tabIndex', '0');
+};
+
+const resetTabIndex = (tableRow, focusedCell) => {
+  for (let i = 0; i < tableRow.length; i++) {
+    const cell = tableRow[i].children[focusedCell];
+    cell && cell.setAttribute('tabIndex', '-1');
+  }
```
LiKang6688, line 11: reset tab index for all cells in specific column

src/table/cells/handle-key-press.js
haxxmaxx: but this is only the row number, how would that work?

LiKang6688: it is the nextCol, the column number.

haxxmaxx: yes, 0 is the first... of course hehe. But it still missing the row index in that case

LiKang6688: yes, to fix the performance issue we talked about yesterday, here only column number is set, then rest the tab index of all cells in this column no matter which row it is.

haxxmaxx: What happens if you have a lot of columns and move fast to the right?

LiKang6688: good point. it is a bit laggy. but compared to move up and down, it is less common to happen. Currently, I do not come up with a way of not using setFocusedCell.

src/table/TableWrapper.jsx
```diff
@@ -30,6 +30,7 @@ export default function TableWrapper(props) {
   const [tableWidth, setTableWidth] = useState();
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(100);
+  const focusedCellCoord = useRef([0, 0]);
```
LiKang6688, line 33: use useRef to set mutable data that is tied to the component instance, but doesn't trigger any renders on change

haxxmaxx: NIce, that is what we want.

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,18 @@
...
+const resetTabIndex = (rowElements, focusedCellCoord) => {
+  const cell = rowElements[focusedCellCoord[0]]?.getElementsByClassName('sn-table-cell')[focusedCellCoord[1]];
+  cell && cell.setAttribute('tabIndex', '-1');
+};
+
+export const handleCellFocus = (cell, focusedCellCoord, rootElement) => {
+  const { rawRowIdx, rawColIdx } = cell;
+  const rowElements = rootElement.getElementsByClassName('sn-table-row');
+  resetTabIndex(rowElements, focusedCellCoord.current);
+  focusedCellCoord.current = [rawRowIdx + 1, rawColIdx];
```
haxxmaxx, line 16: Why are you sett ing `.current`?

LiKang6688: that is decided by react https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs https://reactjs.org/docs/hooks-reference.html#useref

haxxmaxx: Ahh smart

haxxmaxx:
Looks good, I think you should add it for the header as well. Then some tests and you should be done :)

LiKang6688:
> Looks good, I think you should add it for the header as well. Then some tests and you should be done :)

right, I just realized that handleMouseDown in the header is also needed. Thanks.

src/table/cells/handle-cell-focus.js
```diff
@@ -0,0 +1,27 @@
...
+const resetTabIndex = (rowElements, focusedCellCoord) => {
+  const cell = rowElements[focusedCellCoord[0]]?.getElementsByClassName('sn-table-cell')[focusedCellCoord[1]];
+  cell && cell.setAttribute('tabIndex', '-1');
+};
+
+export const handleBodyCellFocus = (cell, focusedCellCoord, rootElement) => {
+  const { rawRowIdx, rawColIdx } = cell;
+  const rowElements = rootElement.getElementsByClassName('sn-table-row');
+  resetTabIndex(rowElements, focusedCellCoord.current);
+  const focusedCellCoordCopy = focusedCellCoord;
```
haxxmaxx, line 16: Why is this needed?

LiKang6688: the lint in circleCI complains no-param-reassign. So did a fix here.

haxxmaxx: Yeah we do that all over the place, so you can actually add an ignore comment here. Still keeping it as a rule since it's good to keep track of.

haxxmaxx: would be good with some simple tests here, on for `handleBodyCellFocus` and one for `handleHeadCellFocus` should be enough

LiKang6688: > Yeah we do that all over the place, so you can actually add an ignore comment here. Still keeping it as a rule since it's good to keep track of.

I think it is better to do a fix like that instead of adding an ignore comment.

LiKang6688: > would be good with some simple tests here, on for `handleBodyCellFocus` and one for `handleHeadCellFocus` should be enough

Yes, I have been writing the unit testing, which will be committed later today.

haxxmaxx: > > Yeah we do that all over the place, so you can actually add an ignore comment here. Still keeping it as a rule since it's good to keep track of.
> 
> I think it is better to do a fix like that instead of adding an ignore comment.

I don't. Pretty sure you are actually just creating a reference to the real `focusedCellCoord` anyway, otherwise `focusedCellCoord` wouldn't be updated where it is stored. So you still have the risk of unexpected behaviour when reassigning arguments, but the linter doesn't catch it :/ Could maybe make it throw warnings instead of errors though

haxxmaxx: Actually, I think this is relevant. https://eslint.org/docs/rules/no-param-reassign#props So reassigning props is fine but not the entire argument. Feel free to set that up if you want :)

haxxmaxx: Actually, I added it 🙃  https://github.com/qlik-oss/sn-table/pull/173

LiKang6688: > Actually, I added it 🙃 #173

Good. Thanks.

src/table/__tests__/TableHeadWrapper.spec.jsx
```diff
@@ -100,4 +107,40 @@ describe('<TableHeadWrapper />', () => {
 
     expect(changeSortOrder).to.not.have.been.calledOnce;
   });
+
+  it('should call bodyHandleKeyPress when keyDown on a header cell', () => {
```
haxxmaxx, line 111: wrong description?

src/table/cells/__tests__/handle-cell-focus.spec.js
```diff
@@ -0,0 +1,67 @@
...
+    const cell = {
+      rawRowIdx: -1,
+      rawColIdx: 0,
+    };
+    const focusedCellCoord = {
+      current: [0, 0],
+    };
+    const rootElement = {
+      getElementsByClassName: () => [
+        { getElementsByClassName: () => [{ focus: sinon.spy(), setAttribute: sinon.spy() }] },
```
haxxmaxx, line 27:
```suggestion
    const cell = { focus: sinon.spy(), setAttribute: sinon.spy() };
    const rootElement = {
      getElementsByClassName: () => [
        { getElementsByClassName: () => [cell] },
```
If you do something like this I don't think you have to return the cell. Shouldn't have returns unless it is really needed

src/table/__tests__/TableHeadWrapper.spec.jsx
LiKang6688: yep, thanks

src/table/cells/__tests__/handle-cell-focus.spec.js
LiKang6688: Perfect. Thanks!

haxxmaxx (approved):

PR #168, opened 2021-04-08
feat: inline styling to minimize style tags

### LINKED ISSUE #160: VF: a lot of <style> tags are added to the header, could possibly be avoided
Seems like it is one per cell. The weird thing is that all the tags look empty
Screenshot 2021-04-07 at 15 59 05

haxxmaxx:
fixes #160


src/table/TableBodyWrapper.jsx
```diff
@@ -5,16 +5,28 @@ import TableBody from '@material-ui/core/TableBody';
...
+    },
+  },
   hoverTableRow: {
     '&&:hover': {
       backgroundColor: 'rgba(0, 0, 0, 0)',
       '& td:not(.selected)': {
-        backgroundColor: ({ hoverBackgroundColor }) => hoverBackgroundColor,
-        color: ({ hoverFontColor }) => hoverFontColor,
+        backgroundColor: ({ hoverBackgroundColor }) => `${hoverBackgroundColor} !important`,
+        color: ({ hoverFontColor }) => `${hoverFontColor} !important`,
```
haxxmaxx, line 29: Since inline styling always has higher priority than classes and this is on row level, we have to use !important here.

src/table/TableBodyWrapper.jsx
```diff
@@ -5,16 +5,28 @@ import TableBody from '@material-ui/core/TableBody';
...
+  cellBase: {
+    '& td': {
+      color: ({ color }) => color,
+      fontSize: ({ fontSize }) => fontSize,
+      padding: ({ padding }) => padding,
+      height: STYLING_DEFAULTS.HEIGHT,
+      lineHeight: STYLING_DEFAULTS.HEAD_LINE_HEIGHT,
+      '&&:focus': {
+        boxShadow: STYLING_DEFAULTS.FOCUS_OUTLINE,
+      },
```
haxxmaxx, line 21: These could be moved to the theme, but that is a different PR.

LiKang6688:
I thought you were fixing the style tags and I still found there are many style tags in this branch. Am I wrong here?
Screenshot 2021-04-14 at 17 01 20

By the way, I found that in those style tags, it writes 'Copyright (c) Microsoft Corporation' which is wired for me.
Screenshot 2021-04-14 at 17 01 43

haxxmaxx:
> I thought you were fixing the style tags and I still found there are many style tags in nebula dev in this branch. Am I wrong here? Should I have a look in the mashup instead?

Yeah I think those are related to the nebula editor, the property panel, current selections etc. If you look at the `main` branch you find a bunch of these: `<style data-jss="" data-meta="makeStyles"></style>`, one for every cell and they all appear to be empty.

LiKang6688:
> > I thought you were fixing the style tags and I still found there are many style tags in nebula dev in this branch. Am I wrong here? Should I have a look in the mashup instead?
> 
> Yeah I think those are related to the nebula editor, the property panel, current selections etc. If you look at the `main` branch you find a bunch of these: `<style data-jss="" data-meta="makeStyles"></style>`, one for every cell and they all appear to be empty.
> 
> If you want to you can have a look in the mashup to see if it is less. It was all the empty ones we wanted to get rid off.

Yes, I got it. Thanks.

src/table/TableBodyWrapper.jsx
```diff
@@ -66,7 +78,7 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
                   value={value}
                   key={column.id}
                   align={column.align}
-                  styling={styling}
+                  styling={{}}
```
LiKang6688, line 81: here the props - styling is assigned with {}. in withStyling.jsx, the styling gets CSS properties  but not in withSelections.jsx. why is it like that?

src/table/TableBodyWrapper.jsx
```diff
@@ -46,7 +58,7 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
   }, []);
 
   return (
-    <TableBody>
+    <TableBody class={classes.cellBase}>
```
LiKang6688, line 61: should be className

haxxmaxx: it sure should

src/table/TableBodyWrapper.jsx
haxxmaxx: So in the case of no column styling and no ongoing selections, there is not styling applied inline. When there is, that styling is added to `styling` in `withColumnStyling` and `withSelections`. Does that answer your question?

LiKang6688: yes. When there is, how is that styling added to `styling`?

haxxmaxx: Just like before, in `styling-utils`. We need an empty object to start with, so that property always exist on the component, and then you can use any combination of `withColumnStyling` and `withSelections`.

src/table/cells/withStyling.jsx
```diff
@@ -1,35 +1,10 @@
...
-    lineHeight: STYLING_DEFAULTS.BODY_LINE_HEIGHT,
-    '&&:focus': {
-      boxShadow: STYLING_DEFAULTS.FOCUS_OUTLINE,
-    },
-  }),
-});
 
 export default function withStyling(CellComponent) {
   const HOC = (props) => {
     const { styling } = props;
```
LiKang6688, line 6: Okay, now I see.  I thought `styling` here was from <TableBody><TableRow><CellRenderer styling={styling}></CellRenderer></TableRow></TableBody>. Now I realize that it can also from <CellComponent styling={selectionStyling} > and <CellComponent {...props} styling={columnStyling} />.

LiKang6688 (approved):
Looks good to me. Not sure I manually test all styling cases.

PR #172, opened 2021-04-14
fix: handle focus correctly for confirm, changing page, changing rpp

haxxmaxx:
Using Li's branch as base for now. The flow here requires some explanation.

- A useRef called `shouldRefocus` keeps track of whether the first cell should be focused after certain re-renders. There is also a setter callback called `setShouldRefocus` to reduce the amount of arguments needed. It is basically set to true if a cell in the table has focus when confirming or changing page with the keyboard shortcut. (could be extended to other cases as well)
- When the data is changed so that the previous focused cell might not exist anymore, we set `focusedCellCoord` to `[0, 0]`and the tabindex for that cell to 0. This is when the number of rows, columns or rows per page change. The exception is in selection mode, where you stay on the same column when changing page.
- If `shouldRefocus` is true, we also focus that cell on after re-render

The only thing that is not working is if you make selections, go the the next page and confirm, you lose focus. But the tabindex is still ok, so I think it is fine.


src/table/TableWrapper.jsx
```diff
@@ -31,10 +33,14 @@ export default function TableWrapper(props) {
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(100);
   const focusedCellCoord = useRef([0, 0]);
+  const shouldRefocus = useRef(false);
   const classes = useStyles();
   const containerMode = constraints.active ? 'containerOverflowHidden' : 'containerOverflowAuto';
   const paginationHidden = constraints.active && 'paginationHidden';
   const paginationFixedRpp = selectionsAPI.isModal() || tableWidth < 400;
+  const setShouldRefocus = () => {
+    shouldRefocus.current = rootElement.getElementsByTagName('table')[0].contains(document.activeElement);
```
LiKang6688, line 42: unit test does not cover here.

src/table/TableWrapper.jsx
```diff
@@ -59,15 +65,21 @@ export default function TableWrapper(props) {
     return () => window.removeEventListener('resize', updateSize);
   }, []);
 
+  // Except for first render, whenever the size of the data changes (number of rows per page, rows or columns),
+  // reset tabindex to first cell. If some cell had focus, focus the first cell as well.
+  useDidUpdateEffect(() => {
+    handleResetFocus(focusedCellCoord, rootElement, shouldRefocus, selectionsAPI.isModal());
+  }, [tableData.rows.length, tableData.size.qcy, tableData.size.qcx]);
+
```
LiKang6688, line 73: smart, nice

src/table/TableWrapper.jsx
haxxmaxx: Testing this function specifically is tricky. It's nice to have it here since all the needed variables are in this scope. I guess I can rewrite it so it runs the logic inside `handle-cell-focus` instead.

src/table/TableWrapper.jsx
```diff
@@ -31,10 +33,14 @@ export default function TableWrapper(props) {
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(100);
   const focusedCellCoord = useRef([0, 0]);
+  const shouldRefocus = useRef(false);
   const classes = useStyles();
   const containerMode = constraints.active ? 'containerOverflowHidden' : 'containerOverflowAuto';
   const paginationHidden = constraints.active && 'paginationHidden';
   const paginationFixedRpp = selectionsAPI.isModal() || tableWidth < 400;
+  const setShouldRefocus = () => {
```
LiKang6688, line 41: name it to setShouldRefocusCell?

LiKang6688 (approved):
Looks good to me.

rollup.config.js
```diff
@@ -41,7 +41,7 @@ export default {
     name: moduleName,
     format: 'umd',
     exports: 'default',
-    sourcemap: false,
+    sourcemap: true,
```
LiKang6688, line 44: why do we change sourcemap in this branch?

src/table/cells/handle-cell-focus.js
```diff
@@ -1,25 +1,43 @@
-export const focusCell = (rowElements, nextCellCoord) => {
-  const nextCell = rowElements[nextCellCoord[0]].getElementsByClassName('sn-table-cell')[nextCellCoord[1]];
-  nextCell.focus();
-  nextCell.setAttribute('tabIndex', '0');
+export const updateFocus = (rowElements, cellCoord, shouldBlur = false) => {
```
LiKang6688, line 1: the fucntion is updateFocus, for me,` shouldFocus = true `is easier to understand.

src/table/TableWrapper.jsx
```diff
@@ -59,15 +65,21 @@ export default function TableWrapper(props) {
     return () => window.removeEventListener('resize', updateSize);
   }, []);
 
+  // Except for first render, whenever the size of the data changes (number of rows per page, rows or columns),
+  // reset tabindex to first cell. If some cell had focus, focus the first cell as well.
+  useDidUpdateEffect(() => {
+    handleResetFocus(focusedCellCoord, rootElement, shouldRefocus, selectionsAPI.isModal());
+  }, [tableData.rows.length, tableData.size.qcy, tableData.size.qcx]);
```
LiKang6688, line 72:
```suggestion
  }, [rows.length, size.qcy, size.qcx]);
```

rollup.config.js
haxxmaxx: No I just have that when developing, will change it

src/table/cells/handle-cell-focus.js
haxxmaxx: Hmm yeah I think I agree, I just wanted to avoid using the same name as `shouldRefocus`, but that is not the same as `shouldFocus`. So I'll change it.

src/table/TableWrapper.jsx
haxxmaxx: good find

PR #174, opened 2021-04-15
optimize table body

LiKang6688:


src/table/TableBodyWrapper.jsx
```diff
@@ -26,21 +26,20 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
   const styling = useMemo(() => getBodyStyle(layout, theme), [layout, theme.name()]);
   const classes = useStyles(styling);
-  const getColumnRenderers = (selectionsEnabled) =>
-    tableData.columns.map((c) => getCellRenderer(!!c.stylingInfo.length, selectionsEnabled));
-  const [columnRenderers, setColumnRenderers] = useState(getColumnRenderers(false));
+  const selectionsEnabled = !!selectionsAPI && !constraints.active;
+  const getColumnRenderers = tableData.columns.map((c) => getCellRenderer(!!c.stylingInfo.length, selectionsEnabled));
+  const [columnRenderers, setColumnRenderers] = useState(() => getColumnRenderers);
```
LiKang6688, line 31: do a Lazy initial state here https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

haxxmaxx:
What are all these commits? I'd suggest creating a new branch since the changes doesn't really have anything with all these commits to do

src/table/TableBodyWrapper.jsx
```diff
@@ -26,21 +26,20 @@ const TableBodyWrapper = ({ rootElement, tableData, constraints, selectionsAPI,
...
+    isEnabled: selectionsEnabled,
   });
 
   useEffect(() => {
-    const selectionsEnabled = !!selectionsAPI && !constraints.active;
     selDispatch({ type: 'set-enabled', payload: { isEnabled: selectionsEnabled } });
-    setColumnRenderers(getColumnRenderers(selectionsEnabled));
-  }, [constraints, columns.length]);
+    setColumnRenderers(getColumnRenderers);
+  }, [selectionsEnabled, columns.length]);
```
haxxmaxx, line 42: Hmm, is selectionsEnabled kept between renders? Otherwise it will change every render and this hook runs every time?

LiKang6688:
> What are all these commits? I'd suggest creating a new branch since the changes doesn't really have anything with all these commits to do

right, fixed

src/table/TableBodyWrapper.jsx
LiKang6688: the selectionsEnabled is same between renders. I think that selectionsEnabled was changed between analysis mode and edit mode.

src/table/TableHeadWrapper.jsx
```diff
@@ -32,7 +32,10 @@ export default function TableHeadWrapper({
   selectionsAPI,
   focusedCellCoord,
 }) {
-  const classes = useStyles(getHeadStyle(layout, theme));
+  const classes = useStyles(
+    useMemo(() => getHeadStyle(layout, theme)),
+    [layout, theme.name()]
+  );
```
haxxmaxx, line 38: I think you should save the head style outside of `useStyles`, the same way we do for body

haxxmaxx (approved):

src/table/TableHeadWrapper.jsx
LiKang6688: okay, look into it

LiKang6688: do you mean 
`const styling = useMemo(() => getBodyStyle(layout, theme), [layout, theme.name()]);
  const classes = useStyles(styling);`
in body?

haxxmaxx: Yes, just to avoid any weird stuff when nesting hooks like that

PR #176, opened 2021-04-16
feat: adding properties to pp and documentation to object-properties

### LINKED ISSUE #175: Add used properties to object-properties and ext

haxxmaxx:
fixes #175 

All the new components in `ext.js` are from the straight table, so all the translation strings are correct etc

### DIFF
--- a/src/ext.js
+++ b/src/ext.js
@@ -200,7 +200,7 @@ const stylingSettings = {
               ],
             },
             hoverColor: {
-              show: (data) => data.content.hoverEffect,
+              show: (data) => !!data.content.hoverEffect,
               ref: 'content.hoverColor',
               translation: 'ThemeStyleEditor.style.hoverStyle',
               type: 'object',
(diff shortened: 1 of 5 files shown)


src/ext.js
```diff
@@ -1,3 +1,227 @@
...
+      items: {
+        chart: {
+          type: 'items',
+          items: {
+            headerFontSize: {
+              show: true,
+              ref: 'header.fontSize',
+              translation: 'ThemeStyleEditor.style.headerFontSize',
+              component: 'integer',
+              // placeholder: () => parseInt(styleService.getStyle('header', 'fontSize'), 10),
```
haxxmaxx, line 152: Commenting this for now, no need to recreate some `styleService` at this point.

src/ext.js
```diff
@@ -1,3 +1,227 @@
+const columnCommonHidden = {
```
haxxmaxx, line 1: A lot of complaining from code climate here. That the code look similar in a lot of places is natural, refactoring it would make it less readable IMO. The file is also really long, could be split up if we think it is a good idea.

LiKang6688:
sorry, I do not understand the added code. Any docs can you recommend me to read?

Caele (approved):

PR #180, opened 2021-04-23
fix cell focus style

LiKang6688:
a quick fix. will look into https://material-ui.com/styles/basics/#material-ui-styles and https://material-ui.com/customization/theming/ to use the theme. 

Not sure if rollup-plugin-postcss is in use.


haxxmaxx (approved):
awsm

PR #182, opened 2021-04-25
use a function to lazily initialize the variable

LiKang6688:


src/index.js
```diff
@@ -35,13 +30,10 @@ export default function supernova(env) {
       const constraints = useConstraints();
       const selectionsAPI = useSelections();
       const theme = useTheme();
-
       const [pageInfo, setPageInfo] = useState({ top: 0, height: 100 });
-      const [muiParameters] = useState(muiSetup(__OPIONAL_THEME_DEPS__));
-      const [tableData] = usePromise(() => manageData(model, layout, pageInfo), [layout, pageInfo]);
```
LiKang6688, line 41: in this case, at first, tableData is empty, then it got value, which results in that component() run twice.

src/index.js
```diff
@@ -35,13 +30,10 @@ export default function supernova(env) {
       const constraints = useConstraints();
       const selectionsAPI = useSelections();
       const theme = useTheme();
-
       const [pageInfo, setPageInfo] = useState({ top: 0, height: 100 });
-      const [muiParameters] = useState(muiSetup(__OPIONAL_THEME_DEPS__));
```
LiKang6688, line 40: removing useState for muiSetup to have a better performance.

src/index.js
```diff
@@ -35,13 +30,10 @@ export default function supernova(env) {
...
-      const [tableData] = usePromise(() => manageData(model, layout, pageInfo), [layout, pageInfo]);
 
       useEffect(() => {
-        if (layout && tableData) {
+        manageData(model, layout, pageInfo).then((tableData) => {
```
haxxmaxx, line 36: Doesn't this mean we run `manageData` everytime `isModal` or `constraints` changes? We only want tor run that if the data has changed.

src/table/Root.jsx
```diff
@@ -1,10 +1,14 @@
...
 import { ThemeProvider, StylesProvider } from '@material-ui/core/styles';
+import muiSetup from '../mui-setup';
 import TableWrapper from './TableWrapper';
 
+// This line is replaced by rollup with an import for internal builds
+const __OPIONAL_THEME_DEPS__ = {}; // eslint-disable-line no-underscore-dangle
+
 export function render(rootElement, props) {
-  const { muiParameters } = props;
+  const muiParameters = muiSetup(__OPIONAL_THEME_DEPS__);
```
haxxmaxx, line 11: Why is this faster? Every time `render` runs , we have to run `muiSetup`?

src/index.js
LiKang6688: I found that every time isModal or constraints changes, component() runs. so manageData need to run anyway.

src/table/Root.jsx
LiKang6688: component runs more often than render. when we put muiSetup() in component, it runs more often.

haxxmaxx: well, it is not supposed to. Since we are using `useState` and using `muiSetup` as the initial value, it should just run once?

src/index.js
haxxmaxx: Same thing here, `tableData` should only update when the data is changed, and then stored. I thought that was what `usePromise` did.

src/table/Root.jsx
LiKang6688: no, but we can use a function to lazily initialize the variable,  it will be executed only on the initial render. I will fix that.

src/index.js
LiKang6688: You are right. Gonna revert it.

src/table/Root.jsx
haxxmaxx: Yupp that would do the job :)

haxxmaxx (approved):

PR #183, opened 2021-04-25
fix focused cell can end up below the header

LiKang6688:


src/mui-setup.js
```diff
@@ -2,8 +2,26 @@ import { createGenerateClassName, createMuiTheme } from '@material-ui/core/style
...
   if (sproutBase.overrides) {
-    sproutBase.overrides.MuiTableContainer.root.height = 'calc(100% - 52px)';
     sproutBase.overrides.MuiListItem.root['&$selected'] = {};
+
+    sproutBase.overrides.MuiTableBody = {
+      root: {
+        display: 'block',
+        overflow: 'auto',
+      },
+    };
```
haxxmaxx, line 12: Is this the same?
```suggestion
    sproutBase.overrides.MuiTableBody.root = {
      display: 'block',
      overflow: 'auto',
    };
```

LiKang6688: no, not really.  'TypeError: Cannot set property 'root' of undefined' as MuiTableBody does not have object root

PR #184, opened 2021-04-26
fix focused cell can end up below the header

LiKang6688:


src/table/TableBodyWrapper.jsx
```diff
@@ -67,7 +67,7 @@ const TableBodyWrapper = ({
   }, []);
 
   return (
-    <TableBody className={classes.cellBase}>
+    <TableBody className={classes.cellBase} style={{ height: `${rootElement.clientHeight - 102}px` }}>
```
haxxmaxx, line 70: Why 102? Would be nice if you can get this in a class, but not in `cellBase` since it is not related to cell.

LiKang6688: the height of <thead> + the height of <TablePagination>, which are fixed value. Sure.

haxxmaxx: the height of what and what? 🙃

LiKang6688: 😂 sorry, I should preview it first. the height of` <TableHead>` + the height of` <TablePagination>`

haxxmaxx: OK, no problem. Well the height of the TableHead varies with the font size?

src/table/TableBodyWrapper.jsx
```diff
@@ -45,7 +48,7 @@ const TableBodyWrapper = ({
 }) => {
   const { rows, columns } = tableData;
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
-  const bodyStyle = useMemo(() => getBodyStyle(layout, theme), [layout, theme.name()]);
+  const bodyStyle = useMemo(() => getBodyStyle(layout, theme, rootElement), [layout, theme.name()]);
   const classes = useStyles(bodyStyle);
```
haxxmaxx, line 52: I'd say the `rootElement` is not part of `bodyStyle`. concat it when calling `useStyles` here instead.
`useStyles({ ...bodyStyles, rootElement.height })`

src/table/TableBodyWrapper.jsx
LiKang6688: good point. I need to update accordingly.

src/table/TableBodyWrapper.jsx
LiKang6688: sure

haxxmaxx (approved):

PR #185, opened 2021-04-26
fix: proper focus styling + refactor

haxxmaxx:
Moving a bunch of static css to the theme.
Also found two bugs and fixed them:
- The hover wasn't updating when changing colors. JSS prefers a different syntax.
- There was a double hover border. It should be the inset one only.

### DIFF
--- a/src/mui-setup.js
+++ b/src/mui-setup.js
@@ -12,17 +12,16 @@ export default function muiSetup(sproutBase) {
       },
     };
 
-    sproutBase.overrides.MuiTableSortLabel.root.color = 'inherit';
-    sproutBase.overrides.MuiTableSortLabel.root['&$active'].color = 'inherit';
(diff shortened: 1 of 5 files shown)


src/mui-setup.js
```diff
@@ -12,11 +12,6 @@ export default function muiSetup(sproutBase) {
       },
     };
 
-    sproutBase.overrides.MuiTableBody.root = {
```
LiKang6688, line 15: could you fix it, please?

LiKang6688 (changes requested):

src/mui-setup.js
haxxmaxx: You are setting the same thing the line above, or am I missing something?

LiKang6688: You are right.

LiKang6688 (approved):

PR #186, opened 2021-04-26
fix: update the height of table body when properties are changed

LiKang6688:


src/index.js
```diff
@@ -55,7 +55,7 @@ export default function supernova(env) {
             changeSortOrder,
           });
         }
-      }, [tableData, constraints, selectionsAPI.isModal(), theme.name()]);
+      }, [tableData, layout, constraints, selectionsAPI.isModal(), theme.name()]);
```
haxxmaxx, line 58: Not sure this is a good idea. It would run render twice every time layout changes. Does the height need to be exact? You can calculate a pretty good estimate based on the font size

LiKang6688: the height is not only affected by the font size but also the number of words

PR #187, opened 2021-04-27
fix: update the height of table body 

LiKang6688:
update the height of table body when properties are changed or resized

### DIFF
--- a/src/table/TableBodyWrapper.jsx
+++ b/src/table/TableBodyWrapper.jsx
@@ -10,6 +10,9 @@ import { bodyHandleKeyPress } from './cells/handle-key-press';
 import { handleClickToFocusBody } from './cells/handle-cell-focus';
 
 const useStyles = makeStyles({
+  body: {
+    height: ({ height }) => `${height}px`,
+  },
   cellBase: {
     '& td': {
       color: ({ color }) => color,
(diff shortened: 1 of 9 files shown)


haxxmaxx (approved):
Nice, I like this approach. Make sure to test if for all kinds of resizing and changing the header font

src/table/TableBodyWrapper.jsx
```diff
@@ -37,13 +37,13 @@ const TableBodyWrapper = ({
...
   focusedCellCoord,
   setShouldRefocus,
 }) => {
   const { rows, columns } = tableData;
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
   const bodyStyle = useMemo(() => getBodyStyle(layout, theme), [layout, theme.name()]);
-  const bodyHeight = rootElement.clientHeight - rootElement.getElementsByTagName('thead')[0]?.clientHeight - 52;
   const classes = useStyles({
     ...bodyStyle,
     height: bodyHeight,
```
haxxmaxx, line 49: Actually, I think it would be good if you add px here. I got into some trouble when concatenating strings inside makeStyles

src/table/TableBodyWrapper.jsx
```diff
@@ -5,32 +5,26 @@ import TableBody from '@material-ui/core/TableBody';
...
 import getCellRenderer from './cells/renderer';
-import { getBodyStyle, STYLING_DEFAULTS } from './styling-utils';
+import { getBodyStyle } from './styling-utils';
 import { bodyHandleKeyPress } from './cells/handle-key-press';
 import { handleClickToFocusBody } from './cells/handle-cell-focus';
 
 const useStyles = makeStyles({
   body: {
-    height: ({ height }) => `${height}px`,
+    height: ({ height }) => height,
```
LiKang6688, line 14: @haxxmaxx  is it what you want?

haxxmaxx: moving it to row 49

src/table/TableBodyWrapper.jsx
haxxmaxx: like this

```suggestion
    height: `${bodyHeight}px`,
```

haxxmaxx (approved):

PR #191, opened 2021-04-30
fix: add missing unit testing

LiKang6688:


src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -276,6 +283,16 @@ describe('handle-key-press', () => {
       expect(evt.stopPropagation).to.not.have.been.calledOnce;
       expect(selState.api.cancel).to.not.have.been.calledOnce;
     });
+
+    it('when other keys are pressed, should not do anything', () => {
+      evt.key = 'Control';
+      bodyHandleKeyPress(evt, rootElement, [rowIndex, colIndex], focusedCellCoord, selState, cell, selDispatch);
+      expect(evt.preventDefault).to.not.have.been.calledOnce;
```
haxxmaxx, line 290: I would actually use `not.have.been.called`. I think that this would pass if it was called twice

LiKang6688: Right. Good point. Gonna fix.

haxxmaxx (approved):

PR #192, opened 2021-04-30
refactor: after doing selections, changing the page info

LiKang6688:


src/table/TableWrapper.jsx
```diff
@@ -42,31 +51,26 @@ export default function TableWrapper(props) {
...
-
-  // should trigger reload
   const handleChangeRowsPerPage = (event) => {
     setRowsPerPage(+event.target.value);
     setPageInfo({ top: 0, height: +event.target.value });
     setPage(0);
   };
 
-  if (!rows.length && page > 0) {
-    handleChangePage(null, 0);
```
LiKang6688, line 59: when selections are made on another page than the first page and rows per page are bigger than the selected rows, in this case, it causes many renders. To reduce the number of re-renders, moving the logic to its parent, component()

haxxmaxx: I see your point. This could be a good improvement, but it is not risk free. So maybe not worth it so close to release

LiKang6688: i agree. not rush to merge.

src/index.js
```diff
@@ -37,23 +37,36 @@ export default function supernova(env) {
...
-            setPageInfo,
-            constraints,
-            selectionsAPI,
-            muiParameters,
-            theme,
-            changeSortOrder,
-          });
+          if (!tableData.rows.length && page > 0) {
+            // when selections are made on other page than first page and
+            // rows per page is bigger than the selected rows
```
haxxmaxx, line 49: This could basically happened anytime you don't get any data back given you are not on the first page, e.g. making selections in another chart

src/table/TableWrapper.jsx
haxxmaxx: I like this idea. But there are two problems though. Moving logic to index making hard to test and passing 4 more properties (minor problem). There is another way and that is always setting to first page when confirming selections. But that wouldn't help for outside updates. I'm still for it but maybe someone else should look at it as well.

PR #193, opened 2021-05-01
feature: focus the first cell after doing selection confirmation

LiKang6688:
to focus on the first cell of the body after selection confirmation instead of focusing on the first cell of the thead.

### DIFF
--- a/src/table/TableBodyWrapper.jsx
+++ b/src/table/TableBodyWrapper.jsx
@@ -40,6 +40,7 @@ const TableBodyWrapper = ({
   bodyHeight,
   focusedCellCoord,
   setShouldRefocus,
+  confirmedSelection,
 }) => {
   const { rows, columns } = tableData;
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
@@ -64,7 +65,7 @@ const TableBodyWrapper = ({
   }, [selectionsEnabled, columns.length]);
 
   useEffect(() => {
-    addSelectionListeners(selectionsAPI, selDispatch, setShouldRefocus);
+    addSelectionListeners(selectionsAPI, selDispatch, setShouldRefocus, confirmedSelection);
   }, []);
 
   return (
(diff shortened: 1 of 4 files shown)


haxxmaxx:
Why? I think we should have the same fallback cell independent of whether you're changing page or confirming selections. This just adds complexity.

LiKang6688:
> Why? I think we should have the same fallback cell independent of whether you're changing page or confirming selections. This just adds complexity.

I think that from the user's perspective, it is a better option. We can get feedback from A360.

haxxmaxx:
> > Why? I think we should have the same fallback cell independent of whether you're changing page or confirming selections. This just adds complexity.
> 
> I think that from the user's perspective, it is a better option. We can get feedback from A360.

Maybe, I've thought about that too. But then always go to the same cell when we reset it. That is just a change [here](https://github.com/qlik-oss/sn-table/blob/main/src/table/cells/handle-cell-focus.js#L34) to `[1, 0]`

LiKang6688:
> That is just a change [here](https://github.com/qlik-oss/sn-table/blob/main/src/table/cells/handle-cell-focus.js#L34) to `[1, 0]`

Are you sure that the change to [1, 0] works? I tried and it did not work. And that's the reason why I handle that in useDidUpdateEffect.

haxxmaxx:
> > That is just a change [here](https://github.com/qlik-oss/sn-table/blob/main/src/table/cells/handle-cell-focus.js#L34) to `[1, 0]`
> 
> Are you sure that the change to [1, 0] works? I tried and it did not work. And that's the reason why I handle that in useDidUpdateEffect.

Nope, I have not tried :) Come to think of it, there is a corner case where you get 0 rows, e.g. making selection in some other chart that results in 0 rows. I think it is better to refocus to the header to make sure there is always a cell with tabindex 0

LiKang6688:
> > > That is just a change [here](https://github.com/qlik-oss/sn-table/blob/main/src/table/cells/handle-cell-focus.js#L34) to `[1, 0]`
> > 
> > 
> > Are you sure that the change to [1, 0] works? I tried and it did not work. And that's the reason why I handle that in useDidUpdateEffect.
> 
> Nope, I have not tried :) Come to think of it, there is a corner case where you get 0 rows, e.g. making selection in some other chart that results in 0 rows. I think it is better to refocus to the header to make sure there is always a cell with tabindex 0

Really good point. I will close the PR then.

PR #195, opened 2021-05-03
feature: set hidden overflow text to display an ellipsis

LiKang6688:
Learning from the old straight table

Screenshot 2021-05-04 at 09 22 30

### DIFF
--- a/src/mui-setup.js
+++ b/src/mui-setup.js
@@ -16,6 +16,25 @@ export default function muiSetup(sproutBase) {
       display: 'table',
       width: '100%',
       tableLayout: 'fixed',
+      ...sproutBase.overrides.MuiTableRow.root,
+    };
+
+    const cellStyle = {
+      maxHeight: '3.7em',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+    };
+
+    sproutBase.overrides.MuiTableCell.root = {
+      ...cellStyle,
+      ...sproutBase.overrides.MuiTableCell.root,
+    };
+
+    sproutBase.overrides.MuiTableCell.head = {
+      '& div': {
+        ...cellStyle,
+      },
+      ...sproutBase.overrides.MuiTableCell.head,
     };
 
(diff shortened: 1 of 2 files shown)


src/table/TableHeadWrapper.jsx
```diff
@@ -64,7 +64,7 @@ export default function TableHeadWrapper({
                 }
                 tabIndex={-1}
               >
-                {column.label}
+                <div>{column.label}</div>
```
haxxmaxx, line 67: Hmm not sure about this. How does it affect the screen reader?

src/mui-setup.js
```diff
@@ -16,6 +16,25 @@ export default function muiSetup(sproutBase) {
       display: 'table',
       width: '100%',
       tableLayout: 'fixed',
+      ...sproutBase.overrides.MuiTableRow.root,
+    };
+
+    const cellStyle = {
+      maxHeight: '3.7em',
```
haxxmaxx, line 23: why 3.7?

LiKang6688: basically at least to show two rows in the table

src/table/TableHeadWrapper.jsx
LiKang6688: Screenshot 2021-05-04 at 10 21 55
looks good in sense-client

LiKang6688: Screenshot 2021-05-04 at 10 25 52
looks good in sense-client

haxxmaxx:
Tested a bit and it doesn't look great. I never get a horizontal scroll bar, so I guess there has to be some kind of min width?

LiKang6688:
> Tested a bit and it doesn't look great. I never get a horizontal scroll bar, so I guess there has to be some kind of min width?

no, it is a side effect from https://github.com/qlik-oss/sn-table/pull/184. Looking into solutions now.

LiKang6688:
> Tested a bit and it doesn't look great. I never get a horizontal scroll bar, so I guess there has to be some kind of min width?

it is an issue when I

PR #197, opened 2021-05-04
fix: revert table body style and scrollbar

LiKang6688:


haxxmaxx:
Should be some stuff in the spec files as well right?

LiKang6688:
> Should be some stuff in the spec files as well right?

looked at that. Seems not.

haxxmaxx (approved):

PR #198, opened 2021-05-04
feature: make fixed header look good

### LINKED ISSUE #201: When scrollin horizontally, the content in the header does not move but is hidden by its right content.
The leftmost header cell is partly hidden

Screenshot 2021-05-05 at 10 18 32

LiKang6688:
fixes #201

### DIFF
--- a/src/mui-setup.js
+++ b/src/mui-setup.js
@@ -12,7 +12,16 @@ export default function muiSetup(sproutBase) {
     sproutBase.overrides.MuiTableCell.root.height = 'auto';
     sproutBase.overrides.MuiTableCell.root.lineHeight = '130%';
     sproutBase.overrides.MuiTableCell.head.height = 'auto';
-    sproutBase.overrides.MuiTableCell.head.lineHeight = '150%';
+    sproutBase.overrides.MuiTableCell.head = {
+      lineHeight: '150%',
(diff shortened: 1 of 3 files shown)


src/mui-setup.js
```diff
@@ -12,7 +12,16 @@ export default function muiSetup(sproutBase) {
     sproutBase.overrides.MuiTableCell.root.height = 'auto';
     sproutBase.overrides.MuiTableCell.root.lineHeight = '130%';
     sproutBase.overrides.MuiTableCell.head.height = 'auto';
-    sproutBase.overrides.MuiTableCell.head.lineHeight = '150%';
+    sproutBase.overrides.MuiTableCell.head = {
+      lineHeight: '150%',
+      position: '-webkit-sticky' /* Safari */,
+      // eslint-disable-next-line no-dupe-keys
+      position: 'sticky',
```
haxxmaxx, line 19: isn't it just overridden?

LiKang6688: not really learning from https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky_positioning

haxxmaxx: Ahh, it is complaining bc this is a js object, not a css definition. Seems to work fine in safari

haxxmaxx (approved):

PR #199, opened 2021-05-04
feature: prevent page navigation on horizontal scroll

LiKang6688:
prevent the page going back and next to visited pages on a horizontal scroll

### DIFF
--- a/src/table/handle-scroll.js
+++ b/src/table/handle-scroll.js
@@ -1,0 +1,13 @@
+const handleScroll = (evt, tableSection) => {
+  evt.stopPropagation();
+  const max = tableSection.current.scrollWidth - tableSection.current.offsetWidth;
+  if (
+    max > 0 &&
+    (tableSection.current.scrollLeft + evt.deltaX < 0 || tableSection.current.scrollLeft + evt.deltaX > max)
+  ) {
+    evt.preventDefault();
+    tableSection.current.scrollLeft = Math.max(0, Math.min(max, tableSection.current.scrollLeft + evt.deltaX));
+  }
+};
+
+export default handleScroll;
--- a/src/table/handle-scroll.js
+++ b/src/table/handle-scroll.js
@@ -1,11 +1,11 @@
 const handleScroll = (evt, tableSection) => {
   evt.stopPropagation();
   let { scrollLeft } = tableSection.current;
-  const { deltaX } = evt;
+  const newScrollLeft = scrollLeft + evt.deltaX;
   const max = tableSection.current.scrollWidth - tableSection.current.offsetWidth;
-  if (max > 0 && (scrollLeft + deltaX < 0 || scrollLeft + deltaX > max)) {
+  if (max > 0 && (newScrollLeft < 0 || newScrollLeft > max)) {
     evt.preventDefault();
-    scrollLeft = Math.max(0, Math.min(max, scrollLeft + deltaX));
+    scrollLeft = Math.max(0, Math.min(max, newScrollLeft));
   }
 };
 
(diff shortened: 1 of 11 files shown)


haxxmaxx:
So do you have the same settings for changing page and scrolling horizontally? I don't and it works fine for me

LiKang6688:
> So do you have the same settings for changing page and scrolling horizontally? I don't and it works fine for me

the same settings. what do you mean?

haxxmaxx:
> > So do you have the same settings for changing page and scrolling horizontally? I don't and it works fine for me
> 
> the same settings. what do you mean?

On your mac. With the trackpad, I use two fingers to scroll and three to go back/forward

LiKang6688:
> > > So do you have the same settings for changing page and scrolling horizontally? I don't and it works fine for me
> > 
> > 
> > the same settings. what do you mean?
> 
> On your mac. With the trackpad, I use two fingers to scroll and three to go back/forward. I think the default settings might overlap

Yes, in the default setting, you can use two fingers to scroll and go back/forward.
Screenshot 2021-05-05 at 12 05 43

haxxmaxx:
> > > > So do you have the same settings for changing page and scrolling horizontally? I don't and it works fine for me
> > > 
> > > 
> > > the same settings. what do you mean?
> > 
> > 
> > On your mac. With the trackpad, I use two fingers to scroll and three to go back/forward. I think the default settings might overlap
> 
> Yes, in the default setting, you can use two fingers to scroll and go back/forward.
> Screenshot 2021-05-05 at 12 05 43

Yes, so I don't think we should compensate for that.

LiKang6688:
> > > > > So do you have the same settings for changing page and scrolling horizontally? I don't and it works fine for me
> > > > 
> > > > 
> > > > the same settings. what do you mean?
> > > 
> > > 
> > > On your mac. With the trackpad, I use two fingers to scroll and three to go back/forward. I think the default settings might overlap
> > 
> > 
> > Yes, in the default setting, you can use two fingers to scroll and go back/forward.
> > Screenshot 2021-05-05 at 12 05 43
> 
> Yes, so I don't think we should compensate for that.
you do not think the PR is needed? from my previous experience, it is a wanted feature from end-users.

src/table/handle-scroll.js
```diff
@@ -0,0 +1,13 @@
+const handleScroll = (evt, tableSection) => {
+  evt.stopPropagation();
+  const max = tableSection.current.scrollWidth - tableSection.current.offsetWidth;
+  if (
+    max > 0 &&
+    (tableSection.current.scrollLeft + evt.deltaX < 0 || tableSection.current.scrollLeft + evt.deltaX > max)
```
haxxmaxx, line 6: `tableSection.current.scrollLeft` is used 3 times, save it as a constant

LiKang6688: sure

src/table/handle-scroll.js
```diff
@@ -0,0 +1,12 @@
+const handleScroll = (evt, tableSection) => {
+  evt.stopPropagation();
+  let { scrollLeft } = tableSection.current;
+  const { deltaX } = evt;
+  const max = tableSection.current.scrollWidth - tableSection.current.offsetWidth;
+  if (max > 0 && (scrollLeft + deltaX < 0 || scrollLeft + deltaX > max)) {
+    evt.preventDefault();
+    scrollLeft = Math.max(0, Math.min(max, scrollLeft + deltaX));
```
haxxmaxx, line 8: I meant like this
```suggestion
  const newScrollLeft = tableSection.current.scrollLeft + evt.deltaX;
  const max = tableSection.current.scrollWidth - tableSection.current.offsetWidth;
  if (max > 0 && (newScrollLeft < 0 || newScrollLeft > max)) {
    evt.preventDefault();
    scrollLeft = Math.max(0, Math.min(max, newScrollLeft));
```

LiKang6688: scrollLeft in line 7 does not seem correct

haxxmaxx (approved):

PR #204, opened 2021-05-05
fix: no need to leave space for the TablePagination in edit mode

LiKang6688:


haxxmaxx (approved):
Ok, fine. This is as safe as it gets, go ahead and merge :)

LiKang6688:
> Ok, fine. This is as safe as it gets, go ahead and merge :)

Thanks

PR #205, opened 2021-05-06
fix: revert header scroll fix

haxxmaxx:
There was a reason for that lint issue, can't use the package in the mashup when you have multiple keys on an object apperently.


LiKang6688 (approved):

src/mui-setup.js
```diff
@@ -11,16 +11,7 @@ export default function muiSetup(sproutBase) {
     sproutBase.overrides.MuiTableCell.root.height = 'auto';
     sproutBase.overrides.MuiTableCell.root.lineHeight = '130%';
     sproutBase.overrides.MuiTableCell.head.height = 'auto';
-    sproutBase.overrides.MuiTableCell.head = {
-      lineHeight: '150%',
-      position: '-webkit-sticky' /* Safari */,
```
LiKang6688, line 16: or just remove this duplicate position

PR #206, opened 2021-05-06
fix: make the header look good when scrolling

LiKang6688:


haxxmaxx:
Sorry, `!important` is not a good way of solving this. I think our best bet is to try to make a change happen in material ui, probably by fixing it ourselves.

LiKang6688:
> Sorry, `!important` is not a good way of solving this. I think our best bet is to try to make a change happen in material ui, probably by fixing it ourselves.

Yes, I agree. I am creating a PR for MUI

Issue #166, opened 2021-04-08
When using NVDA screenreader it is not possible to move with arrow keys in the number of rows drop-down, until you press enter again


mlb76:
When using NVDA screenreader it is not possible to move with arrow keys in the number of rows drop-down. You can tab to the drop-down and open it, but you can not move between the options in the drop down. This is only an issue when using the NVDA screen reader in the table mashup.

haxxmaxx:
https://github.com/mui-org/material-ui/issues/26033 This is an issue on MUIs own documentation page, so put a bug on them. There is a work-around: just pressing space/enter one more time and the arrows work. That will have to do for now.

PR #218, opened 2021-05-26
fix: rows per page drop down is inaccessible

LiKang6688:
fixed:
Screenshot 2021-05-27 at 11 45 15

### DIFF
--- a/src/table/TableWrapper.jsx
+++ b/src/table/TableWrapper.jsx
@@ -98,7 +98,7 @@ export default function TableWrapper(props) {
         rowsPerPage={rowsPerPage}
         page={page}
         SelectProps={{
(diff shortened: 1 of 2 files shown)


src/table/TableWrapper.jsx
```diff
@@ -97,6 +97,10 @@ export default function TableWrapper(props) {
         count={size.qcy}
         rowsPerPage={rowsPerPage}
         page={page}
+        SelectProps={{
```
LiKang6688, line 100: following https://material-ui.com/components/tables/#custom-pagination-actions

haxxmaxx:
This would solve this issue as well: https://github.com/qlik-oss/sn-table/issues/166

haxxmaxx (approved):

PR #219, opened 2021-05-26
fix: columns can be sorted, but this is not indicated to screen reader users

LiKang6688:


src/table/TableHeadWrapper.jsx
```diff
@@ -65,6 +76,11 @@ export default function TableHeadWrapper({
                 tabIndex={-1}
               >
                 {column.label}
+                {layout.qHyperCube.qEffectiveInterColumnSortOrder[0] === columnIndex ? (
```
LiKang6688, line 79: following https://material-ui.com/components/tables/#sorting-amp-selecting

haxxmaxx (approved):

PR #220, opened 2021-05-26
fix: a scope of row. is missing in the first cell in each row

### LINKED ISSUE #217: Row headers not specified
The first cell in each row should be marked up as a <th> and be given a scope of "row."

LiKang6688:
fixes #217


haxxmaxx:
I don't know if we want this... The way I understand it you want a `scope` tag on every "header cell" for either a column or a row. Now the leftmost column is just like any other column, it does not necessarily specify what is in the row. In most cases, I assume, that you put the most descriptive dimension to the left. But not always. Take this example:
Screenshot 2021-05-27 at 11 19 55

It has the name of the car in the third column, which is the unique identifier. I think we should discuss this with A360

LiKang6688:
> I don't know if we want this... The way I understand it you want a `scope` tag on every "header cell" for either a column or a row. Now the leftmost column is just like any other column, it does not necessarily specify what is in the row. In most cases, I assume, that you put the most descriptive dimension to the left. But not always. Take this example:
> Screenshot 2021-05-27 at 11 19 55
> 
> It has the name of the car in the third column, which is the unique identifier. I think we should discuss this with A360

I see your point and I agree with you. it is like example 2 here https://www.w3.org/WAI/EO/Drafts/tutorials/tables/scope/.

haxxmaxx:
Updated the issue, let's keep it simple and always have it in the first column

haxxmaxx (approved):

PR #228, opened 2021-06-10
fix: columns use client side property for identification

LiKang6688:


src/ext.js
```diff
@@ -293,11 +288,6 @@ const definition = {
               defaultValue: true,
               show: false,
             },
-            cId: {
```
haxxmaxx, line 296: Above here, autosort is still included, so it will be declared twice

src/table/handle-data.js
```diff
@@ -20,7 +20,7 @@ export function getColumnInfo(layout, colIndex) {
       isDim,
       width: 200,
       label: info.qFallbackTitle,
-      id: info.cId,
+      id: index,
```
haxxmaxx, line 23: `colIndex` should work fine. All the key needs to be is a unique identifier in the current list. Not sure if necessarry, but I think a prefix would be nice, so `col-0` etc

src/table/handle-data.js
```diff
@@ -36,7 +36,7 @@ export default async function manageData(model, layout, pageInfo) {
   const matrix = dataPages[0].qMatrix;
 
   // using filter to remove hidden columns (represented with false)
-  const columns = columnOrder.map((c) => getColumnInfo(layout, c)).filter(Boolean);
+  const columns = columnOrder.map((c, index) => getColumnInfo(layout, c, index)).filter(Boolean);
   const rows = matrix.map((r, rowIdx) => {
     const row = { key: rowIdx };
```
haxxmaxx, line 41: Same thing here, `row-0`...

haxxmaxx, line 41: Same thing here, `row-0`...

src/table/handle-data.js
LiKang6688: sure

src/ext.js
LiKang6688: Deleted.

src/table/handle-data.js
```diff
@@ -36,9 +36,9 @@ export default async function manageData(model, layout, pageInfo) {
   const matrix = dataPages[0].qMatrix;
 
   // using filter to remove hidden columns (represented with false)
-  const columns = columnOrder.map((c) => getColumnInfo(layout, c)).filter(Boolean);
+  const columns = columnOrder.map((c, index) => getColumnInfo(layout, c, index)).filter(Boolean);
   const rows = matrix.map((r, rowIdx) => {
-    const row = { key: rowIdx };
+    const row = { key: `col-${rowIdx}` };
```
haxxmaxx, line 41:
```suggestion
    const row = { key: `row-${rowIdx}` };
```

src/table/handle-data.js
```diff
@@ -20,7 +20,7 @@ export function getColumnInfo(layout, colIndex) {
       isDim,
       width: 200,
       label: info.qFallbackTitle,
-      id: info.cId,
+      id: `col-${index}`,
```
haxxmaxx, line 23: I meant that you should use `colIndex` instead of `index`

haxxmaxx (approved):

PR #230, opened 2021-06-16
feat(a11y): the screen reader should always says the column name first and then the row content

LiKang6688:
PR:
Screenshot 2021-06-16 at 15 34 20


haxxmaxx (approved):
LGTM

PR #235, opened 2021-07-05
chore: update dependency

LiKang6688:


quanho (approved):

haxxmaxx (approved):

haxxmaxx:
You need to change the install step as I did here: https://github.com/qlik-trial/sn-kpi/pull/102/files

LiKang6688:
> You need to change the install step as I did here: https://github.com/qlik-trial/sn-kpi/pull/102/files

Perfect! Thank you. I was exactly looking for it.

rollup-core.config.js
```diff
@@ -59,7 +59,7 @@ export default {
     }),
     checkInternals()
       ? replace({
-          'const __OPIONAL_THEME_DEPS__ = {};': "import { sproutBase } from '@qlik/sprout-theme';",
+          'const __OPIONAL_THEME_DEPS__ = {};': "import { sproutBase } from '@qlik-trial/sprout-theme';",
```
niekvanstaveren, line 62: Do we still need to do this if we remove the dependency all together?

rollup.config.js
```diff
@@ -58,7 +58,7 @@ export default {
     }),
     checkInternals()
       ? replace({
-          'const __OPIONAL_THEME_DEPS__ = {};': "import { sproutBase } from '@qlik/sprout-theme';",
+          'const __OPIONAL_THEME_DEPS__ = {};': "import { sproutBase } from '@qlik-trial/sprout-theme';",
```
niekvanstaveren, line 61: Same here

rollup-core.config.js
LiKang6688: yes, we do. If I understand it correctly when the sn-table is used as a mashup or extension, it will check node_modules/@qlik-trial exists or not first, if it exists, it does use the @qlik-trial/sprout-theme from node_modules.

LiKang6688: I found that that's one of the reasons, in the sn-table project, we do not directly use `nebula build,` instead, we use `rollup -c rollup.config.js && rollup -c rollup-core.config.js`

niekvanstaveren: Yes that is true. But now you are removing the sprout-theme dependency all together. So the check becomes useless since it will never be there right?

LiKang6688: it also checks node_modules in other projects, right? when it is used as a mashup or extension, it checks node_modules/@qlik-trial in sense-client.

niekvanstaveren: I'm not sure what you mean with "used as a mashup"...

These config files are used when running the `yarn build` and `yarn build:watch` commands. Those are run to package the chart. When packaging this you are not running in sense-client in any way so not sure how it checks the node_modules of sense-client. Am I missing something here?

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - restore_cache: *restore_yarn_cache
       - run: 
           name: Install dependencies
-          command: yarn install --frozen-lockfile --ignore-optional
```
LiKang6688: yarn install --ignore-optional does not work as expected. so removing optionalDependencies.

PR #243, opened 2021-09-01
[fix/#161]: fixed table navigation move to the next/previous line

a-m-dev:
in this PR, I have tweaked the functionality for `arrowKeysNavigation` a bit in order to support the row navigation when reaching to the end or beginning of each row in table. 

also since the only purpose of this PR is to navigate between rows in table there are two constrains for it: 
1. **when we reach to the beginning of the table (`[0, 0]` coordinates)**
2. **when we reach to the end of the table (`[rowLenght - 1, colLength - 1]` coordinates)**

in those cases we are still making user to remain in those cells until they actually want to move to the next element by using `shift + tab` in order to move to the previous focused element or by just using `tab` in order to move to the next element on the screen.

### DIFF
--- a/src/table/cells/__tests__/handle-key-press.spec.js
+++ b/src/table/cells/__tests__/handle-key-press.spec.js
@@ -89,6 +89,50 @@ describe('handle-key-press', () => {
       expect(nextRow).to.eql(0);
       expect(nextCol).to.eql(0);
     });
+
+    it('should move to the next row when we reach to the end of the current row', () => {
+      evt.key = 'ArrowRight';
+      rowAndColumnCount.rowCount = 3;
+      rowAndColumnCount.columnCount = 3;
+      rowIndex = 1;
+      colIndex = 3;
+      const [nextRow, nextCol] = arrowKeysNavigation(evt, rowAndColumnCount, [rowIndex, colIndex]);
+      expect(nextRow).to.eql(2);
+      expect(nextCol).to.eql(0);
+    });
+
+    it('should move to the prev row when we reach to the beganing of the current row', () => {
+      evt.key = 'ArrowLeft';
+      rowAndColumnCount.rowCount = 3;
+      rowAndColumnCount.columnCount = 3;
+      rowIndex = 2;
+      colIndex = 0;
+      const [nextRow, nextCol] = arrowKeysNavigation(evt, rowAndColumnCount, [rowIndex, colIndex]);
+      expect(nextRow).to.eql(1);
+      expect(nextCol).to.eql(2);
+    });
+
+    it('should stay at the first row and first col of table when we reached to the beganing of the table', () => {
+      evt.key = 'ArrowLeft';
+      rowAndColumnCount.rowCount = 2;
+      rowAndColumnCount.columnCount = 2;
+      rowIndex = 0;
+      colIndex = 0;
+      const [nextRow, nextCol] = arrowKeysNavigation(evt, rowAndColumnCount, [rowIndex, colIndex]);
+      expect(nextRow).to.eql(0);
+      expect(nextCol).to.eql(0);
+    });
+
+    it('should stay at the end row and end col of table when we reached to the end of the table', () => {
+      evt.key = 'ArrowRight';
+      rowAndColumnCount.rowCount = 2;
+      rowAndColumnCount.columnCount = 2;
+      rowIndex = 2;
+      colIndex = 2;
+      const [nextRow, nextCol] = arrowKeysNavigation(evt, rowAndColumnCount, [rowIndex, colIndex]);
+      expect(nextRow).to.eql(2);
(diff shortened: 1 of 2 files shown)


src/table/cells/handle-key-press.js
```diff
@@ -27,12 +27,32 @@ export const arrowKeysNavigation = (evt, rowAndColumnCount, cellCoord, selState)
...
+    case 'ArrowRight': {
+      if (nextCol < rowAndColumnCount.columnCount - 1) !isSelectedTable && nextCol++;
+      else if (!isSelectedTable) {
+        // move to the next row
+        // if we are at the end of the table do nothing
+        if (nextRow + 1 > rowAndColumnCount.rowCount - 1) break;
+        else {
+          nextRow = nextRow + 1;
+          nextCol = 0;
+        }
```
haxxmaxx: The logic here is quite confusing. First off, you are mixing `if` and `&&` outside the if. On row 31, make it into one expression inside the if and add {} to wrap `nextCol++`. Second, I think the second case can be reduced to one if statement:
`!isSelectableTable && nextRow + 1 < rowAndColumnCount.rowCount - 1` (not sure if you need the + 1)

You can also use `nextRow++` on row 37 for consistency

Same thing goes for the `arrowLeft` case

a-m-dev: Changes Done!

src/table/cells/handle-key-press.js
```diff
@@ -28,10 +28,20 @@ export const arrowKeysNavigation = (evt, rowAndColumnCount, cellCoord, selState)
       nextRow > 0 && (!isSelectedTable || nextRow !== 1) && nextRow--;
       break;
     case 'ArrowRight':
-      nextCol < rowAndColumnCount.columnCount - 1 && !isSelectedTable && nextCol++;
+      if (nextCol < rowAndColumnCount.columnCount - 1 && !isSelectedTable) {
+        nextCol++;
+      } else if (!isSelectedTable && nextRow < rowAndColumnCount.rowCount - 1) {
```
haxxmaxx, line 33: Ok being a bit picky here, but you are checking `!isSelectedTable` twice. Cleaner IMO to first have an if checking `!isSelectedTable`, wrapping the other two cases. Minor detail, it is fine as it is

haxxmaxx:
Need some new unit test cases, then it should look pretty good

src/table/cells/handle-key-press.js
a-m-dev: you mean if we added some code to check if `isSelectedTable` is going to be true then break from that case?

a-m-dev: Done, I wrapper the both cases with `if (!isSelectedTable) { ... } ` first

haxxmaxx: yes, but I'd say check `!isSelectableTable` and do stuff in that case

a-m-dev:
**4 unit tests added:** 
1. one for moving to the next row by pressing right key
2. one for moving to the previous row by pressing left key
3. one for staying in the very first cell of table if already was there and keep pressing the left key
4. and finally last for for staying into the very last cell if already was there and keep pressing right key

haxxmaxx:
> **4 unit tests added:**
> 
> 1. one for moving to the next row by pressing right key
> 2. one for moving to the previous row by pressing left key
> 3. one for staying in the very first cell of table if already was there and keep pressing the left key
> 4. and finally last for for staying into the very last cell if already was there and keep pressing right key

Nice! 3 and 4 actually already exist, but they are only testing on one cell which is kind of weird. Remove the old ones I'd would say since yours are better

a-m-dev:
actually i did not removed them, i refined them a bit, since we were creating a table of 1 by 1 matrix in tests, the tests we passing for them without checking the actual functionality of them, so i did expanded it for a 1 by 3 matrix ( lets say ) and now we can navigate between cols of the test tables and they are passing now perfectly.

haxxmaxx (approved):

src/table/cells/__tests__/handle-key-press.spec.js
```diff
@@ -89,6 +75,50 @@ describe('handle-key-press', () => {
...
+      rowAndColumnCount.rowCount = 2;
+      rowAndColumnCount.columnCount = 2;
+      rowIndex = 0;
+      colIndex = 0;
+      const [nextRow, nextCol] = arrowKeysNavigation(evt, rowAndColumnCount, [rowIndex, colIndex]);
+      expect(nextRow).to.eql(0);
+      expect(nextCol).to.eql(0);
+    });
+
+    it.only('should stay at the end row and end col of table when we reached to the end of the table', () => {
```
haxxmaxx, line 112: remove only

a-m-dev: ✅ Fixed

PR #246, opened 2021-09-02
[FIX#216]: fix for proper `aria-*` attributes in table sorting

a-m-dev:
This PR will add `aria-sort` and `aria-label` attributes for table head cells while sorting. 
currently it has been tested with a chrome extension ( this one: https://chrome.google.com/webstore/detail/empty-title/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en-GB ) and also with Mac OS's default voice over feature. 

Unit Tests:
added 2 tests for: 
1. check to see if `aria-sort` property has been rendered properly 
2. check to see if `aria-label` property has been rendered properly


src/table/TableHeadWrapper.jsx
```diff
@@ -46,12 +51,18 @@ export default function TableHeadWrapper({
...
+          const isTableSortActive = layout.qHyperCube.qEffectiveInterColumnSortOrder[0] === columnIndex;
+
           return (
             <TableCell
               key={column.id}
               align={column.align}
               className={`${classes.head} sn-table-head-cell sn-table-cell`}
               tabIndex={tabIndex}
+              aria-label={ariaLabelText}
+              aria-sort={column.sortDirection}
```
haxxmaxx, line 65: What does the screen reader say here? The direction is now in three places. `aria-label`, `aria-sort` and the hidden label. AFAIK it is better to have a hidden label, that's why Li implemented that. Also, what does it say when the sort isn't active on a column? Is it sorted on that column, but not primarily?

a-m-dev: For each column there is a default sorting order that comes from `tableData` i guess, and screen reader reads through the `aria-label` not `aria-sort`. 
when sort is not active on a column, it will tell user the default order of the column that comes from `tableData`.

btw, this is the default behaviour of the previous table which i have implemented here.

haxxmaxx: So does aria-sort do anything? And you don't hear the hidden label anymore?
I think we need to include if it is actively sorted or not. There is no indication on which column it is sorted on primarily.

a-m-dev: I just included the `aria-sort` in order to match with the older version since you mentioned that the previous table is the complete version of this one. but seems like you are right, it does not do any thing so far if the screen reader is going to read the `aria-label` only.

One suggestion that i can make here is that we can add the current (active) sort to the `aria-label` attributes for columns. by doing this way, user will always here a sentence like this: 
`columnName. Sorted ascending. **currently table sorted on ColumnX in ascending way.** Press space key to sort on this column.`

haxxmaxx: Ah ok. Sorry that I was unclear: I guess we want the old behaviour more or less, not the implementation necessarily.

I think that aria-label overrides the other announcements, that's why it is not best practice to use in some cases. So if we keep aria-sort, then we could announce the instructions using the hidden label. I had a look here: https://adrianroselli.com/2021/04/sortable-table-columns.html#Announce Seems like some screen readers announce when you update aria-sort, some does not (including NVDA which we support). So we definitely need to test this with NVDA.

About the active column, I misunderstood that a bit. There is only one column which is sorted. The other ones should have `aria-sort="none"`, which is default (so actually no need to add aria-sort in those cases at all). I don't think we need to say which column that is sorted when you are on a non-sorted column.

src/table/TableHeadWrapper.jsx
```diff
@@ -43,15 +48,22 @@ export default function TableHeadWrapper({
 
   return (
     <TableHead>
-      <TableRow className="sn-table-row">
+      <TableRow className="sn-table-row" role="row">
```
haxxmaxx, line 51: why?

src/table/TableHeadWrapper.jsx
```diff
@@ -65,22 +77,13 @@ export default function TableHeadWrapper({
...
-                tabIndex={-1}
-              >
+              <TableSortLabel active={isCurrentColumnActive} direction={column.sortDirection} tabIndex={-1}>
                 {column.label}
-                {layout.qHyperCube.qEffectiveInterColumnSortOrder[0] === columnIndex ? (
-                  <span className={classes.visuallyHidden}>
-                    {column.sortDirection === 'desc' ? 'sorted descending' : 'sorted ascending'}
-                  </span>
-                ) : null}
+                <span className={classes.visuallyHidden}>{currentSortDir}. Press space to sort on this column.</span>
```
haxxmaxx, line 86: No need to have space in the string, it is implied since it is a button. Also, need to take `isCurrentColumnActive` into account. Shouldn't announce ascending/descending if `isCurrentColumnActive` is false

src/table/TableWrapper.jsx
```diff
@@ -85,7 +85,7 @@ export default function TableWrapper(props) {
       onKeyDown={(evt) => updatePage(evt, size.qcy, page, rowsPerPage, handleChangePage, setShouldRefocus)}
     >
       <TableContainer ref={tableSection} className={classes[containerMode]}>
-        <Table stickyHeader aria-label={`showing ${rows.length + 1} rows and ${columns.length} columns`}>
+        <Table stickyHeader role="table" aria-label={`showing ${rows.length + 1} rows and ${columns.length} columns`}>
```
haxxmaxx, line 88: This is a mui table, don't think you can add a role like that, and it shouldn't be necessary

src/table/TableHeadWrapper.jsx
a-m-dev: ✅ fixed

src/table/TableWrapper.jsx
a-m-dev: ✅ fixed

src/table/TableHeadWrapper.jsx
haxxmaxx: no I meant the word `space`. A button should listen to both space and enter. And since the screen reader says button, no need to explain what key to use

src/table/TableHeadWrapper.jsx
```diff
@@ -46,12 +51,19 @@ function TableHeadWrapper({
...
+
           return (
             <TableCell
               key={column.id}
               align={column.align}
               className={`${classes.head} sn-table-head-cell sn-table-cell`}
               tabIndex={tabIndex}
+              role="button"
+              scope="col"
+              aria-sort={isCurrentColumnActive ? `${column.sortDirection}ending` : null}
```
haxxmaxx, line 65: creative 👍

src/table/__tests__/TableHeadWrapper.spec.jsx
```diff
@@ -143,4 +143,104 @@ describe('<TableHeadWrapper />', () => {
 
     expect(handleCellFocus.handleClickToFocusHead).to.have.been.calledOnce;
   });
+
+  it('should has `scope` property set to col', () => {
```
haxxmaxx, line 147:
```suggestion
  it('should have `scope` property set to col', () => {
```

src/table/__tests__/TableHeadWrapper.spec.jsx
```diff
@@ -143,4 +143,104 @@ describe('<TableHeadWrapper />', () => {
...
+    );
+
+    const firstColQuery = queryByText(tableData.columns[0].label).closest('th');
+    const secondColQuery = queryByText(tableData.columns[1].label).closest('th');
+
+    expect(firstColQuery).to.have.attribute('aria-sort', 'ascending');
+    expect(secondColQuery).to.not.have.attribute('aria-sort');
+  });
+
+  it('should has `aria-pressed` property', () => {
```
haxxmaxx, line 183: You can make all of these tests into one. no need to re-render every time since there is no difference

a-m-dev: Sure, I was trying to stick to one of the testing rules which indicated that: "each test case might only check for one thing"

but if you thinking that could be redundant to do at some point, i could combine them as well 👌

src/table/__tests__/TableHeadWrapper.spec.jsx
a-m-dev: ✅ Fixed

src/table/TableHeadWrapper.jsx
a-m-dev: 🙏

src/table/__tests__/TableHeadWrapper.spec.jsx
a-m-dev: ✅ Fixed

haxxmaxx (approved):

LiKang6688:
I tested the PR in windows PTE and found the keyboard navigation would get stuck when it is back to the header.

https://user-images.githubusercontent.com/4471489/133332782-80a9a9ee-38a0-4bbd-8931-22b69335f472.mov

I am not sure why yet.

src/table/TableHeadWrapper.jsx
LiKang6688: Screenshot 2021-09-15 at 13 23 58

got a warning message from eslint:
The attribute aria-sort is not supported by the role button.

a-m-dev: as far as i know, we should have put role as `columnheader` and that is the recommended way to handle it. But as I was talking with Max, the a360 side recommending us to have a role button for that since we have sorting going on there.

a-m-dev:
i'm currently dont have access to a windows machine, How i can test that senario @LiKang6688?

a-m-dev:
> I tested the PR in windows PTE and found the keyboard navigation would get stuck when it is back to the header.
> 
>  move-issue.mov 
> I am not sure why yet.


I was testing another issue on windows PTE and just figured out since the referred issue by Li has been merged before, i tested that as well and figured out it's working correctly 

https://user-images.githubusercontent.com/29652890/134015495-d934362a-4011-405b-9899-ba987cf1ce1a.mov

src/table/TableHeadWrapper.jsx
```diff
@@ -65,22 +78,16 @@ function TableHeadWrapper({
...
+              <TableSortLabel active={isCurrentColumnActive} direction={column.sortDirection} tabIndex={-1}>
                 {column.label}
-                {layout.qHyperCube.qEffectiveInterColumnSortOrder[0] === columnIndex ? (
-                  <span className={classes.visuallyHidden}>
-                    {column.sortDirection === 'desc' ? 'sorted descending' : 'sorted ascending'}
-                  </span>
-                ) : null}
+                <span className={classes.visuallyHidden} data-testid={`VHL-for-col-${columnIndex}`}>
+                  {isCurrentColumnActive && `${currentSortDir}. `}
+                  Press space or enter to sort on the column.
```
haxxmaxx, line 89: I think I wrote `the column` to you, but I meant `this column` 🙃

src/table/TableHeadWrapper.jsx
```diff
@@ -15,6 +15,10 @@ const useStyles = makeStyles({
     fontSize: ({ fontSize }) => fontSize,
     padding: ({ padding }) => padding,
   },
+  tableHeadBtn: {
```
haxxmaxx, line 18: Not used, remove

src/table/TableHeadWrapper.jsx
a-m-dev: ✅ Fixed

src/table/TableHeadWrapper.jsx
a-m-dev: ✅ Fixed

haxxmaxx (approved):

src/table/TableHeadWrapper.jsx
```diff
@@ -35,24 +35,34 @@ function TableHeadWrapper({
...
+          const isCurrentColumnActive =
+            layout.qHyperCube.qEffectiveInterColumnSortOrder[0] === tableData.columnOrder[columnIndex];
+
           return (
             <TableCell
               key={column.id}
               align={column.align}
               className={`${classes.head} sn-table-head-cell sn-table-cell`}
               tabIndex={tabIndex}
+              scope="col"
```
haxxmaxx, line 65: I thought this was added automatically?

src/table/__tests__/TableHeadWrapper.spec.jsx
```diff
@@ -152,10 +160,52 @@ describe('<TableHeadWrapper />', () => {
...
+        translator={translator}
+      />
+    );
+
+    const firstColHiddenLabel = queryByTestId('VHL-for-col-0');
+    const secondColHiddenLabel = queryByTestId('VHL-for-col-1');
+
+    expect(firstColHiddenLabel).to.have.text('SNTable.SortLabel.SortedAscending. SNTable.SortLabel.PressSpaceToSort');
+    expect(secondColHiddenLabel).to.have.text('SNTable.SortLabel.PressSpaceToSort');
+  });
```
haxxmaxx, line 210: You can merge these to one to reduce one render

PR #247, opened 2021-09-03
[Fix#179]: Fixed the a11y issue with table head while moving up in table with keyboard

a-m-dev:
in this PR the issue of hiding focused row behind the table head, while moving upward in table has been fixed.

I tries to remove the use case for `useRef([0, 0])` and kept tracking of `focusedCellCoord` in a state. 
**In order to prevent extra re-renders of `<TableHead />` and `<TableBody />` components, I used `React.memo` for them with a bit of extra check for preventing those re-renders.** 

The result is equivalent in both cases (with and without using ref for keep tracking of the `focusedCellCoord` coordination ).

Here is the final fixed result: 

https://user-images.githubusercontent.com/29652890/132035880-ecf8f4fb-3855-4188-944d-c4dff4933f9f.mov


src/table/TableBodyWrapper.jsx
```diff
@@ -126,8 +129,13 @@ TableBodyWrapper.propTypes = {
...
+  handleFocusedCellCordsUpd: PropTypes.func.isRequired,
   setShouldRefocus: PropTypes.func.isRequired,
+  handleSetMovingTop: PropTypes.func.isRequired,
 };
 
-export default React.memo(TableBodyWrapper);
+export default memo(TableBodyWrapper, (prevProps, nextProps) => {
+  if (prevProps.focusedCellCoord.toString() !== nextProps.focusedCellCoord.toString()) return true;
+  else return false;
+});
```
haxxmaxx, line 141:
```suggestion
export default memo(TableBodyWrapper, (prevProps, nextProps) => prevProps.focusedCellCoord.toString() !== nextProps.focusedCellCoord.toString());
```

haxxmaxx: So this does not update the component, when focusCellCoord has changed? Could there be cases when we want to update when it has changed? When testing it, it seems to work, but we need more manual testing for sure.

src/table/TableHeadWrapper.jsx
```diff
@@ -97,6 +98,11 @@ TableHeadWrapper.propTypes = {
...
   constraints: PropTypes.object.isRequired,
-  focusedCellCoord: PropTypes.object.isRequired,
+  focusedCellCoord: PropTypes.arrayOf(PropTypes.number),
   selectionsAPI: PropTypes.object.isRequired,
 };
+
+export default memo(TableHeadWrapper, (prevProps, nextProps) => {
+  if (prevProps.focusedCellCoord.toString() !== nextProps.focusedCellCoord.toString()) return true;
+  else return false;
+});
```
haxxmaxx, line 108: Same thing here, simplify

src/table/TableWrapper.jsx
```diff
@@ -73,10 +75,44 @@ export default function TableWrapper(props) {
...
+    } else if (isMovingTop) {
+      setIsMovingTop(false);
+
+      const rowElements = rootElement.getElementsByClassName('sn-table-row');
+      const [x, y] = focusedCellCoordsState;
+      const cell = rowElements[x]?.getElementsByClassName('sn-table-cell')[y];
+
+      if (cell.offsetTop - cell.offsetHeight * 2 < tableSection.current.scrollTop) {
+        tableSection.current.scrollTo({
+          top: tableSection.current.scrollTop - cell.offsetHeight || 0,
```
haxxmaxx, line 93: why `|| 0`? Could `tableSection.current.scrollTop - cell.offsetHeight` become NaN or something?

src/table/TableWrapper.jsx
```diff
@@ -73,10 +75,44 @@ export default function TableWrapper(props) {
...
+        });
+      }
+    }
+  }, [tableSection, focusedCellCoordsState, isMovingTop, rootElement]);
+
+  const handleFocusedCellCordsUpd = useCallback((newCoords) => {
+    setFocusedCellCoordsState(newCoords);
+  }, []);
+
+  const handleSetMovingTop = useCallback((val) => setIsMovingTop(val), []);
```
haxxmaxx, line 104: Do these make any difference, compared to passing just the setState?

src/table/TableWrapper.jsx
```diff
@@ -73,10 +75,44 @@ export default function TableWrapper(props) {
...
+      const cell = rowElements[x]?.getElementsByClassName('sn-table-cell')[y];
+
+      if (cell.offsetTop - cell.offsetHeight * 2 < tableSection.current.scrollTop) {
+        tableSection.current.scrollTo({
+          top: tableSection.current.scrollTop - cell.offsetHeight || 0,
+          behavior: 'smooth',
+        });
+      }
+    }
+  }, [tableSection, focusedCellCoordsState, isMovingTop, rootElement]);
```
haxxmaxx, line 98: Do we want to listen to all of these? e.g. pretty sure `rootElement` wont change unless you remount the whole thing. Why isn't just `focusedCellCoordsState` enough?

src/table/cells/handle-key-press.js
```diff
@@ -59,31 +59,31 @@ export const preventDefaultBehavior = (evt) => {
...
-  focusedCellCoord,
   changeSortOrder,
   layout,
   isDim,
-  isAnalysisMode
+  isAnalysisMode,
+  setFocusedCellCoordsState
 ) => {
   switch (evt.key) {
     case 'ArrowUp':
```
haxxmaxx, line 82: Just realized we don't need to do anything for arrowUp, since you are. already at the top. You can remove it :)

src/table/cells/handle-key-press.js
```diff
@@ -102,18 +102,20 @@ export const bodyHandleKeyPress = (
...
+  handleSetMovingTop
 ) => {
   switch (evt.key) {
     case 'ArrowUp':
+      handleSetMovingTop(true);
     case 'ArrowDown':
     case 'ArrowRight':
     case 'ArrowLeft': {
-      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, focusedCellCoord, selState);
+      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, selState, setFocusedCellCoordsState);
```
haxxmaxx, line 118: This is a bit weird, is this valid syntax? Because for arrowUp it runs both `handleSetMovingTop` and `moveFocus` right?

src/table/TableBodyWrapper.jsx
a-m-dev: Yes, when `focusCellCoord` state changes, we are not going to re-render our `<TableHead />` and `<TableBody />` component. Since we don't actually need the value of it and the only thing we need inside of these component is the setter function for that state which has been used inside of some helper function like `headHandleKeyPress`

src/table/TableHeadWrapper.jsx
a-m-dev: Done ✅

src/table/TableWrapper.jsx
a-m-dev: Good catch, 
My intention here was about the calculated value could get below 0 and i was preventing it. i figured out that the compassion is wrong here since negative value when converting to boolean is still a number and would give me true.

I need to change this a bit like check if value actually getting below 0, then fallback to 0 instead.

a-m-dev: I'll fix it asap

src/table/TableWrapper.jsx
a-m-dev: you are right, in our case it does not makes any difference if we pass the state setter by itself. 

i'll fix it asap 🙏

src/table/TableWrapper.jsx
a-m-dev: Fixed ✅
I think probably i was afraid of some eslint related error ( by habit ) that i put it there

src/table/TableWrapper.jsx
```diff
@@ -73,10 +75,44 @@ export default function TableWrapper(props) {
     };
   }, []);
 
+  useEffect(() => {
```
haxxmaxx, line 78: If possible, would be nice to break the content of this useEffect out of the TableWrapper. Don't want to crowd this file and need to have it testable.

src/table/cells/handle-key-press.js
a-m-dev: Good Catch ✅
it's because we already are at the head of table we don't need to move upper that that 👌

src/table/cells/handle-key-press.js
a-m-dev: Yes, it's totally fine, I want to run `handleSetMovingTop(true)` function specifically (only) for 'ArrowUp' case and not for other ones. since i don't put `break` for it will will run that function for that case and JS interpreter will move to other cases until it reaches to the break statement. 

it is the same concept of putting multiple cases together and running some part of code for all.

src/table/TableWrapper.jsx
a-m-dev: Sure, I'll create a function for it to isolate the work that it does

src/table/cells/handle-key-press.js
haxxmaxx: cool, didn't know that. Looks like you need a comment thoough https://eslint.org/docs/rules/no-fallthrough

rollup.config.js
```diff
@@ -33,7 +33,7 @@ export default {
     name: moduleName,
     format: 'umd',
     exports: 'default',
-    sourcemap: false,
+    sourcemap: true,
```
haxxmaxx, line 36: remove

src/table/TableWrapper.jsx
```diff
@@ -35,7 +35,8 @@ export default function TableWrapper(props) {
   const [tableWidth, setTableWidth] = useState();
   const [page, setPage] = useState(0);
   const [rowsPerPage, setRowsPerPage] = useState(100);
-  const focusedCellCoord = useRef([0, 0]);
+  const [focusedCellCoordsState, setFocusedCellCoordsState] = useState([0, 0]);
```
haxxmaxx, line 38: No need to have state in the name

rollup.config.js
a-m-dev: Fixed ✅

src/table/TableWrapper.jsx
a-m-dev: fixed ✅

a-m-dev:
Refined the code and added 3 unit tests to cover the functionality: 
1. one for the time that ref is still not set up yet 
2. one for the time that we are in 2nd most top row that should detect and dont hide the focused row behind the `<TableHead />`
3. and last for the time we are navigating toward top in the table by keyboard

also in this amended PR, the `setMovingTop` state was removed and the behaviour is still the same

src/table/TableWrapper.jsx
```diff
@@ -86,8 +101,17 @@ export default function TableWrapper(props) {
     >
       <TableContainer ref={tableSection} className={classes[containerMode]}>
         <Table stickyHeader aria-label={`showing ${rows.length + 1} rows and ${columns.length} columns`}>
-          <TableHeadWrapper {...props} focusedCellCoord={focusedCellCoord} />
-          <TableBodyWrapper {...props} focusedCellCoord={focusedCellCoord} setShouldRefocus={setShouldRefocus} />
+          <TableHeadWrapper
+            {...props}
+            focusedCellCoord={focusedCellCoord}
+            handleFocusedCellCordsUpd={setfocusedCellCoord}
```
haxxmaxx, line 107: I'd say no need to rename it

src/table/TableBodyWrapper.jsx
```diff
@@ -126,8 +127,12 @@ TableBodyWrapper.propTypes = {
...
-  focusedCellCoord: PropTypes.object.isRequired,
+  focusedCellCoord: PropTypes.arrayOf(PropTypes.number),
+  handleFocusedCellCordsUpd: PropTypes.func.isRequired,
   setShouldRefocus: PropTypes.func.isRequired,
 };
 
-export default React.memo(TableBodyWrapper);
+export default memo(
+  TableBodyWrapper,
+  (prevProps, nextProps) => prevProps.focusedCellCoord.toString() !== nextProps.focusedCellCoord.toString()
```
haxxmaxx, line 137: I'd say this needs a comment. Unless you know about this fix, it's hard to understand when it updates or not. same thing in headWrapper

src/table/TableBodyWrapper.jsx
```diff
@@ -126,8 +127,12 @@ TableBodyWrapper.propTypes = {
   selectionsAPI: PropTypes.object.isRequired,
   layout: PropTypes.object.isRequired,
   theme: PropTypes.object.isRequired,
-  focusedCellCoord: PropTypes.object.isRequired,
+  focusedCellCoord: PropTypes.arrayOf(PropTypes.number),
```
haxxmaxx, line 130: isRequired

src/table/TableHeadWrapper.jsx
```diff
@@ -97,6 +98,11 @@ TableHeadWrapper.propTypes = {
   layout: PropTypes.object.isRequired,
   changeSortOrder: PropTypes.func.isRequired,
   constraints: PropTypes.object.isRequired,
-  focusedCellCoord: PropTypes.object.isRequired,
+  focusedCellCoord: PropTypes.arrayOf(PropTypes.number),
```
haxxmaxx, line 101: isRequired

src/table/TableWrapper.jsx
```diff
@@ -4,12 +4,12 @@ import Table from '@material-ui/core/Table';
 import TableContainer from '@material-ui/core/TableContainer';
 import TablePagination from '@material-ui/core/TablePagination';
 import PropTypes from 'prop-types';
-import React, { useEffect, useState, useRef } from 'react';
+import React, { useEffect, useState, useRef, useCallback, memo } from 'react';
```
haxxmaxx, line 7: useCallback not used anymore

src/table/TableWrapper.jsx
```diff
@@ -48,6 +48,7 @@ export default function TableWrapper(props) {
 
   const handleChangePage = (event, newPage) => {
     setPageInfo({ top: newPage * rowsPerPage, height: rowsPerPage });
+    setfocusedCellCoord([0, 0]);
```
haxxmaxx, line 51: why is this added?

src/table/TableWrapper.jsx
a-m-dev: ✅ Fixed

src/table/cells/handle-cell-focus.js
```diff
@@ -14,25 +14,29 @@ export const updateFocus = (rowElements, cellCoord, shouldFocus = true) => {
 export const handleClickToFocusBody = (cell, focusedCellCoord, rootElement) => {
   const { rawRowIdx, rawColIdx } = cell;
   const rowElements = rootElement.getElementsByClassName('sn-table-row');
-  updateFocus(rowElements, focusedCellCoord.current, false);
-  focusedCellCoord.current = [rawRowIdx + 1, rawColIdx];
-  updateFocus(rowElements, [rawRowIdx + 1, rawColIdx]);
+  updateFocus(rowElements, focusedCellCoord, false);
+  updateFocus(rowElements, [rawRowIdx + 1, rawColIdx], true);
```
haxxmaxx, line 18: Shouldn't `focusedCellCoord` be updated here, i.e. call `setFocusedCellCoord`

src/table/cells/handle-cell-focus.js
```diff
@@ -14,25 +14,29 @@ export const updateFocus = (rowElements, cellCoord, shouldFocus = true) => {
...
 };
 
 export const handleClickToFocusHead = (columnIndex, focusedCellCoord, rootElement) => {
   const rowElements = rootElement.getElementsByClassName('sn-table-row');
-  updateFocus(rowElements, focusedCellCoord.current, false);
-  focusedCellCoord.current = [0, columnIndex];
-  updateFocus(rowElements, [0, columnIndex]);
+  updateFocus(rowElements, focusedCellCoord, false);
+  updateFocus(rowElements, [0, columnIndex], true);
```
haxxmaxx, line 24: and here

src/table/TableBodyWrapper.jsx
a-m-dev: That was why i used `if ... else` before in order to clear it up a bit in. By default the logic behind comparing `prevProps` and `nextProps` always confusing, unless you make it clear that for example i want to return true in these cases and false in other ones. 

My suggestion here could be the same `if...else` because it's a bit clear and simple to understand in this case.

src/table/TableBodyWrapper.jsx
a-m-dev: ✅ Fixed

src/table/TableHeadWrapper.jsx
a-m-dev: ✅ Fixed

src/table/TableWrapper.jsx
a-m-dev: ✅ Fixed

src/table/cells/handle-cell-focus.js
```diff
@@ -41,3 +45,27 @@ export const handleResetFocus = (focusedCellCoord, rootElement, shouldRefocus, h
     rowElements[0]?.getElementsByClassName('sn-table-cell')[nextcell[1]].setAttribute('tabIndex', '0');
   }
 };
+
+export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement }) => {
+  const MIN_ROW_COUNT = 2;
+  if (!tableSection.current.scrollTo) return;
```
haxxmaxx, line 51: isn't `tableSection.current` undefined before ref is set? In that case add a `?`

src/table/TableWrapper.jsx
a-m-dev: I guess it's not required any more since we do it in `handleResetFocus`

a-m-dev: ✅ I'll remove it

src/table/cells/__tests__/handle-cell-focus.spec.js
```diff
@@ -73,38 +73,112 @@ describe('handle-key-press', () => {
...
+    let tableSection;
+    let rowHeight;
+
+    beforeEach(() => {
+      rowHeight = 100;
+      scrollTo = sinon.spy();
+      tableSection = { current: { scrollTo } };
+    });
+
+    it('should do any thing when ref is not setup yet', () => {
```
haxxmaxx, line 146: 'Should not do anything...'

src/table/cells/__tests__/handle-cell-focus.spec.js
```diff
@@ -73,38 +73,112 @@ describe('handle-key-press', () => {
...
+      tableSection.current = {};
+
+      handleCellFocus.handleNavigateTop({ tableSection, focusedCellCoord, rootElement });
+      expect(scrollTo).to.not.have.been.called;
+    });
+
+    it('should scroll to top of the element when it dedetects cursor is behind `<TableHead />`', () => {
```
haxxmaxx, line 153: detects

src/table/cells/handle-cell-focus.js
a-m-dev: no, it's always an empty object, but we could also check that by adding an optional chaining (?)

a-m-dev: ✅ Fixed

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: ✅ Fixed

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: ✅ Fixed

src/table/TableBodyWrapper.jsx
haxxmaxx: That is just bad syntax IMO. Not that hard to understand the condition itself, but why it should update when focusCellCoord stays the same, and not update when it changes. Kind of counter-intuitive. Usually you update when something else changes. That part needs an explanation I think

src/table/cells/__tests__/handle-cell-focus.spec.js
haxxmaxx: Oh actually, this one should say something like: `should scroll to the top when you reach the top two rows`

src/table/cells/__tests__/handle-cell-focus.spec.js
```diff
@@ -73,38 +73,112 @@ describe('handle-key-press', () => {
...
+      focusedCellCoord = [1, 0];
+
+      handleCellFocus.handleNavigateTop({ tableSection, focusedCellCoord, rootElement });
+      expect(scrollTo).to.have.been.calledOnce;
+    });
+
+    it('should scroll towords up automatically if it detects the cursor gets behind <TableHead />', () => {
```
haxxmaxx, line 160: scroll upwards

src/table/cells/__tests__/handle-cell-focus.spec.js
```diff
@@ -73,38 +73,112 @@ describe('handle-key-press', () => {
...
+            return { getElementsByClassName: () => [cell] };
+          }),
+      };
+      // targetOffsetTop = tableSection.current.scrollTop - cell.offsetHeight;
+      // 700 - 100 = 600 => so our scrollTo function migth be called with 600
+      let targetOffsetTop = 600;
+
+      handleCellFocus.handleNavigateTop({ tableSection, focusedCellCoord, rootElement });
+      expect(scrollTo).to.have.been.calledOnce;
+      expect(scrollTo).to.have.been.calledWith({ top: targetOffsetTop, behavior: 'smooth' });
```
haxxmaxx, line 181: calledOnceWith

src/table/cells/handle-cell-focus.js
```diff
@@ -41,3 +45,27 @@ export const handleResetFocus = (focusedCellCoord, rootElement, shouldRefocus, h
...
+    });
+  } else {
+    const [x, y] = focusedCellCoord;
+    const rowElements = rootElement.getElementsByClassName('sn-table-row');
+    const cell = rowElements[x]?.getElementsByClassName('sn-table-cell')[y];
+
+    if (cell.offsetTop - cell.offsetHeight * MIN_ROW_COUNT < tableSection.current.scrollTop) {
+      const targetOffsetTop = tableSection.current.scrollTop - cell.offsetHeight;
+      tableSection.current.scrollTo({
+        top: targetOffsetTop <= 0 ? 0 : targetOffsetTop,
```
haxxmaxx, line 66: This is the same as Math.max(0, targetOffsetTop), but both are fine

src/table/cells/handle-cell-focus.js
a-m-dev: ✅ Fixed

src/table/cells/handle-cell-focus.js
a-m-dev: ✅ Fixed

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: ✅ Fixed

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: ✅ Fixed

src/table/cells/handle-cell-focus.js
a-m-dev: also updated the test cases to check if `setFocusedCellCoord` has been called ( of course it does and passes the unit test ✅ )

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: ✅ Fixed

src/table/TableBodyWrapper.jsx
a-m-dev: ✅ Done, I added this: 
```
we need to check if `focusedCellCoord` updated, then no need to re-render this component
because it causes unnecessary extra re-render to this component when we move the focus to another cell and we want to avoid that
```

haxxmaxx (approved):

src/table/cells/handle-cell-focus.js
```diff
@@ -11,33 +11,76 @@ export const updateFocus = (rowElements, cellCoord, shouldFocus = true) => {
...
+        behavior: 'smooth',
+      });
+    }
+  }
+};
+
+export const handleRestoreFocusOnPageChange = ({ rootElement, nextCoords, setfocusedCellCoord }) => {
```
haxxmaxx, line 75: What's the difference between this and `handleResetFocus`?

a-m-dev: ✅ fixed

src/table/cells/handle-cell-focus.js
```diff
@@ -1,5 +1,5 @@
-export const updateFocus = (rowElements, cellCoord, shouldFocus = true) => {
-  const cell = rowElements[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[cellCoord[1]];
+export const updateFocus = ({ rowElements = {}, cellCoord = [], shouldFocus = true, providedCell = undefined }) => {
+  const cell = rowElements[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[cellCoord[1]] || providedCell;
```
haxxmaxx, line 2: providedCell not used anymore

src/table/cells/handle-cell-focus.js
```diff
@@ -11,33 +11,74 @@ export const updateFocus = (rowElements, cellCoord, shouldFocus = true) => {
   }
 };
 
-export const handleClickToFocusBody = (cell, focusedCellCoord, rootElement) => {
+export const findCellWithTabStop = (rootElement) => rootElement.querySelector("td[tabindex='0'], th[tabindex='0']");
+
+export const handleClickToFocusBody = (cell, rootElement, setfocusedCellCoord, constraints) => {
+  if (constraints.active) return;
```
haxxmaxx, line 17: No need to do this. If we consider being able to focus a cell in edit mode a bug, we can remove it, but in a different PR. But it might actually be useful (the old table does it)

src/table/TableHeadWrapper.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, memo } from 'react';
```
haxxmaxx, line 1: memo not used

src/table/cells/handle-cell-focus.js
a-m-dev: ✅ ok, that is reasonable, i'll remove it

src/table/TableHeadWrapper.jsx
a-m-dev: ✅ Fixed

src/table/TableBodyWrapper.jsx
```diff
@@ -35,16 +35,16 @@ const useStyles = makeStyles({
   },
 });
 
-const TableBodyWrapper = ({
```
LiKang6688, line 38: @a-m-dev hi, I am curious that any specific reason to change to use `function` ?

a-m-dev: Hey Li, 
Not that much, i guess, i might have change it because of `sinon.stub` thing that we had a bit of problem in our test files, but both are fine

PR #249, opened 2021-09-07
fix: put sprout back

haxxmaxx:
Sprout was incorrectly removed here #239 and it broke the styling.


niekvanstaveren:
You cannot fetch this dependency from the open source repo. Also the reason why we removed it

haxxmaxx:
@niekvanstaveren Yupp been talking to Tobias about it. Created a new [issue](https://github.com/qlik-oss/sn-table/issues/250), will start working on it now

PR #251, opened 2021-09-09
feat: include sprout as json

haxxmaxx:
Copied the json from what we used to import. The whole thing is 1500 lines, reduced it to 500. Removing all overrides etc concerning components we don't use. Could probably be stripped down further but not worth the effort.


Caele (approved):
Just rename it to not cause future confusion, else looks good.

PR #252, opened 2021-09-09
feat: add custom pagination and first and last buttons

### LINKED ISSUE #244: Pagination: add go to last/first page
Need to be added by customizing `<TablePagination>`

eliseeborn:
Fixes #244. 

things left:
- [ ] translations for aria labels?
- [ ] translations for titles?
- [x] unit test
- [ ] remove focus/hover styling on disabled buttons?


haxxmaxx:
We need to look into translations in general. Let's talk about it on monday

package.json
```diff
@@ -45,6 +45,7 @@
     "@babel/preset-react": "^7.13.13",
     "@babel/plugin-transform-react-jsx": "^7.14.3",
     "@material-ui/core": "^4.11.4",
+    "@material-ui/icons": "^4.11.2",
```
haxxmaxx, line 48: Don't think we want to include the entire thing, it is pretty huge. Can you put `@material-ui/core/IconButton` etc here instead? If it works

src/table/TableWrapper.jsx
```diff
@@ -29,6 +34,90 @@ const useStyles = makeStyles({
   },
 });
 
+const useStyles2 = makeStyles({
```
haxxmaxx, line 37: rename to `useStylesPaginationActions` or something

src/table/TableWrapper.jsx
```diff
@@ -29,6 +34,90 @@ const useStyles = makeStyles({
...
+    onChangePage(event, Math.max(0, Math.ceil(count / rowsPerPage) - 1));
+  };
+
+  const onFirstPage = page === 0;
+  const onLastPage = page >= Math.ceil(count / rowsPerPage) - 1;
+
+  return (
+    <div className={classes.root}>
+      <IconButton 
+        onClick={onFirstPage ? undefined : handleFirstPageButtonClick}
```
haxxmaxx, line 75:
```suggestion
        onClick={onFirstPage && handleFirstPageButtonClick}
```

src/table/TableWrapper.jsx
```diff
@@ -29,6 +34,90 @@ const useStyles = makeStyles({
...
+  const handleBackButtonClick = (event) => {
+    onChangePage(event, page - 1);
+  };
+
+  const handleNextButtonClick = (event) => {
+    onChangePage(event, page + 1);
+  };
+
+  const handleLastPageButtonClick = (event) => {
+    onChangePage(event, Math.max(0, Math.ceil(count / rowsPerPage) - 1));
```
haxxmaxx, line 66: How could it be < 0? if there is only one page? then it should be disabled anyway I think?

package.json
eliseeborn: That doesn't seem to be possible, it returns a 405.

haxxmaxx:
Try merging main. It might help you with the hover styling. Also, I think you can break it out to it's own component, TableWrapper is getting pretty big

package.json
haxxmaxx: Ah ok. We might actually just bundle the 4 that we use, then it is fine :) Let's discuss it on planning

haxxmaxx:
Is the styling looking fine? Then I think we can merge this and create a new task for translations

src/table/TablePaginationActions.jsx
```diff
@@ -0,0 +1,90 @@
...
+      <IconButton 
+        onClick={!onLastPage && handleLastPageButtonClick}
+        aria-disabled={onLastPage}
+        aria-label="last page"
+        title="Last page"
+        className={onLastPage && classes.disabled}
+      >
+        <LastPageIcon />
+      </IconButton>
+    </div>
```
a-m-dev, line 81: one thing that always comes handy in this kind of situations is using an array of (configuration) objects instead of putting multiple `<IconButton />` element in our jsx files. the basic idea is about preventing large module files. 

if we could have an aray of configs for each button like below: 
```
const tableBtnsArr = [
  {
    order: 0,
    IconProps: {
      onClick: !onLastPage && handleLastPageButtonClick,
      'aria-disabled' : onLastPage,
      'aria-label': 'last page',
      title: 'Last Page',
      className: onLastPage && classes.disabled,
    },
    IconToRender: <LastPageIcon />
  }, 
  { ...another button here },
  { ...another button here },
].sort((a, b) => a.order - b.order)
```

and then in your return clause you can do:
```
tableBtnsArr.map(btn => (
    <IconButton 
        key={btn.order}
        {...btn.IconProps}
      >
         {btn.IconToRender}
      </IconButton>
))
```

this will prevent from large component files all the time and fix the problem for code climate as well.

src/table/TableWrapper.jsx
```diff
@@ -29,6 +30,8 @@ const useStyles = makeStyles({
   },
 });
 
+
+
```
LiKang6688, line 34: please clear these two lines above

src/table/TablePaginationActions.jsx
haxxmaxx: Every object in the array will be similar and you might get the code climate problem anyway. We shouldn't worry too much about the similarity issues, but when it is identical you should most likely refactor.

eliseeborn: yeah, I think leaving the jsx as it is in this case will be clearer, but I'll keep it in mind :)

a-m-dev: rather than that, it's looking fine to me 👌

src/table/TablePaginationActions.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const handleBackButtonClick = (event) => {
+    onChangePage(event, page - 1);
+  };
+
+  const handleNextButtonClick = (event) => {
+    onChangePage(event, page + 1);
+  };
+
+  const handleLastPageButtonClick = (event) => {
+    onChangePage(event, Math.ceil(count / rowsPerPage) - 1);
```
haxxmaxx, line 37: Could you use `onLastPage` in here?

haxxmaxx (approved):

src/table/TablePaginationActions.jsx
```diff
@@ -0,0 +1,91 @@
...
+      >
+        <LastPageIcon />
+      </IconButton>
+    </div>
+  );
+}
+
+TablePaginationActions.propTypes = {
+  count: PropTypes.number.isRequired,
+  onPageChange: PropTypes.func.isRequired,
```
LiKang6688, line 88: should be `onChangePage`

LiKang6688:
Before you merge, could you use Prettier to format the code, please?

LiKang6688 (approved):

PR #256, opened 2021-09-14
feat: use nebulas keyboard hook to proberly focus in and out of table

### LINKED ISSUE #10: Handle focusing in and out of table using useAccessibility hook
Where do we get focus initially when tabbing to the table? Do we keep tabIndex on a cell?
This might need updates in nebula, since I don't think there is a way to pass a callback to be called on an enter event on a sheet object (maybe not relevant for phase 1)

### LINKED ISSUE #135: Handle clicking outside the table
When loosing focus from the table by clicking somewhere else, it should remove tabindex. This can be done with a listener on the document, looking if target has a parent inside the table. Possibly also send something to nebuala via the useAccessibility hook.

haxxmaxx:
fixes #10 and fixes #135

Using the new keyboard hook in nebula to catch when the table should recieve focus, by listening to `keyboard.active`. Also calling `keyboard.blur` when you press esc, click outside or tab out. Then calling `keyboard.focus` when clicking to fokus.

Needed to rewrite the focus handling quite a bit since we don't want to blur when we move focus, just reset the tabindex. This is because we have a focusout listener that only should catch when we actually focus out of the table, not when moving to a new cell basically.


src/table/TableWrapper.jsx
```diff
@@ -53,33 +54,54 @@ export default function TableWrapper(props) {
...
+    tableWrapperRef.current.addEventListener('focusout', focusOutCallback);
     return () => {
-      window.removeEventListener('resize', updateSize);
-      tableSection.current.removeEventListener('wheel', (evt) => handleScroll(evt, tableSection));
+      window.removeEventListener('resize', resizeCallback);
+      tableSectionRef.current.removeEventListener('wheel', scrollCallback);
+      tableWrapperRef.current.removeEventListener('focusout', focusOutCallback);
     };
-  }, []);
+  }, [keyboard]);
```
haxxmaxx, line 86: might be able to remove the dependency here

haxxmaxx, line 86: might be able to remove the dependency here

package.json
```diff
@@ -45,10 +45,10 @@
...
     "@babel/plugin-transform-react-jsx": "^7.14.9",
     "@material-ui/core": "^4.12.3",
-    "@nebula.js/cli": "1.7.0",
-    "@nebula.js/cli-build": "1.7.0",
-    "@nebula.js/cli-sense": "1.7.0",
-    "@nebula.js/cli-serve": "1.7.0",
+    "@nebula.js/cli": "2.0.0-beta.1",
+    "@nebula.js/cli-build": "2.0.0-beta.1",
+    "@nebula.js/cli-sense": "2.0.0-beta.1",
+    "@nebula.js/cli-serve": "2.0.0-beta.1",
```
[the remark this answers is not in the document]
haxxmaxx: My idea is that we can create a release with this, so that it can easily be tested, by including this version of nebula and a table release on a branch in the client. Then we can make another release with nebula 2.0

rollup.config.js
```diff
@@ -34,7 +34,7 @@ export default {
     name: moduleName,
     format: 'umd',
     exports: 'default',
-    sourcemap: false,
+    sourcemap: true,
```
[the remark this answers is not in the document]
haxxmaxx: yes will fix before I merge

package.json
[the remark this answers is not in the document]
niekvanstaveren: Okay, but that should probably be a beta release as well then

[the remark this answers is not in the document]
haxxmaxx: sounds good

src/table/cells/handle-key-press.js
```diff
@@ -3,7 +3,20 @@ import { updateFocus } from './handle-cell-focus';
 
 const isCtrlShift = (evt) => evt.shiftKey && (evt.ctrlKey || evt.metaKey);
 
-export const updatePage = ({ evt, totalRowSize, page, rowsPerPage, handleChangePage, setShouldRefocus }) => {
+export const preventDefaultBehavior = (evt) => {
+  evt.stopPropagation();
+  evt.preventDefault();
+};
+
+export const handleWrapperKeyDown = ({
```
LiKang6688, line 11: how about renaming it to `handleTableWrapperKeyDown`

src/table/TableWrapper.jsx
```diff
@@ -64,53 +65,76 @@ export default function TableWrapper(props) {
...
+
+    window.addEventListener('resize', resizeCallback);
+    tableSectionRef.current.addEventListener('wheel', scrollCallback);
+    tableWrapperRef.current.addEventListener('focusout', focusOutCallback);
     return () => {
-      window.removeEventListener('resize', updateSize);
-      tableSection.current.removeEventListener('wheel', (evt) => handleScroll(evt, tableSection));
+      window.removeEventListener('resize', resizeCallback);
+      tableSectionRef.current.removeEventListener('wheel', scrollCallback);
+      tableWrapperRef.current.removeEventListener('focusout', focusOutCallback);
```
niekvanstaveren, line 78: Not fully understand why we are doing this, but if we want to know about resize why do we not use the useRect from Nebula?

src/table/TableWrapper.jsx
```diff
@@ -64,53 +65,76 @@ export default function TableWrapper(props) {
...
     handleNavigateTop({
-      tableSection,
+      tableSectionRef,
       focusedCellCoord,
       rootElement,
     });
-  }, [tableSection, focusedCellCoord]);
+  }, [tableSectionRef, focusedCellCoord]);
+
+  useDidUpdateEffect(() => {
```
niekvanstaveren, line 90: What is the benefit of this useDidUpdateEffect?

src/table/cells/handle-cell-focus.js
```diff
@@ -1,67 +1,75 @@
-export const updateFocus = ({ rowElements = {}, cellCoord = [], shouldFocus = true, providedCell = undefined }) => {
+export const findCellWithTabStop = (rootElement) => rootElement.querySelector("td[tabindex='0'], th[tabindex='0']");
+
+export const updateFocus = ({ focusType, rowElements = [], cellCoord = [], providedCell = undefined }) => {
   const cell = providedCell || rowElements[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[cellCoord[1]];
-  if (cell) {
-    if (shouldFocus) {
+  if (!cell) return;
+
+  switch (focusType) {
```
niekvanstaveren, line 7: The move of this focus is to get the blue border I guess? Is tabindex needed as well for this?

src/table/cells/handle-key-press.js
```diff
@@ -144,6 +150,7 @@ export const bodyHandleKeyPress = (
     }
     // Esc: Cancels selections
     case 'Escape': {
+      if (!isAnalysisMode || !selState.rows.length) break;
```
LiKang6688, line 153: do we really need this line?

src/table/cells/handle-key-press.js
haxxmaxx: sure

src/table/TableWrapper.jsx
haxxmaxx: It doesn't run on first render. if we use `useEffect` it will blur even though nothing is focused

src/table/cells/handle-cell-focus.js
haxxmaxx: yes. We want to be able to tab to the pagination stuff in the footer, then tab back

src/table/cells/handle-key-press.js
haxxmaxx: yes. if you press escape when there are no selections, the listener on the tableWrapper should catch the escape event and blur.

haxxmaxx: I'll change the comment to reflect that

src/table/TableWrapper.jsx
haxxmaxx: Not part of this PR. I think Li tried that and it didn't work, ask him 🙃

src/table/cells/handle-key-press.js
haxxmaxx: done

src/table/cells/handle-key-press.js
haxxmaxx: done

src/table/TableWrapper.jsx
LiKang6688: @niekvanstaveren  do you mean the `window.addEventListener('resize', resizeCallback);`?  I gonna look into the useRect.

LiKang6688 (approved):
LGTM

PR #258, opened 2021-09-14
fix: variable name confusion

LiKang6688:
Screenshot 2021-08-24 at 15 10 13

`selState && selState.rows.length > 0;` to check if you have unconfirmed selections, so one or more cells are selected but not confirmed yet.

I feel it have the same meaning with selectionsAPI.isModal(), but selectionsAPI is a react hook and can't be used in regular JavaScript functions..

### DIFF
--- a/src/table/cells/handle-key-press.js
+++ b/src/table/cells/handle-key-press.js
@@ -18,17 +18,18 @@ export const updatePage = (evt, totalRowSize, page, rowsPerPage, handleChangePag
 
 export const arrowKeysNavigation = (evt, rowAndColumnCount, cellCoord, selState) => {
   let [nextRow, nextCol] = cellCoord;
-  const isSelectedTable = selState && selState.rows.length > 0;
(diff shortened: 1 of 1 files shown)


a-m-dev (approved):
looks fine 👌

src/table/cells/handle-key-press.js
```diff
@@ -18,17 +18,18 @@ export const updatePage = (evt, totalRowSize, page, rowsPerPage, handleChangePag
 
 export const arrowKeysNavigation = (evt, rowAndColumnCount, cellCoord, selState) => {
   let [nextRow, nextCol] = cellCoord;
-  const isSelectedTable = selState && selState.rows.length > 0;
+  // check if you have unconfirmed selections, so one or more cells are selected but not confirmed yet.
+  const isUnconfirmedSelection = selState && selState.rows.length > 0;
```
haxxmaxx, line 22: I think the correct terminology is `hasUnconfirmedSelection`

LiKang6688: Better! Thanks. updated.

PR #262, opened 2021-09-15
fix: select style

LiKang6688:
head:
Screenshot 2021-09-15 at 20 24 00

fixed:
Screenshot 2021-09-15 at 20 19 53


haxxmaxx:
I mean it is still good if we fix this, just that it should be done in `mui-setup` instead.

PR #265, opened 2021-09-16
chore: move build to nebula

LiKang6688:
Learning from sn-kpi to migrates the rollup build to use Nebula-CLI:

- Utilize Nebula build to create a package and a core package (umd and esm respectively)
- Add nebula.config.js to replace process.env.PACKAGE_VERSION
- Add prepack to validate nebula package


haxxmaxx:
Really nice update. Would it be hard to check the version number across `sn-table.qext`, `package.json` and `core/package.json`? It would simplify the release process :)

LiKang6688:
> Really nice update. Would it be hard to check the version number across `sn-table.qext`, `package.json` and `core/package.json`? It would simplify the release process :)

we need to checkout the version number across between `package.json` and `core/package.json`. Once we bump sn-table, we need update both in `package.json` and `core/package.json`. 
And the version number of `sn-table.qext` is from `package.json`.

tools/validate-nebula-package.js
```diff
@@ -0,0 +1,125 @@
...
+  // make sure no mention of qlik-trial or @qlik
+  if (/@qlik/.test(cleanedPkg)) {
+    throw new Error('package.json may not contain any mention of @qlik');
+  }
+  if (/qlik-trial/.test(cleanedPkg)) {
+    throw new Error('package.json may not contain any mention of qlik-trial');
+  }
+  if (/jfrog|qliktech/.test(cleanedPkg)) {
+    throw new Error('package.json may not contain any mention of qliktech or jfrog');
+  }
```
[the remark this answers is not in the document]
LiKang6688: Removed

tools/validate-nebula-package.js
```diff
@@ -0,0 +1,114 @@
...
+
+  // readme
+  if (!fs.existsSync(path.resolve(dir, 'README.md'))) {
+    throw new Error('Missing README.md file');
+  }
+
+  return cleanedPkg;
+};
+
+module.exports = validate;
```
niekvanstaveren, line 114: Is this something we need to do for all nebula charts? If so should this not be part of nebula itself?

tools/build.js
```diff
@@ -0,0 +1,37 @@
...
+if (watch) {
+  buildArgs.watch = true;
+}
+
+const main = async () => {
+  console.log('---> BUILDING SUPERNOVA');
+  build(buildArgs);
+};
+
+main();
```
niekvanstaveren, line 37: What is in here that cannot be handled using the standard nebula cli?

tools/validate-nebula-package.js
LiKang6688: No, I do not think like that. External developers do not do that.

niekvanstaveren: So we don't do this for all nebula charts that we create?

LiKang6688: I think that doing this for all nebula charts that we create is good. But nebula is also used by others who do not need to do this.

PR #268, opened 2021-09-17
[Fix#214]: Fix for table cell selection a11y issue

### LINKED ISSUE #214: It is not clear to users that elements can be selected and their selected status is not conveyed to screen reader users.

a-m-dev:
Hi, this is a PR for fixing the table cell selection status being conveyed by screen reader. 

In this PR a hidden label has been used for screen readers and a function called `getSrNotation` added to get the latest value for the label element. 

I put that function inside of component, but if you think we should extract it to another module, it's totally fine and doable.

fix #214


src/table/TableBodyWrapper.jsx
```diff
@@ -68,56 +72,96 @@ const TableBodyWrapper = ({
...
+
+    setTimeout(() => {
+      setSrNotation(getSrNotation());
+    }, MAX_NOTATION_DELAY);
+  }, [focusedCellCoord, selState]);
+
+  const getSrNotation = () => {
+    if (focusedCellCoord.toString() == '0,0') return '';
+
+    const [x, y] = focusedCellCoord;
```
haxxmaxx, line 86: fyi these are not x, y, if anything y, x. Or just rowIdx, colIdx for clarity. I think I put them in that order since that is the way the dom is ordered

a-m-dev: ✅ Fixed

src/table/TableBodyWrapper.jsx
```diff
@@ -68,56 +72,96 @@ const TableBodyWrapper = ({
...
+  useEffect(() => {
+    setSrNotation('');
+
+    setTimeout(() => {
+      setSrNotation(getSrNotation());
+    }, MAX_NOTATION_DELAY);
+  }, [focusedCellCoord, selState]);
+
+  const getSrNotation = () => {
+    if (focusedCellCoord.toString() == '0,0') return '';
```
haxxmaxx, line 84: I guess this for the entire first row?

rollup.config.js
```diff
@@ -34,7 +34,7 @@ export default {
     name: moduleName,
     format: 'umd',
     exports: 'default',
-    sourcemap: false,
+    sourcemap: true,
```
haxxmaxx, line 37: Don't forget to revert this

src/table/TableBodyWrapper.jsx
```diff
@@ -68,56 +72,96 @@ const TableBodyWrapper = ({
...
 
+  useEffect(() => {
+    setSrNotation('');
+
+    setTimeout(() => {
+      setSrNotation(getSrNotation());
+    }, MAX_NOTATION_DELAY);
+  }, [focusedCellCoord, selState]);
+
+  const getSrNotation = () => {
```
haxxmaxx, line 83: Move this to handle-focus file or something

src/table/TableBodyWrapper.jsx
a-m-dev: you mean we should check it for the entire first row? 
currently it is for the very first cell

rollup.config.js
a-m-dev: ✅ Sure

src/table/TableBodyWrapper.jsx
a-m-dev: ✅ Sure, Fixed
I'm now checking it with `startsWith('0')`

haxxmaxx: why not just `focusedCellCoord[0] === 0`

src/table/TableBodyWrapper.jsx
```diff
@@ -58,6 +60,8 @@ function TableBodyWrapper({
     colIdx: -1,
     isEnabled: selectionsEnabled,
   });
+  const [srNotation, setSrNotation] = useState('');
+  const MAX_NOTATION_DELAY = 250;
```
haxxmaxx, line 64: Did you try without the delay?

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,31 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
+    notation += 'Value is not selected';
+    return notation;
+  } else {
+    const isSingularSelection = selState.rows.length === 1;
+    const selectionNote = `There ${isSingularSelection ? 'is' : 'are'} ${selState.rows.length} selected value${
+      isSingularSelection ? '' : 's'
+    } on ${tColumnName} column currently.`;
+
+    if (isCellSelected) notation += `${tCellContent},value is selected. ${selectionNote}`;
+    else notation += `${tCellContent},value is not selected.`;
```
haxxmaxx, line 106:
```suggestion
    notation += isCellSelected ? `${tCellContent}, value is selected. ${selectionNote}` : `${tCellContent}, value is not selected.`;
```
\+ a line break somewhere

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,31 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
 };
+
+export const getCellSrNotation = ({ focusedCellCoord, rootElement, selState, tableData }) => {
+  if (focusedCellCoord.toString().startsWith('0')) return '';
+
+  const [rowIdx, colIdx] = focusedCellCoord;
+  const rowElements = rootElement.getElementsByClassName('sn-table-row');
+  const cell = rowElements[rowIdx]?.getElementsByClassName('sn-table-cell')[colIdx];
+
+  const tColumnName = tableData.columns[colIdx].label;
```
haxxmaxx, line 91: I don't remember, did we discuss this? Optimally the screen reader should say the column name from the markup alone. If it doesn't than maybe we can have it here.

src/table/TableBodyWrapper.jsx
a-m-dev: yes that could be done as well 
✅ Fixed

src/table/TableBodyWrapper.jsx
a-m-dev: Yes, and the result was not really what we were expected. 

### here was my idea: 
Since react batches the `setState` updates in lifecycle methods and (synthetic) event handlers, this will be the case in our current situation if we try to remove the `setTimeout`, and in result, we would get the name of the cell only from screen reader ( witch is the default behaviour of it ) since the label would **not** update in subsequent row names that has the same value ( like two or couple of rows below together with same value, for example imagine row [6,0] = Sweden, [7,0]= Sweden, [8,0] = Sweden, ... ) and this will happen and happen until we get to another cell that has a different value.

### here was the result: 
This was the exact behaviour that i get from my test result in windows machine without `setTimeout`. and since the label does not updated because of the batched state updates, which would result in same label element value, i did not get NVDA screen reader to pick up the changes.

So what is the escape case from this situation? there are two: 
1) set timeout and 
2) ajax calls 
which both of them are not handled by react, it's JS related behaviour and API. 

In our case i'm using the setTimeout to escape out this behaviour of state update batch of react. I think the change we could do here is to do a favour for react and put the functionality at the end of event queue by setting the time to `0` in our setTimeout.

src/table/cells/handle-cell-focus.js
a-m-dev: ✅ Yes, NVDA says the column name

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,31 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
+  const tCellContent = cell && cell.innerText;
+  const isCellSelected = cell && cell.classList.contains('selected');
+
+  let notation = '';
+  if (!selState.rows.length) {
+    notation += 'Value is not selected';
+    return notation;
```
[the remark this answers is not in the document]
a-m-dev: this part has been updated in newer push

src/table/cells/handle-cell-focus.js
a-m-dev: this part has been updated in latest push

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,49 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
+    const tCellContent = cell && cell.innerText;
+    const isCellSelected = cell && cell.classList.contains('selected');
+
+    let notation = '';
+
+    if (selState.rows.length) {
+      const isSingularSelection = selState.rows.length === 1;
+      const selectionNote = `There ${isSingularSelection ? 'is' : 'are'} ${selState.rows.length} selected value${
+        isSingularSelection ? '' : 's'
+      } currently.`;
```
haxxmaxx, line 103: We need to consider how to solve this with strings. There is a way sending in a variable to a string, and then it places it at a specified place in the string. In this case the # of values. Also, this way of patching a string together, depending if there is one or multiple values selected wont work with strings. We could have just two different strings tho like this:
`There is <1> selected value currently`
`There are <2> selected values currently `

haxxmaxx: Just need to make sure we can add arguments to strings

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,49 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
+
+      if (prevSelectedCount < selState.rows.length) {
+        // if we select cell
+        notation += `Value is selected. ${selectionNote}`;
+      } else if (prevSelectedCount > selState.rows.length) {
+        // if we deselect cell
+        notation += `Value is deselected. ${selectionNote}`;
+      } else if (isCellSelected) {
+        // if we are in selection mode and move to selected cell
+        notation += `${tCellContent},value is selected.`;
```
haxxmaxx, line 113: Wouldn't this be announced twice? since it will already read the cell content

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,49 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
+      } else {
+        // if we are in selection mode and move to unselected cell
+        notation += `${tCellContent},value is not selected.`;
+      }
+    } else if (selState.rows.length === 0 && prevSelectedCount > 0) {
+      // if we deselect last (selected) cell which means we close the selection mode
+      notation += 'Value deselected and exited selection mode.';
```
haxxmaxx, line 120: We actually might want something like this for pressing enter and escape. I'd say that is bonus or at least a different PR

src/table/cells/handle-cell-focus.js
a-m-dev: I will separate it into a function of `getCellSelectionStatusNote`.

speaking of ( function ) template literals, i think we could handle it with just a simple function. and remove the redundant code from return function of `getMemoisedSrNotation` as well

a-m-dev: ✅ Fixed

src/table/cells/handle-cell-focus.js
a-m-dev: for some reason only **when we are in selection** state, chrome vox will not pick up the cell content when we changing the focus, but the NVDA is fine with it.
so the reason for adding cell content here was that and by adding it, chrome vox works fine and NVDA will convey it twice.

✅ So Since we are targeting NVDA as the main screen reader, i just removed it in order to prevent the double announcement issue.

src/table/cells/handle-cell-focus.js
haxxmaxx: Sure, would be nice to decrease the the complexity of this function. But all it will be later is something like this:
`selectionNote = translator(isSingularSelection ? 'table.singleSelection' : 'table.multipleSelection', selState.rows.length);`

haxxmaxx: To clarify, we can't do this before Li has merged the translation PR. The way it is now is fine :)

src/table/cells/handle-cell-focus.js
a-m-dev: yes, should we create a separate issue for it?

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,53 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
     }
   }
 };
+
+const getCellSelectionStatusNote = (rows) => {
+  if (!Array.isArray(rows) || !rows.length) return '';
```
haxxmaxx, line 85: This will never happen since the call is inside `if (selstate.rows.length)` already

src/table/cells/handle-cell-focus.js
```diff
@@ -80,3 +80,53 @@ export const handleNavigateTop = ({ tableSection, focusedCellCoord, rootElement
...
+        // if we deselect cell
+        notation += `value is deselected. ${selectionNote}`;
+      } else if (isCellSelected) {
+        // if we are in selection mode and move to selected cell
+        notation += 'value is selected.';
+      } else {
+        // if we are in selection mode and move to unselected cell
+        notation += 'value is not selected.';
+      }
+    } else if (selState.rows.length === 0 && prevSelectedCount > 0) {
```
haxxmaxx, line 122: length can't be negative so no need to check

src/table/cells/handle-cell-focus.js
a-m-dev: ✅ Fixed

src/table/cells/__tests__/handle-cell-focus.spec.js
```diff
@@ -248,4 +248,108 @@ describe('handle-key-press', () => {
...
+
+    it('should return `value selected` and selected rows count while we are selecting multiple rows', () => {
+      const focusedCellCoord = [2, 1];
+      selState = { rows: ['row#01', 'row#02'] };
+      getCellSrNotation = handleCellFocus.getMemoisedSrNotation(1);
+
+      const notation = getCellSrNotation({ focusedCellCoord, rootElement, selState });
+
+      expect(notation).to.equal('value is selected. There are 2 selected values currently');
+    });
```
haxxmaxx, line 308: You already have tests for `getCellSelectionStatusNote` so you only need one of these two

src/table/TableBodyWrapper.jsx
niekvanstaveren: Isn't this because the information is put into a separate label? Should the information not be present on every single cell instead?

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: Yes, you are right, 
Just put it here because of keeping our coverage rate up a bit . if we remove either case, then we might miss related code block in coverage report.

But if you think it's still unnecessary, i can remove one.

haxxmaxx: I don't think it improves coverage. Both of them should run line 109 in handle-cell-focus

src/table/TableBodyWrapper.jsx
```diff
@@ -26,12 +26,13 @@ const useStyles = makeStyles({
     },
   },
   srOnly: {
-    position: 'absolute',
-    left: '-10000px',
-    top: 'auto',
-    width: '1px',
+    clip: 'rect(0 0 0 0)',
+    clipPath: 'inset(50%)',
```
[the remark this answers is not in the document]
a-m-dev: Yes, I think this is the recommended way for styling the screen reader texts since we have a good support for `clip` and semi-good support for `clip-path` properties.

the previous version uses the `left: -SOME_LARGE_INT` that was not quite good in my opinion.

src/table/cells/__tests__/handle-cell-focus.spec.js
a-m-dev: ok, i'll remove the singular test then

a-m-dev: ✅ Fixed

src/table/TableBodyWrapper.jsx
a-m-dev: this way was our first implementation and since if we use `aria-live="assertive"`, that looked a bit unnecessary to put any other information to cell except the cell content; because we update the label and screen reader picks it up.  

even by shoving more info into cell content, we still going to need setTimeout here, because we need to trigger the screen reader to pick up the changes from label.

LiKang6688:
Do you have a link to windows PTE then I can have a manual test? @a-m-dev

a-m-dev:
Of course, Here it is: rd-htm-pte-3.rdlund.qliktech.co

src/table/cells/handle-cell-focus.js
```diff
@@ -89,6 +89,57 @@ export const handleNavigateTop = ({ tableSectionRef, focusedCellCoord, rootEleme
   }
 };
 
+export const getCellSelectionStatusNote = (rows, translator) =>
+  rows.length === 1
+    ? translator.get('SNTable.SelectionLabel.OneSelectedValue')
+    : translator.get('SNTable.SelectionLabel.SelectedValues', [rows.length]);
+
+export const getMemoisedSrNotation = (prevCount = 0) => {
```
niekvanstaveren, line 97: How is this Memoised?

src/table/TableBodyWrapper.jsx
```diff
@@ -26,25 +26,29 @@ const useStyles = makeStyles({
...
+    whiteSpace: 'nowrap',
+    width: '1px',
   },
 });
 
 function TableBodyWrapper({
   rootElement,
   tableData,
   constraints,
+  translator,
```
niekvanstaveren, line 43: This is not added in the props validation

src/table/TableBodyWrapper.jsx
```diff
@@ -26,25 +26,29 @@ const useStyles = makeStyles({
...
   selectionsAPI,
   layout,
   theme,
   setShouldRefocus,
   setFocusedCellCoord,
+  focusedCellCoord,
```
niekvanstaveren, line 49: Also not added in the props validation

src/table/cells/handle-cell-focus.js
```diff
@@ -89,6 +89,57 @@ export const handleNavigateTop = ({ tableSectionRef, focusedCellCoord, rootEleme
...
+      // if we deselect last (selected) cell which means we close the selection mode
+      notation += translator.get('SNTable.SelectionLabel.ExitedSelectionMode');
+    }
+
+    prevSelectedCount = selState.rows.length;
+    return notation;
+  };
+};
+
+export const getCellSrNotation = getMemoisedSrNotation();
```
niekvanstaveren, line 142: This logic gets confusing in this way. Would a hook not be a better solution? Then you can keep the previous state as well. Probably good to add a refactor task for this as well

src/table/cells/handle-cell-focus.js
a-m-dev: I'm passing the `prevCount` value here only for testing purpose, in test cases we need to increase this value in order to be test different cases. in the code side we never increase it by passing it to the function.

src/table/TableBodyWrapper.jsx
a-m-dev: ✅. Fixed

src/table/cells/handle-cell-focus.js
a-m-dev: maybe, i'm not sure if we could bring functionality with hook, but of course i'll create a refactor issue for that. 👌

a-m-dev: added this one on the github board: https://github.com/qlik-oss/sn-table/issues/332

niekvanstaveren (approved):

PR #270, opened 2021-09-19
chore: Migration MUI from v4 to v5

LiKang6688:
Migrations:

- using @mui/styles 
The legacy styling solution of MUI for a temporary solution because of sprout.
https://mui.com/styles/basics/

- CSS injection order
migrate  from StylesProvider to `<StyledEngineProvider injectFirst>`
https://mui.com/guides/migration-v4/#style-library

- Theme structure
use `adaptV4Theme`
https://mui.com/guides/migration-v4/#theme-structure

- Table
onChangePage to onPageChange
onChangeRowsPerPage to onRowsPerPageChange

- Select style
MuiSelect.root does not work
use MuiNativeSelect.select

Update:
make `tableSection.current` into optional chaining
   
To Do:
regression testing

Screenshot 2021-10-18 at 15 04 15


src/table/TableWrapper.jsx
```diff
@@ -66,7 +66,7 @@ export default function TableWrapper(props) {
   useEffect(() => {
     const updateSize = () => setTableWidth(rootElement.clientWidth);
     window.addEventListener('resize', updateSize);
-    tableSection.current.addEventListener('wheel', (evt) => handleScroll(evt, tableSection));
+    tableSection.current?.addEventListener('wheel', (evt) => handleScroll(evt, tableSection));
```
a-m-dev, line 69: this one sometimes is unreliable, if it was based on typescript, it would still give you some errors here. 
i think it's better to specifically check for this one in a `if` statement in the beginning of the `useEffect`. like 
```
if(!tableSection.current) return;
```

LiKang6688: sure. Changed to 
`tableSection.current && tableSection.current.addEventListener('wheel', (evt) => handleScroll(evt, tableSection));`

Worth to mention that it is only a workaround for the unit testing issue. in unit testing, tableSection.current is regarded as null.

a-m-dev: Great, it's looking good to me ✅

src/table/TableWrapper.jsx
```diff
@@ -66,7 +66,7 @@ export default function TableWrapper(props) {
   useEffect(() => {
     const updateSize = () => setTableWidth(rootElement.clientWidth);
     window.addEventListener('resize', updateSize);
-    tableSection.current.addEventListener('wheel', (evt) => handleScroll(evt, tableSection));
+    tableSection.current && tableSection.current.addEventListener('wheel', (evt) => handleScroll(evt, tableSection));
```
haxxmaxx, line 69: What happens now in the return if we try to remove an event listener that does not exist?

LiKang6688: good question. It does not complain about anything. I tried to find a hint from documentation but find none.

LiKang6688: i tested that by directly replacing
`tableSection.current && tableSection.current.addEventListener('wheel', (evt) => handleScroll(evt, tableSection)); `
with
`tableSection.current.removeEventListener('wheel', (evt) => handleScroll(evt, tableSection));`

haxxmaxx: ah ok we are fine then. I guess it doesn't do anything if it does fine the one it wants to delete

haxxmaxx (approved):

src/index.js
```diff
@@ -38,6 +40,7 @@ export default function supernova(env) {
       const selectionsAPI = useSelections();
       const theme = useTheme();
       const keyboard = useKeyboard();
+      const rect = useRect();
```
LiKang6688, line 43: @niekvanstaveren @haxxmaxx  switched to useRect in the branch.

haxxmaxx: Put it in a seperate PR, has nothing to do with mui migration

LiKang6688: right, https://github.com/qlik-oss/sn-table/pull/293

src/mui-setup.js
```diff
@@ -1,4 +1,5 @@
-import { createGenerateClassName, createMuiTheme } from '@material-ui/core/styles';
+import { createTheme, adaptV4Theme } from '@mui/material/styles';
```
haxxmaxx, line 1: So are both of these depricated? If so for how long can we use them?

src/index.js
```diff
@@ -42,7 +40,6 @@ export default function supernova(env) {
       const rect = useRect();
 
       const [pageInfo, setPageInfo] = useState(() => ({ top: 0, height: 100 }));
-      const [muiParameters] = useState(() => muiSetup());
```
haxxmaxx, line 45: We had it here so we don't have to run muiSetup everytime the render `uesEffect` is triggered. Is it not working when it was here or can we keep it here as a state?

haxxmaxx:
Attach #282 properly or break it out into a seperate PR :)

LiKang6688:
> Attach #282 properly or break it out into a seperate PR :)

Okay, I will break it out into a separate PR.

src/index.js
haxxmaxx: ok sure

PR #278, opened 2021-09-28
Fix#214 draft listbox

a-m-dev:
in this draft pr, i'm checking for the role of listbox possibility


src/table/TableBodyWrapper.jsx
```diff
@@ -81,11 +83,10 @@ function TableBodyWrapper({
             const cell = row[column.id];
             const value = cell.qText;
             const CellRenderer = columnRenderers[columnIndex];
+
             return (
               CellRenderer && (
                 <CellRenderer
-                  scope={columnIndex === 0 ? 'row' : null}
-                  component={columnIndex === 0 ? 'th' : null}
```
haxxmaxx, line 88: Why is this removed?

haxxmaxx, line 88: Why is this removed?

src/table/TableBodyWrapper.jsx
```diff
@@ -108,10 +109,7 @@ function TableBodyWrapper({
                     )
                   }
                   onMouseDown={() => handleClickToFocusBody(cell, rootElement, setfocusedCellCoord)}
-                >
-                  <div className={classes.srOnly}>{column.label}</div>
-                  {value}
```
haxxmaxx, line 113: I completely forgot about this, if we can get rid of it that is great

src/table/cells/withStyling.jsx
```diff
@@ -1,10 +1,37 @@
...
+});
 
 export default function withStyling(CellComponent) {
   const HOC = (props) => {
-    const { styling } = props;
-    return <CellComponent {...props} className={`${styling.selectedCellClass} sn-table-cell`} style={styling} />;
+    const classes = useStyles();
+    const { styling, column, selState, value } = props;
+
+    const isInSelectionMode = selState.rows.length && Number(column.id.replace(/^\D+/g, '')) === selState.colIdx;
```
haxxmaxx, line 22: What is up with the regex? The column index is on the `cell` which is on the props, easier to use that

src/table/cells/withStyling.jsx
```diff
@@ -1,10 +1,37 @@
...
+    return (
+      <CellComponent
+        {...props}
+        role={isInSelectionMode ? 'option' : null}
+        aria-selected={isInSelectionMode ? !!styling.selectedCellClass : null}
+        className={`${styling.selectedCellClass} sn-table-cell`}
+        style={styling}
+      >
+        {value}
+        {!!styling.selectedCellClass && <span className={classes.srOnly}>selected</span>}
```
haxxmaxx, line 32: Is this still needed?

src/table/TableBodyWrapper.jsx
a-m-dev: by mistake, 
✅ Fixed

a-m-dev: ✅ Fixed

src/table/cells/withStyling.jsx
a-m-dev: Great suggestion, i was wondering how i can find the columnIdx in the hoc. 
✅ Fixed

src/table/cells/withStyling.jsx
a-m-dev: yes, when we are in selection mode and if we move the focus to a selected value, it will not convey that it cell is selected. so i check it for only selected values to make sure that sr conveys the status for it.

haxxmaxx: so it says nothing for unselected values?

haxxmaxx: so it says nothing for unselected values?

haxxmaxx: so it says nothing for unselected values?

a-m-dev: no, it picks up unselected value, but not selected ones
it only pick up selected ones when we select them

haxxmaxx:
We wont use this approach

Issue #288, opened 2021-09-30
Bottom toolbar in table should be responsive 


mlb76:
Bottom toolbar in table should be responsive , so when the table is at it's smallest only number of rows and pages and previous and next button is displayed.

PR #293, opened 2021-09-30
fix: bottom toolbar in table should be responsive

LiKang6688:
useRect replaces  window.addEventListener('resize', resizeCallback);


haxxmaxx:
This doesn't solve #288. We need it to be more responsive for even smaller tables, removing the first/last buttons, then the string for what strings are showing. Also, this is not a problem for view data release since we mount the table every time you toggle view data -> you can't resize the table without resizing the window. So this has to wait for now

haxxmaxx (approved):

Issue #260, opened 2021-09-15
Selection is not confirmed after tabbing out from the table


mlb76:
Selection is not confirmed after tabbing out from the table,

1. Tab into the table
2. Make a selection in the table
3. Tab outside of the table

Expected result: Selection is confirmed
Actual result: Table is still in selection mode and selection is not confirmed

haxxmaxx:
I think the trick to this one is always to confirm when the object looses focus. Thus this depends on the new a11y hook in nebula

a-m-dev:
Also, one more thing to consider is that when we are focused on pagination, the ESC key would not cancel the selection.

PR #304, opened 2021-10-05
while focus is on pagination and pressing esc key should do nothing

a-m-dev:
when we are focusing on pagination and have some selection, pressing esc key should not bring the focus back to the table wrapper. 

this PR fixes that. 

#260


LiKang6688 (approved):
LGTM. Please update the commit title.

haxxmaxx (approved):

PR #313, opened 2021-10-06
fix: disabled button should be able to be focusable for accessility purpose

LiKang6688:


src/table/TablePaginationActions.jsx
```diff
@@ -39,38 +43,42 @@ export default function TablePaginationActions(props) {
   return (
     <div className={classes.root}>
       <IconButton
-        onClick={handleFirstPageButtonClick}
-        disabled={onFirstPage}
+        onClick={!onFirstPage ? handleFirstPageButtonClick : () => {}}
```
haxxmaxx, line 46: Nice, this should solve the error message

LiKang6688: Yes

haxxmaxx (approved):
LGTM. Create an issue and attach it so we can track it in the board.

PR #319, opened 2021-10-07
feat: focus in and out of selection toolbar

### LINKED ISSUE #284: Selection toolbar not working properly inside sense
In the old table the the selection toolbar is attached the active column when selecting. It is also possible to navigate to it using the keyboard (shift + tab when you have focus on a table cell).

This is a special case for the table, so there is logic in there we need to mimic.
Having the toolbar placed correctly would be a bonus, but from a a11y perspective the toolbar becomes inaccessible if it can't be reached using the keyboard.

Note that this will mess with the focus out logic. Right now we are looking at if the focus exits the table element and telling nebula that we no longer have focus there. The selection toolbar is located somewhere on the object element in the DOM. So this might need changes in the keyboard hook

haxxmaxx:
fixes #284

- Adding special cases for tab and tab + shift so that we end up on the confirm button on the selection toolbar.
- When there is a confirm event and we don't have focus inside the table at all, run `keyboard.blur(true)`. This means that we get focus on the entire chart. A side effect of this is that **any confirm event will focus the entire chart**, including clicking on the confirm button and clicking outside. Are we ok with this side effect??
- cancel gives focus back on a table cell. We could add the same behaviour here if we want (focusing the entire chart)

Kapture 2021-10-08 at 10 16 18

### DIFF
--- a/src/table/__tests__/selections-util.spec.js
+++ b/src/table/__tests__/selections-util.spec.js
@@ -6,18 +6,33 @@ describe('selections-utils', () => {
     let api;
     let selDispatch;
     let setShouldRefocus;
+    let keyboard;
+    let containsActiveElement;
+    let tableWrapperRef;
 
     beforeEach(() => {
-      selDispatch = () => {};
-      setShouldRefocus = () => {};
+      selDispatch = sinon.spy();
+      setShouldRefocus = sinon.spy();
       api = {
         on: sinon.spy(),
         removeListener: sinon.spy(),
       };
+      keyboard = { active: false, blur: sinon.spy() };
+      containsActiveElement = true;
+      tableWrapperRef = {
+        current: {
+          contains: () => containsActiveElement,
+        },
+      };
+    });
+
+    afterEach(() => {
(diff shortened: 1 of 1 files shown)


src/table/cells/handle-cell-focus.js
```diff
@@ -94,3 +94,7 @@ export const handleFocusoutEvent = (evt, shouldRefocus, blur) => {
     blur(false);
   }
 };
+
+// get the object, find the patent of the confirmbutton in slection toolbar parent and focus that element
+export const focusConfirmButton = (element) =>
+  element.closest('.qv-object-wrapper')?.querySelector('.sel-toolbar-confirm')?.parentElement?.focus();
```
Caele, line 100: only valid if client?
So nothing outside right now, right?

haxxmaxx: the normal tab events will be canceled everywhere. We can add a `keyboard.enabled` check for now, or just don't care that it breaks outside the client for now

Caele (approved):

PR #322, opened 2021-10-08
fix for double esc key press to cancel

### LINKED ISSUE #321: Two clicks are needed to deselect a value and after deselecting a value arrows key navigation in the table does not work
Precondition:  Use NVDA screenreader and Chrome

1.  Go to a visualisation and choose to view data
2.  Enter a table a column and press space to select the value
3.  Press esc to deselect value -> nothing happens
4.  Press esc again -> value is deselected
5.  Try to move to another row or column using arrow keys

Actual result: Nothing happens. It is not possible to move between columns and rows using arrow keys

a-m-dev:
fixes #321 
NVDA turns to focus mode and the problem of pressing Esc key two times to cancel selection and more worse than than the problem of navigation on table cell rises. 

it turns out that NVDA has two basic modes at general: 
1. focus mode
2. browse mode
and in our cases it turns focus mode on by a typewriting machine sound ( like a glitch ). 

in order to prevent foucs mode on NVDA at all, we need to use `role="application"`. this will prevent this issue. 

also i added a simple test case to ensure that we always put `role="application"` on our `<TableContainer />`. 

**SPOILER ALERT:** this could be the possible solution for the same `role="button"` issue that we had previously while the sorting of the table. and that was exactly the same and turns focus mode on in the NVDA. but i'll check that in  another PR and will see if that was the case there as well.

### DIFF
--- a/src/table/TableWrapper.jsx
+++ b/src/table/TableWrapper.jsx
@@ -132,6 +132,7 @@ export default function TableWrapper(props) {
         ref={tableSectionRef}
         className={classes[containerMode]}
         tabIndex="-1"
+        role="application"
         data-testid="table-wrapper"
       >
         <Table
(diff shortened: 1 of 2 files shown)


LiKang6688:
@a-m-dev could you send me the url to windows PTE. so I can manually test it.

a-m-dev:
Sure, Here it is: rd-htm-pte-2.rdlund.qliktech.com

haxxmaxx (approved):

niekvanstaveren (commented):
While this works now, how do we think this will impact future work? Next step for the table is to enable search functionality. In that case the screen reader probably needs to switch into forms mode to enable the user to type in a search string. From what I read this will disable that completely

PR #341, opened 2021-10-12
[Fix#314]: Fixes sort announcements on table header cells

### LINKED ISSUE #314: After pressing sort the change sort order on a column NVDA does not let the user know that sorting is changed
Preconditions: Use NVDA screenreader + Chrome

1.  Move to a visualization using key board shortcuts¨
2. Press enter on the visualization -> data is displayed in the new table
3. Press enter to move into the table
4. Use arrow keys to move any column header
5. Press space to change the sorting order of that column

Actual result:  Order in column is changed but NVDA only reads space 
Expected result:  Order in column is changed and NVDA reads the column name again and how it is sorted.

a-m-dev:
fixes #314 

this fix will make screen readers like NVDA to announce the sort direction when pressing column header cells

checklist:
- [x] Resolve conflicts
- [x] Unit tests
- [x] Manual verification

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
       - image: circleci/node:14.15.0
       - image: qlikcore/engine:12.1126.0
         command: -S AcceptEULA=yes
-      - image: browserless/chrome:1.48.0-puppeteer-1.20.0
+      - image: browserless/chrome:1.49.1-puppeteer-1.20.0
     steps:
       - attach_workspace:
           at: ~/table
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install commitlint --edit $1
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn lint-staged
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,4 @@
 .cache/
 dist/
-
+.DS_Store
 *.html
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+};
--- a/core/package.json
+++ b/core/package.json
@@ -1,7 +1,7 @@
 {
   "module": "esm/index.js",
   "peerDependencies": {
-    "@nebula.js/stardust": "^2.0",
+    "@nebula.js/stardust": "^2.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   }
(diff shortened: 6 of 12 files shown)


src/table/TableHeadWrapper.jsx
```diff
@@ -43,17 +43,17 @@ function TableHeadWrapper({
...
   const headStyle = useMemo(() => getHeadStyle(layout, theme), [layout, theme.name()]);
   const classes = useStyles(headStyle);
-  const SORT_NOTATIONS = useMemo(() => ({
-    asc: translator.get('SNTable.SortLabel.SortedAscending'),
-    desc: translator.get('SNTable.SortLabel.SortedDescending'),
-  }));
+  // const SORT_NOTATIONS = useMemo(() => ({
+  //   asc: translator.get('SNTable.SortLabel.SortedAscending'),
+  //   desc: translator.get('SNTable.SortLabel.SortedDescending'),
+  // }));
```
a-m-dev, line 49: this part is language specific thing, and I'm thinking about if we rely on aria-sort  to handle table sort and considering it announces sort direction ( sorted ascending/descending ) which is actually i’m having now in my local branch, we will not going to need translations for SNTable.SortLabel.SortedAscending and  SNTable.SortLabel.SortedDescending  respectively. that is why i'm commenting them here in my draft PR.
in that case, will browser/users OS announce the sort directions in the users selected choice of language?

src/table/TableHeadWrapper.jsx
```diff
@@ -85,7 +87,6 @@ function TableHeadWrapper({
                 {column.label}
                 {isFocusInHead && (
```
a-m-dev, line 88: we also could have the check for isCurrentColumnActive in order to announce the "Press Space ..." thing only on actively sorted columns? what do you think? is it a good idea or not?

src/table/TableHeadWrapper.jsx
```diff
@@ -64,6 +64,8 @@ function TableHeadWrapper({
               align={column.align}
               className={`${classes.head} sn-table-head-cell sn-table-cell`}
               tabIndex={tabIndex}
+              aria-sort={isCurrentColumnActive ? `${column.sortDirection}ending` : null}
+              aria-pressed={isCurrentColumnActive}
```
haxxmaxx, line 68: I thought you needed role="button" for this?

src/table/TableWrapper.jsx
```diff
@@ -135,6 +135,7 @@ export default function TableWrapper(props) {
         ref={tableSectionRef}
         className={classes[containerMode]}
         tabIndex="-1"
+        role="application"
```
haxxmaxx, line 138: I thought thiswas already merged?

a-m-dev:
Great, Here is the link for windows PTE machine: `https://rd-htm-pte-3.rdlund.qliktech.com/sense/app/c271b37a-b526-4d2a-b694-cea4b5d0e8fc`

please just make sure you change the browser ( for sense-client and sn-table to pick up the language translations ) and nvda ( to announce in the target language ).

LiKang6688:
Screenshot 2021-10-15 at 10 15 45
I see that nvda says 'space', is it needed?

PR #345, opened 2021-10-13
fix: make sure focus works as intended with keyboard.enabled off

### LINKED ISSUE #339: Keyboard.enabled not true behaviour
When keyboard.enabled is falsey, we need to make sure we don't interfere with the focus handlig too much.
- There should always be a tabindex=0 on one and only one of the table cells
- all pagination stuff should be tabable
- selections should not be canceled when you tab in/out
- keyboard.focus/blur should never be called
- Nothing should happen if keyboard.active changes

This basically means we need a sanity check everywhere where we call blur/focus and keep tabindexes as static as possible

haxxmaxx:
fixes #339


src/table/utils/__tests__/handle-accessiblity.spec.js
```diff
@@ -1,7 +1,11 @@
 import { expect } from 'chai';
 import * as handleAccessibility from '../handle-accessibility';
 
+<<<<<<< HEAD:src/table/cells/__tests__/handle-cell-focus.spec.js
+describe('handle-cell-focus', () => {
+=======
 describe('handle-accessibility', () => {
+>>>>>>> main:src/table/utils/__tests__/handle-accessiblity.spec.js
```
LiKang6688, line 8: glad to see that lint catch the issue :)

haxxmaxx: hehe, well I should've caught it to begin with 🙃

PR #346, opened 2021-10-14
feat: dropdown pagination

### LINKED ISSUE #245: pagination: Add more specific page navigation
Currently you can only go to next/prev (soon first/last) page. We need to be able to go to a specific page, or at least jump x number of pages at a time.

There is no support for this in `<TablePagination>`, but there is a different component called pagination https://material-ui.com/components/pagination/. Proably not straight forward to combine the two, since we need features from both. My best bet is to implement a go to row/page dropdown for the tablePagination in Material UI.

eliseeborn:
Screenshot 2021-10-14 at 12 41 22
Adds dropdown for selecting page in the pagination actions. Fixes #245 

Todo:
- [x] add unit test


src/table/TablePaginationActions.jsx
```diff
@@ -46,25 +75,56 @@ export default function TablePaginationActions(props) {
...
-      <IconButton
-        onClick={!onFirstPage ? handleBackButtonClick : () => {}}
+          {options.map(value => {
+            return <option value={value.value}>{value.label}</option>
+          })}
+        </Select>
+      </FormControl>
+      {tableWidth > 350 && (
+      <IconButton 
+        onClick={!onFirstPage && handleFirstPageButtonClick}
```
LiKang6688, line 117: Screenshot 2021-10-14 at 12 50 10
The reason why I modified it to` !onFirstPage ? handleBackButtonClick : () => {} `is to prevent the issue above when you click the button if the button is aria-disabled.

src/table/TablePaginationActions.jsx
```diff
@@ -46,25 +75,56 @@ export default function TablePaginationActions(props) {
...
-          className={onFirstPage && classes.disabled}
+      <FormControl className={`${classes.formControl} ${thinTable && classes.hidden}`}>
+        <InputLabel className={classes.caption} htmlFor="pagination-dropdown">{translator.get('SNTable.Pagination.SelectPage')}: </InputLabel>
+        <Select
+          native
+          className={classes.dropdown}
+          id="pagination-dropdown"
+          value={page}
+          onChange={handleSelectPage}
+          label="Page"
```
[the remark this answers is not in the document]
eliseeborn: Probably - it is not written out anywhere, but it might be read by the screenreader

src/table/components/TablePaginationActions.jsx
```diff
@@ -46,24 +75,55 @@ export default function TablePaginationActions(props) {
...
+
   const onFirstPage = page === 0;
   const onLastPage = page >= Math.ceil(count / rowsPerPage) - 1;
 
+  const options = [];
+  for(let i = 0; i < Math.ceil(count / rowsPerPage); i++){
+    options.push({value: i, label: i+1})
+  }
+
+  const thinTable = tableWidth < 650;
```
haxxmaxx, line 90: Hmm we should discuss the order that the different components are removed. To me it makes more sense to have dropdown + next/previous than to have all buttons

eliseeborn: Yeah, and we should probably decide on one way to remove the components as well - I think how Li removed first/last might be neater.

haxxmaxx: agree

src/table/components/TablePaginationActions.jsx
```diff
@@ -46,11 +75,45 @@ export default function TablePaginationActions(props) {
...
+    options.push({ value: i, label: i + 1 });
+  }
+
+  const thinTable = tableWidth < 650;
+
   return (
     <div className={classes.root}>
+      <FormControl className={`${classes.formControl} ${thinTable && classes.hidden}`}>
+        <InputLabel className={classes.caption} htmlFor="pagination-dropdown">
+          {translator.get('SNTable.Pagination.SelectPage')}:{' '}
```
haxxmaxx, line 96:
```suggestion
          {`${translator.get('SNTable.Pagination.SelectPage')}: `}
```
Does this work? Think syntax is a bit more standard

src/table/components/TablePaginationActions.jsx
```diff
@@ -10,12 +13,38 @@ import { focusConfirmButton } from '../utils/handle-accessibility';
...
+    display: 'none',
+  },
+  dropdown: {
+    cursor: 'pointer',
+    minWidth: '16px',
+    maxHeight: '32px',
+    transform: 'translate(0px, 6px)',
+  },
+  input: {
+    padding: '6px 32px 7px 8px',
```
haxxmaxx, line 46: These values are a bit weird. But since there is no real UX design yet, this is fine for :)

eliseeborn: yeah, they're taken directly from the rows per page dropdown so that they look the same, but it is a bit messy with all the random css

PR #347, opened 2021-10-14
refactor: file structure + some file and variable names

### LINKED ISSUE #274: Refactoring
Couple of things that I think would simplify the structure:
- Rename `handle-cell-focus` to `handle-accessibility` since it then makes sense to include sr logic in there
- Move `handleNavigateTop` to `handle-scroll` since it has more with scrolling than updating the focus
- New folder structure
  - One for components
  - One for helper files, maybe call it `utils`
- Rename `selState` to `selectionState`

haxxmaxx:
fixes #274 
- All react files + renderer -> folder `components`
- All util files -> folder `utils`
- All files that `index` uses -> folder `src`
- move `handleNavigateTop` -> `handleScroll` file
- rename variable `selState` -> `selectionState`
- rename file `handle-cell-focus` -> `handle-accessibilty` (handles both focus and screen readers)
- fix codeclimate issue for too many arguments for `addSelectionListeners`


haxxmaxx:
> It will be good to add `handles both focus and screen readers` in handle-accessibilty as a comment.

Good idea. I've been thinking about adding jsdoc for all util functions, could be done in the same PR. If not jsdoc, there definitely could be some comments clarifying some flows/functions

src/table/components/__tests__/withSelections.spec.jsx
```diff
@@ -39,36 +39,46 @@ describe('withSelections', async () => {
...
 
-    expect(selectionsUtils.selectCell).to.have.been.calledWith(selState, cell);
+    expect(selectionsUtils.selectCell).to.have.been.calledWith(selectionState, cell);
   });
   it('should not call selectCell on mouseUp when measure', () => {
     cell.isDim = false;
 
-    const { queryByText } = render(<HOC selState={selState} cell={cell} selDispatch={selDispatch} styling={styling} />);
+    const { queryByText } = render(
+      <HOC selectionState={selectionState} cell={cell} selDispatch={selDispatch} styling={styling} />
```
a-m-dev, line 60: it could be awesome if we change this one as well `selDispatch`

PR #369, opened 2021-10-27
feat: set focus on nebula selection toolbar when tabbing

haxxmaxx:
Added `keyboard.focusSelection` of upcoming release of nebula. This enables us to focus one of the button in the selection toolbar. Since you navigate between the buttons in the selections toolbar with tab, this becomes a tab loop, which should be pretty intuitive for screen-reader users.


src/table/utils/handle-accessibility.js
```diff
@@ -69,6 +69,13 @@ export const handleFocusoutEvent = (evt, shouldRefocus, blur) => {
   }
 };
 
-// get the object, find the patent of the confirm button in selection toolbar parent and focus that element
-export const focusConfirmButton = (element) =>
-  element.closest('.qv-object-wrapper')?.querySelector('.sel-toolbar-confirm')?.parentElement?.focus();
+export const focusSelectionToolbar = (element, keyboard, last) => {
+  const clientConfirmButton = element
```
haxxmaxx, line 73: Don't have a better way of doing this. We could introduce some kind of inside client flag, but this works just fine

nebula.config.js
```diff
@@ -10,4 +10,7 @@ module.exports = {
   build: {
     replacementStrings,
   },
+  serve: {
+    keyboardNavigation: false,
```
haxxmaxx, line 14: This can be set to true to test the "sheet mode"

src/table/utils/handle-accessibility.js
niekvanstaveren: Which confirm button is this? Would it not make sense if this is also exposed in keyboard?

haxxmaxx: This is the confirm button in the client selection toolbar. The reason why we can focus the nebula one is that we can access the keyboard hook. keyboard don't have access to the client one

niekvanstaveren (approved):

PR #376, opened 2021-11-02
feat: make rendering testing work with enigma-mocker and test fixtures 

LiKang6688:
Nebula serve using mocked Qlik core engine (Enigma mocker) to render the visualization charts based on the testing fixture file

Testing cases using Puppeteer capture screenshots of those rendered charts and matches against the baseline to ensure our charts look as intended and present clients with expected visuals.

Visual regression Testing flow:

Start Nebula serve and launch Puppeteer
Iterate testing fixture files
Create test case per testing fixture file
Render chart based on testing fixture file in Nebula serve using Enigma mocker
Puppeteer capture screenshot
Compare screenshot with the baseline image


.circleci/config.yml
```diff
@@ -46,7 +46,7 @@ jobs:
   build:
     <<: *defaults
     docker:
-      - image: circleci/node:14.15.0
+      - image: circleci/node:16.7-browsers
```
haxxmaxx, line 49: Maybe we should bump in package too?

LiKang6688: good point.

LiKang6688: Thanks. Updated.

haxxmaxx: nice, make sure to do some sanity testing, could have some side-effects

haxxmaxx (approved):

PR #378, opened 2021-11-03
fix: add missing key to check if translated string is different from its id

LiKang6688:
if the translated string is different from its id, we assume translations already exist for the current locale

adding the key 'SNTable.Accessibility.RowsAndColumns' to do that to have a better performance.


haxxmaxx:
Or just run `registerLocale` once? Or is there some case where we need to rerun it?

haxxmaxx:
> > Or just run `registerLocale` once? Or is there some case where we need to rerun it?
> 
> Object.keys(all).forEach((key) => { translator.add(all[key]); }); only needs to be run once.
> 
> but registerLocale runs every time when we change the language no matter whether we put registerLocale in or out of useEffect.

Ah of course, in that case we need to rerun it sometimes. I guess you can have a separate useEffect for `registerLocale` but that is maybe overkill. Quite nice to get the logic away from index actually

haxxmaxx:
btw do we have any tests on locale files?

LiKang6688:
> btw do we have any tests on locale files?

no, we do not. I will add that in this PR.

LiKang6688:
> btw do we have any tests on locale files?

Done

haxxmaxx (approved):

haxxmaxx:
I just merged some new translations so you got some enoying conflicts, but feel free to merge when that is fixed

PR #379, opened 2021-11-04
[FIX#315]: SR announcements for pagination part of the table

### LINKED ISSUE #287: Buttons are read as unavailable without button label when changing page
If you press space on a button; for example first page, last page, previous page or next button, when a new page is loaded and button is no longer available the only thing read is unavailable. This could be misunderstood by the customers, because you are not told what is unavailable, but it could also be that the users understands that it is the button that is no longer available. We need to check this with A360.

### LINKED ISSUE #315: When changing page or changing the row page setting in the new table you get no information on which rows that now are displayed
When changing page or changing the row page setting in the new table you get no information on which rows that now are displayed. This information is just given when you enter the table, but would be needed when you change page settings or when you change page, otherwise it will be hard for the user to know on what table page they are.

a-m-dev:
fixes #287 
fixes #315 

This PR fixes all announcements mainly for pagination part of the table, including: 
1. changing page to the next/prev page
2. changing page to the very first/last page
3. changing showing items per page 
4. selecting the specific page,
by using the `announce` function that has been provided from `<TableWrapper />`

### current status of this PR: 
- [x] Resolve conflicts
- [x] Unit tests
- [x] Manual verification


locale/locales/en.json
```diff
@@ -31,6 +31,34 @@
     "value": "Last page",
     "comment": "tooltip and aria-label for the last page button. (tew 210929)"
   },
+  "SNTable.Pagination.MovedToFirstPage": {
+    "value": "You are in the first page.",
```
haxxmaxx, line 35:
```suggestion
    "value": "You are on the first page.",
```

haxxmaxx: and same for the rest

haxxmaxx: Actually these should be the same as `MovedToNextPage`
```suggestion
    "value": "Moved to the first page.",
```

locale/locales/en.json
```diff
@@ -31,6 +31,34 @@
...
+  "SNTable.Pagination.MovedToPreviousPage": {
+    "value": "Moved to the previous page.",
+    "comment": "Label for Switching to the previous page."
+  },
+  "SNTable.Pagination.PageStatusReport": {
+    "value": "You are currently in page {0} out of {1} pages.",
+    "comment": "Label for page status while changing pages."
+  },
+  "SNTable.Pagination.RowsPerPageChange": {
+    "value": "Rows per page has changed. Now showing {0} items on each page of table.",
```
haxxmaxx, line 55:
```suggestion
    "value": "Rows per page has changed. Now showing {0} rows on each page.",
```

haxxmaxx:
I think there are too many strings, it might actually cause confusion. At minimum we need:
- one for changing rows per page, announcing the new value
- one for changing page, announcing the new page and total page

And I think that could be enough actually

a-m-dev:
totally Agreed, you need to think about the statements all the time in order to figure out what they are used for!

So the first will handled by this string: "SNTable.Pagination.RowsPerPageChange"
but for the 2nd items you've mentioned, we need to combine all of these translations to one ( skipping the first/last page translations at all ):
1. SNTable.Pagination.MovedToFirstPage
2. SNTable.Pagination.MovedToLastPage
3. SNTable.Pagination.MovedToNextPage
4. SNTable.Pagination.MovedToPreviousPage
5. SNTable.Pagination.PageStatusReport
6. SNTable.Pagination.CustomePageSelection

locale/locales/en.json
a-m-dev: ✅ fixed

locale/locales/en.json
a-m-dev: ✅ fixed

locale/locales/en.json
```diff
@@ -31,6 +31,14 @@
     "value": "Last page",
     "comment": "tooltip and aria-label for the last page button. (tew 210929)"
   },
+  "SNTable.Pagination.RowsPerPageChange": {
+    "value": "Rows per page has changed. Now showing {0} rows on each page.",
+    "comment": "Label for rows per page change."
+  },
+  "SNTable.Pagination.PageStatusReport": {
+    "value": "You are currently on page {0} out of {1} pages.",
```
haxxmaxx, line 39: I think this one should be similar to the rowsPerPage one, maybe
```suggestion
    "value": "Page has changed. Now showing page {0} out of {1}.",
```

haxxmaxx:
Since we now use the same string for every page change, I think you could just do it from `handlePageChange`, same way as for rows per page, would simplify this a lot.

a-m-dev:
I dont think so, since the `handleChangePage` function uses page index rather than the actual page ( for example it uses index 0 for page 1, and totalPages -1 for the last page or it uses page + 1 for the next page but we are using page + 2 there ) 

one possible solution for this situation would be to pre-calculate the user-friendly page number in the function and pass it to the `handleChangePage` as third argument (like some argument named: `actualPageNumber`) in order to use it inside of the `announce` if we want to do so.

a-m-dev:
Just implemented the idea and updated the`handleChangePage` to receive two arguments `{ pageIdx, actualPageIdx }`

locale/locales/en.json
a-m-dev: ✅ Fixed

a-m-dev:
no, not in all cases, it could be the `page`, `page + 1`, `page + 2` as well

haxxmaxx:
> no, not in all cases, it could be the `page`, `page + 1`, `page + 2` and even `totalPages` as well

Then the second argument is still + 1 of the first argument, so you can do this
`  const handleChangePage = (pageId) => {
    setPageInfo({ ...pageInfo, page: pageIdx });
    announce({ keys: [['SNTable.Pagination.PageStatusReport', [pageIdx + 1, totalPages]]], politeness: 'assertive' });
  };`

a-m-dev:
Hmmm, now i understand what you mean. 
Great catch Max 👍

src/table/components/TablePaginationActions.jsx
```diff
@@ -55,22 +55,26 @@ const useStyles = makeStyles({
...
-    onPageChange(event, 0);
+  const handleFirstPageButtonClick = () => {
+    onPageChange(0);
   };
 
-  const handleBackButtonClick = (event) => {
-    onPageChange(event, page - 1);
+  const handleBackButtonClick = () => {
+    // we are doing announcement before page state update
+    // we dont need to subtract1, the value is already indicating the current page number
```
haxxmaxx, line 66: These comments do not apply anymore

locale/locales/en.json
```diff
@@ -31,6 +31,14 @@
     "value": "Last page",
     "comment": "tooltip and aria-label for the last page button. (tew 210929)"
   },
+  "SNTable.Pagination.RowsPerPageChange": {
+    "value": "Rows per page has changed. Now showing {0} rows on each page.",
```
haxxmaxx, line 35: Maybe we should say that you go back to the first page, since you always do that so something like
```suggestion
    "value": "Rows per page has changed. Now showing {0} rows on each page and moved to first page.",
```

src/table/components/TablePaginationActions.jsx
a-m-dev: ✅ Fixed

locale/locales/en.json
a-m-dev: you mean "the first row" right?

a-m-dev: ✅ fixed

src/table/components/TableWrapper.jsx
```diff
@@ -59,14 +69,21 @@ export default function TableWrapper(props) {
   const containerMode = constraints.active ? 'containerOverflowHidden' : 'containerOverflowAuto';
   const paginationHidden = constraints.active && 'paginationHidden';
   const paginationFixedRpp = selectionsAPI.isModal() || rect.width < 550;
-  const announce = useMemo(() => announcementFactory(rootElement, translator), [translator.language]);
+  const announce = announcer || useMemo(() => announcementFactory(rootElement, translator), [translator.language]);
```
a-m-dev, line 72: Since `announcer` function has been created inside of the `<TableWrapper />` component and passed down to the dom tree, there is no way to mock it or using `sinon.stub` in order to track it's calls. 

So the `announce` prop here is only for testing purpose here. 
If you think there is a better way for doing it, please let me know

haxxmaxx: Well, or don't test it. We already test the functionality of the page select in `TablePaginationActions.spec.jsx`. The test you are adding is not really testing much new, mostly adding coverage. I'd say it is fine if we miss those two rows, everything around it is tested

a-m-dev: but i think there is going to be a problem if we skip these tests, since we only test **the visibility** and  the calls of **`onPageChange`** function only in the `TablePaginationActions.spec.tsx` file.

I think we should have some checks around if we announce to the user in case of changes and **specifically what we are announcing to them.**

these two modified test cases on `TableWrapper.spec.jsx` file, only checks for that and the `TablePaginationActions.spec.jsx` file does not even checks for announcements.

we should prevent the possibility of if someone tries to remove/modifies the announce part of the code for pagination stuff by accident, and we can achieve it by these two test cases because (firstly if someone tries too, the test cases will break and secondly) we don't check for announcements of pagination somewhere else in our codebase.

i think it's better for us to keep them

haxxmaxx: Good point. Even though there is no real logic in these functions, it is still important that we do the correct announcement. Keep them but make a comment in the `props` deconstructing that this property should only be used for testing

src/table/components/__tests__/TableWrapper.spec.jsx
```diff
@@ -111,10 +113,45 @@ describe('<TableWrapper />', () => {
...
+        rect={rect}
+        announcer={announcer}
+      />
+    );
+
+    fireEvent.change(getByTestId('pagination-dropdown'));
+
+    expect(setPageInfo).to.have.been.calledOnce;
+    expect(announcer).to.have.been.calledOnceWith({
+      keys: [['SNTable.Pagination.PageStatusReport', sinon.match.any]],
```
haxxmaxx, line 152: Change `sinon.match.any` to the actual value of new page

src/table/components/__tests__/TableWrapper.spec.jsx
```diff
@@ -111,10 +113,45 @@ describe('<TableWrapper />', () => {
...
         rect={rect}
+        announcer={announcer}
       />
     );
+
     fireEvent.change(getByTestId('select'), { target: { value: 25 } });
+
     expect(setPageInfo).to.have.been.calledWith({ page: 0, rowsPerPage: 25 });
+    expect(announcer).to.have.been.calledOnceWith({
+      keys: [['SNTable.Pagination.RowsPerPageChange', sinon.match.any]],
```
haxxmaxx, line 124: same here

haxxmaxx: actually make 25 into a const, so we see that the value we fire is the same as `setPageInfo` and `announcer` is called with

src/table/components/TableWrapper.jsx
a-m-dev: ✅ Done!

src/table/components/__tests__/TableWrapper.spec.jsx
a-m-dev: ✅ Fixed

src/table/components/__tests__/TableWrapper.spec.jsx
a-m-dev: ✅ Fixed

src/table/components/__tests__/TableWrapper.spec.jsx
```diff
@@ -111,10 +113,46 @@ describe('<TableWrapper />', () => {
         keyboard={keyboard}
         translator={translator}
         rect={rect}
+        announcer={announcer}
       />
     );
+
     fireEvent.change(getByTestId('select'), { target: { value: 25 } });
```
haxxmaxx, line 120: use `targetRowsPerPage` here as well

haxxmaxx (approved):

src/table/components/__tests__/TableWrapper.spec.jsx
a-m-dev: ✅ Fixed

src/locale/locales/en.json
```diff
@@ -31,6 +31,14 @@
     "value": "Last page",
     "comment": "tooltip and aria-label for the last page button. (tew 210929)"
   },
+  "SNTable.Pagination.RowsPerPageChange": {
+    "value": "Rows per page has changed. Now showing {0} rows on each page and moved to the first page.",
```
tklkan, line 35: This string won't work for localization. In a number of the languages we translate in, there are multiple plural forms. This repo doesn't have the pluralization jsons that other repos have yet, so we'll have to rework these strings. Something like:

`Rows per page has changed (Rows: {0}). Now showing the first page.` 
`Rows per page has changed {0}. Now showing the first page.`

tklkan (approved):

src/locale/locales/en.json
```diff
@@ -31,6 +31,14 @@
     "value": "Last page",
     "comment": "tooltip and aria-label for the last page button. (tew 210929)"
   },
+  "SNTable.Pagination.RowsPerPageChange": {
+    "value": "Rows per page has changed (Rows: {0}). Now showing the first page.",
```
haxxmaxx, line 35: @tklkan I think your other suggestion was better `Rows per page has changed to {0}. Now showing the first page.`. Or would that have the same problem?

tklkan: Actually that one should work. Let me update this.

PR #385, opened 2021-11-07
refactor: use react context api

LiKang6688:


LiKang6688:
Hi @a-m-dev, I just saw your comment in this linked issue. Could I have your help to have a look at my way of using the react context API?

haxxmaxx:
This is a good start. Would be nice to have `announce` on the context as well. So now it fits better in index and it could be memoized in a nice way :)

a-m-dev:
> Hi @a-m-dev, I just saw your comment in this linked issue. Could I have your help to have a look at my way of using the react context API?

sorry about being a bit late, but let me check it asap

src/contexts/rootContext.js
```diff
@@ -0,0 +1,4 @@
+import { createContext } from 'react';
+
+const RootContext = createContext();
```
a-m-dev, line 3: we could have a hook for using the context value exported from this file and then use it everywhere instead of importing `useContext` and the context itself, something like: 
```jsx
export const RootContext = createContext()
export const useRootContext = () => useContext(RootContext)
```

src/table/Root.jsx
```diff
@@ -12,7 +13,9 @@ export function render(rootElement, props) {
       <StyledEngineProvider injectFirst>
         <StylesProvider generateClassName={muiParameters.generateClassName}>
           <ThemeProvider theme={muiParameters.theme}>
-            <TableWrapper {...props} />
+            <RootContext.Provider value={props}>
```
a-m-dev, line 16: there is two situations for providing the values to the component tree. 
1. one is providing it from here
2. the other one is providing it from Table wrapper, since we might need some stuff from inside of table wrapper to be available for us in our context


so i think it could be really better if we could have an isolated layer between `<ThemeProvider />` and `<TableWrapper/>` component for our context and all logics should goes inside of that. 

in future, if we want to switch for custom hooks for each of the works that we do for table (selection, sorting, pagination, a11y and ... ) that layer would helps us a lot, and it could be the single source for our logic. 

just consider the context layer as a hub, that gathers all of the required data variables and functions ( as well as utils hooks ) inside of the context itself and them memoizes  that and exposes into the component tree.

src/table/components/TableBodyWrapper.jsx
```diff
@@ -8,6 +8,7 @@ import getCellRenderer from './renderer';
 import { getBodyStyle } from '../utils/styling-utils';
 import { bodyHandleKeyPress } from '../utils/handle-key-press';
 import { handleClickToFocusBody } from '../utils/handle-accessibility';
+import RootContext from '../../contexts/rootContext';
```
a-m-dev, line 11: this was the reason for creating the custom hook of `useRootContext` instead of importing two items on each use case

src/table/components/TableBodyWrapper.jsx
```diff
@@ -27,19 +28,9 @@ const useStyles = makeStyles({
...
-  theme,
-  setShouldRefocus,
-  setFocusedCellCoord,
-  keyboard,
-  tableWrapperRef,
-  announce,
-}) {
+function TableBodyWrapper({ setShouldRefocus, setFocusedCellCoord, tableWrapperRef, announce }) {
+  const { rootElement, layout, tableData, constraints, selectionsAPI, keyboard, theme } = useContext(RootContext);
+  console.log({ rootElement });
```
a-m-dev, line 33: log gotcha here

src/table/components/TableHeadWrapper.jsx
```diff
@@ -8,6 +8,7 @@ import TableSortLabel from '@mui/material/TableSortLabel';
 import { getHeadStyle } from '../utils/styling-utils';
 import { headHandleKeyPress } from '../utils/handle-key-press';
 import { handleClickToFocusHead } from '../utils/handle-accessibility';
+import RootContext from '../../contexts/rootContext';
```
a-m-dev, line 11: same here

src/table/components/TableHeadWrapper.jsx
```diff
@@ -33,19 +34,9 @@ const useStyles = makeStyles({
...
-  layout,
-  changeSortOrder,
-  constraints,
-  translator,
-  selectionsAPI,
-  focusedCellCoord,
-  setFocusedCellCoord,
-  keyboard,
-}) {
+function TableHeadWrapper({ focusedCellCoord, setFocusedCellCoord }) {
```
a-m-dev, line 37: if we have our context as a separate layer, we can move the focused cell logic into the context by it's own custom hook and get all required stuff again from context

this is the core benefit of context fo us.

src/table/components/TableWrapper.jsx
```diff
@@ -13,6 +12,7 @@ import { handleTableWrapperKeyDown } from '../utils/handle-key-press';
 import { updateFocus, handleResetFocus, handleFocusoutEvent } from '../utils/handle-accessibility';
 import { handleScroll, handleNavigateTop } from '../utils/handle-scroll';
 import announcementFactory from '../utils/announcement-factory';
+import RootContext from '../../contexts/rootContext';
```
a-m-dev, line 15: same here as well for extra import

src/table/components/TableWrapper.jsx
```diff
@@ -140,9 +140,8 @@ export default function TableWrapper(props) {
             `${columns.length}`,
           ])}
         >
-          <TableHeadWrapper {...props} setFocusedCellCoord={setFocusedCellCoord} focusedCellCoord={focusedCellCoord} />
+          <TableHeadWrapper setFocusedCellCoord={setFocusedCellCoord} focusedCellCoord={focusedCellCoord} />
```
a-m-dev, line 143: same here as well. 
if we move the logic for focused cell coords into the hook inside of context, we wont pass props to the functions like this one

src/table/components/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -12,30 +12,28 @@ import * as selectionsUtils from '../../utils/selections-utils';
...
-
-    tableData = await manageData(model, generateLayout(1, 1), { top: 0, height: 100 });
-    constraints = {};
-    selectionsAPI = {
-      isModal: () => true,
-    };
-    theme = {
-      getColorPickerColor: () => {},
-      name: () => {},
+    providerProps = {
```
a-m-dev, line 24: i think a test util function is required here in order to get some arguments (based on the test case that we are writing it) and gives us whatever react-testing-library apis, or something similar to that. 

the purpose should be only prevent  providing the context in all test cases using `<RootContext.Provider value={...}>`

src/table/components/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -52,35 +50,27 @@ describe('<TableBodyWrapper />', async () => {
...
 
     const { queryByText } = render(
-      <TableBodyWrapper
-        tableData={tableData}
-        constraints={constraints}
-        selectionsAPI={selectionsAPI}
-        theme={theme}
-        layout={layout}
-      />
+      <RootContext.Provider value={providerProps}>
```
a-m-dev, line 69: same as test util here

a-m-dev:
the general idea of context manager should be: 
think of it as a hub, a hub of utils functions, data variables, api fetchers, custom hooks per each of useSelection, useSort, useAnnounce, usePagination and other hooks (which is our future work should toward to). 

it should gather all of stuff from hooks and utils and data variables, and then provide it to the component tree. 

by doing this way, our table will easily becomes maintainable and extendable, because if we asked foradd a new feature to the table, the only work for us should be creating a custom hook for the functionality and putting it into the context that we have.

haxxmaxx:
Can this be closed since @a-m-dev started working on it?

PR #389, opened 2021-11-10
[Fix#331]:  Fix for concurrent announcement while changing page

### LINKED ISSUE #331: Incorrect or no information is read in the table when switching page in table using ctrl+shift + arrow in selection state
Incorrect or no information is read in the table when switching page in table using ctrl+shift + arrow in selection state.

Precondition: Use NVDA screenreader + Chrome

1. tab to a visualization
2. press enter to view data
3. press enter to enter the table
4. Use arrows to move to a column with a dimension value in the table
5. Press space to select the value
6. Press ctrl+shitf + right/left arrow to change page in the table
 -> column value that you moved from is read, not the column you moved to
7. Press ctrl+shitf + right/left arrow to change page in the table again
8. Nothing is read

a-m-dev:
fixes #331 

this is a draft PR for the issue 

in order to test this, you need to merge #315 first in the branch and then test it on PTE, since #315 introduces the announcement for pagination and this one fixed the focus announcement

that is the point that concurrency happens and this PR fixes that!

### current status of this PR: 
- [x] Resolve conflicts
- [x] Unit tests
- [x] Manual verification


src/table/components/TableWrapper.jsx
```diff
@@ -125,7 +126,8 @@ export default function TableWrapper(props) {
         })
       }
     >
-      <div id="sn-table-announcer" aria-live="polite" aria-atomic="true" className={classes.screenReaderOnly} />
```
a-m-dev, line 128: i thinks it would be better if we have a component that holds these two elements 
are you agree?

src/table/utils/handle-accessibility.js
```diff
@@ -58,7 +59,20 @@ export const handleResetFocus = ({
     const focusType = shouldRefocus.current ? 'focus' : 'addTab';
     shouldRefocus.current = false;
     const rowElements = rootElement.getElementsByClassName('sn-table-row');
-    updateFocus({ focusType, rowElements, cellCoord: nextCell });
```
a-m-dev, line 61: these will be cleaned up in the main PR, no worries on these

haxxmaxx:
Not sure if this is worth it. Adding another element, a bunch of logic to keep track of which one to set, and a special case in `resetFocus` just to fix a specific case. Let's discuss on daily

a-m-dev:
by having another `aria-live` element, we are handling concurrency. as you remember we were checking out the `react-live-announcer` package and they were using a similar solution for it ( [source code here](https://github.com/AlmeroSteyn/react-aria-live/blob/0e399857b7e3d0da6d8e5b678814433c3d825237/src/modules/Announcer.js#L75) )

in case of modifying the `reasetFocus` function, i think that is time that we can detect the focus has been reset while we having the selection.

haxxmaxx:
Hmm ok sure. It is not too much, as long as it has no side effects :+1:

src/table/components/TableWrapper.jsx
haxxmaxx: yes that would be nice

src/table/utils/announcement-factory.js
```diff
@@ -35,7 +46,21 @@ export default function announcementFactory(rootElement, translator, junkCharIdx
...
     hasJunkChar++;
 
-    const announceElement = rootElement.querySelector('#sn-table-announcer');
+    const announceElement01 = rootElement.querySelector('#sn-table-announcer--01');
+    const announceElement02 = rootElement.querySelector('#sn-table-announcer--02');
+
+    let announceElement = null;
+    if (lastAnnounceElement === null) {
+      announceElement = announceElement01;
+      lastAnnounceElement = announcerElements.first;
```
haxxmaxx, line 55: This naming is a confusing. `previousAnnounceElement` would be better

src/table/utils/handle-accessibility.js
```diff
@@ -58,7 +59,20 @@ export const handleResetFocus = ({
...
+    // updateFocus({ focusType, rowElements, cellCoord: nextCell });
+
+    const providedCell = rowElements[nextCell[0]]?.getElementsByClassName('sn-table-cell')[nextCell[1]];
+    updateFocus({ focusType, providedCell });
+
+    const hasActiveClassName = providedCell?.classList?.contains('selected');
```
haxxmaxx, line 67: Move this inside the if block. And name it `hasSelectedClassName`.

src/table/utils/announcement-factory.js
```diff
@@ -35,7 +46,21 @@ export default function announcementFactory(rootElement, translator, junkCharIdx
...
+    } else if (lastAnnounceElement && lastAnnounceElement === announcerElements.first) {
+      announceElement = announceElement02;
+      lastAnnounceElement = announcerElements.second;
+    } else {
+      announceElement = announceElement01;
+      lastAnnounceElement = announcerElements.first;
+    }
```
haxxmaxx, line 62: Had trouble understanding this. The following should do the same thing
```suggestion
    if (lastAnnounceElement === announcerElements.first) {
      announceElement = announceElement02;
      lastAnnounceElement = announcerElements.second;
    } else {
      announceElement = announceElement01;
      lastAnnounceElement = announcerElements.first;
    }
```

src/table/components/TableWrapper.jsx
a-m-dev: ✅   Fixed

src/table/utils/handle-accessibility.js
a-m-dev: ✅.  Done

src/table/utils/announcement-factory.js
a-m-dev: ✅.  Fixed

src/table/utils/handle-accessibility.js
a-m-dev: ✅.  Done

src/table/utils/announcement-factory.js
a-m-dev: ✅.  Fixed

src/table/components/Announce.jsx
```diff
@@ -0,0 +1,27 @@
...
+    clipPath: 'inset(50%)',
+    height: '1px',
+    overflow: 'hidden',
+    position: 'absolute',
+    whiteSpace: 'nowrap',
+    width: '1px',
+  },
+});
+
+const Announce = () => {
```
haxxmaxx, line 16: `Announcer` would make more sense

haxxmaxx: Or something that is a not the same as the `announce` function

a-m-dev: I think `AnnounceElements` would fits well in this case, because `Announcer` refers to the functionality but these are just elements for announcing. so in the latest commit i will switch the name to "AnnounceElements"

src/table/components/AnnounceElements.jsx
```diff
@@ -0,0 +1,27 @@
...
+  },
+});
+
+const AnnounceElements = () => {
+  const classes = useStyles();
+
+  return (
+    <>
+      <div id="sn-table-announcer--01" aria-live="polite" aria-atomic="true" className={classes.screenReaderOnly} />
+      <div id="sn-table-announcer--02" aria-live="polite" aria-atomic="true" className={classes.screenReaderOnly} />
```
haxxmaxx, line 22: We should make sure that all id's and non-encrypted classes start with `sn-table`

haxxmaxx (approved):

src/table/components/AnnounceElements.jsx
a-m-dev: Sure, I'll check the entire project

src/table/utils/announcement-factory.js
```diff
@@ -32,11 +42,22 @@ export default function announcementFactory(rootElement, translator, junkCharIdx
...
+    if (previousAnnouncementElement === announcerElements.first) {
+      announceElement = announceElement02;
+      previousAnnouncementElement = announcerElements.second;
+    } else {
+      announceElement = announceElement01;
+      previousAnnouncementElement = announcerElements.first;
+    }
+
+    if (announceElement.innerHTML.endsWith(` ­`)) announceElement.innerHTML = notation;
+    else announceElement.innerHTML = `${notation} ­`;
```
haxxmaxx, line 59:
```suggestion
    announceElement.innerHTML = announceElement.innerHTML.endsWith(` ­`)) ? notation : `${notation} ­`;
```

a-m-dev: of course 
✅ fixed as well

PR #390, opened 2021-11-12
fix: the type of value supplied to tabIndex is wrong

LiKang6688:
Invalid prop `tabIndex` of type `string` supplied to `ForwardRef(ButtonBase)`, expected `number`. I thought it was fixed before, but just a minor issue.


haxxmaxx (approved):
I must have changed it when I added `keyboard.enabled`

PR #394, opened 2021-11-15
refactor: generalize action buttons

haxxmaxx:
Been thinking about how to remove the duplicate code here for a while. This is obviously more compact and general, but is it as readable as it was before? If hard to understand I'm not sure this is an approvement


eliseeborn:
I think this makes seeing what the differences between the buttons are clearer, but it might make it a little harder to read the first time you're looking at it? 
Either way I think it's better this way :)

haxxmaxx:
> I think this makes seeing what the differences between the buttons are clearer, but it might make it a little harder to read the first time you're looking at it? Either way I think it's better this way :)

Yeah that is my take too, then I think we go with this

eliseeborn (approved):

PR #395, opened 2021-11-16
[FIX#323]: Adding announce for when user focuses on selection toolbox 

### LINKED ISSUE #323: Screen reader does not read that we have moved to the selection toolbar or the buttons within it
Screen reader does not read that vi have moved to the selection toolbar or the buttons within it

a-m-dev:
fix #323 

Based on @haxxmaxx work related to nebula selection, just a bit tweaks and modifying the test cases to support (better to say on tests perspective, to make sure that we are) announce while user focuses on selection toolbar!


src/locale/locales/en.json
```diff
@@ -3,6 +3,10 @@
     "value": "Showing {0} rows and {1} columns.",
     "comment": "number of rows and columns announced by the screen reader. (tew 210929)"
   },
+  "SNTable.Accessibility.SelectionToolbar": {
+    "value": "Selection Toolbar",
```
haxxmaxx, line 7:
```suggestion
    "value": "Moved to selection Toolbar",
```

src/table/utils/handle-accessibility.js
```diff
@@ -69,12 +69,13 @@ export const handleFocusoutEvent = (evt, shouldRefocus, keyboard) => {
...
 };
 
-export const focusSelectionToolbar = (element, keyboard, last) => {
+export const focusSelectionToolbar = (element, keyboard, last, announce) => {
   const clientConfirmButton = element
     .closest('.qv-object-wrapper')
     ?.querySelector('.sel-toolbar-confirm')?.parentElement;
   if (clientConfirmButton) {
     clientConfirmButton.focus();
+    announce({ keys: 'SNTable.Accessibility.SelectionToolbar' });
```
haxxmaxx, line 78: This should run in both cases, so put it before the if block

src/locale/locales/en.json
a-m-dev: ✅ Fixed

src/table/utils/handle-accessibility.js
a-m-dev: ✅ Fixed

haxxmaxx (approved):

haxxmaxx:
Wait to merge this until I have merged `crl/focus-sel-nebula`

haxxmaxx:
Close this since we will not solve this inside the table

PR #399, opened 2021-11-22
feat: sync version between package.json and api-spec

LiKang6688:
whenever you make a release,  the version in the api-spec  should always be equal to the version of the table


haxxmaxx (approved):
🥇

haxxmaxx:
pls merge if it is done :)

PR #403, opened 2021-11-23
[WIP] Refactor: General idea behind `useRootContext`

### LINKED ISSUE #357: Refactoring code to apply the useContext and useReducer hooks together

a-m-dev:
fixes #357 

These simple changes is based on @LiKang6688 great job on the [context PR](https://github.com/qlik-oss/sn-table/pull/385), with a little bit of tweaks, it makes us able to separate each domain of the table (like selection, scrolling, pagination, a11y, ...) by custom hook and provide it through `RootContextManager` which is available for us by the context in entire project. 

in this PR, as you can see I just took the functionality of scroll part of the table and separated in a custom hook, this does a bit less to be honest but in it's turn, it cleared a bit of complexity in our `<TableWrapper />` component regarding the scrolling. 

you may still look at the `<TableWrapper />` component and say that it does not made any differences but if we start doing other thing specially the selection or a11y part, you will see the difference because that domain of table holds the most part of the logic i guess.

the important point is to separate logic to custom hooks, these (custom) hooks can return some values or not, based on the job that they are doing! in this PR's case we are only tracking the `scroll` and do something based on that, so no need to return anything from hook, but in case of things like selection or a11y, we should and provide it through context to being used by other stuff.

also you will see a so called **renderProps pattern** in the `RootContextManager` file, which is required to be able to use the created ref directly inside of the place that we are using the context. since we only have to use the value of context inside of the context provider in our components, and we are providing our context in our `<TableWrapper />` component, we are not going to be able to use the context value directly in our `<TableWrapper />` so calling children as function in our `RootContextManager` file (which literally is our renderProps here) is what we need to do there.


src/table/components/TableBodyWrapper.jsx
```diff
@@ -27,23 +29,15 @@ const useStyles = makeStyles({
...
-}) {
+function TableBodyWrapper({ setShouldRefocus, setFocusedCellCoord, tableWrapperRef }) {
+  const {
+    nebulaProps: { rootElement, layout, tableData, constraints, selectionsAPI, keyboard, theme },
+  } = useRootContext();
   const { rows, columns } = tableData;
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
   const bodyStyle = useMemo(() => getBodyStyle(layout, theme), [layout, theme.name()]);
   const classes = useStyles(bodyStyle);
+  const announce = useAnnounce();
```
haxxmaxx, line 40: This is great and all, but I'm curious what it looks like when we have tackled a bunch of the utility functions. Now we get `announce` from `useAnnounce`, but we can only fetch it like this in a file using react. Later in this file we pass it to `bodyHandleKeyPress`. Is there a way of getting around that? defining `bodyHandleKeyPress` somewhere so you can run `useAnnounce` to get it, so we don't have to pass it?

a-m-dev: Yes, absolutely

later when we are dealing with useA11y hook, we will initiate our factory about announcement inside of `useA11y` hook and then provide it thought the context, which will be available to use thought the entire project.

in case of util functions like `bodyHandleKeyPress` as much as we can, we will move the functionality inside of the hook, but the general rule still will be there for util functions. this means that util functions will still kept as utilities  as much as we agree on their functionality.

src/contexts/RootContextManager.js
```diff
@@ -0,0 +1,40 @@
...
+ * this manager will cover:
+ *  1. selection
+ *  2. scroll ✅
+ *  3. focus
+ *  4. a11y
+ *  5. pagination
+ *  6. utils as well
+ * @returns contextValueObj
+ */
+export const RootContextManager = ({ focusedCellCoord, ...props }) => {
```
a-m-dev, line 15: `focusedCellCoord` will be moved into the related hook

PR #406, opened 2021-11-29
fix: make sure page for getHyperCubeData is never more than 10k cells

haxxmaxx:
- When there is more than 100 columns you can't change the rows per page manually
- Set height to the smallest possible value. On the last page we always fetched with `qHeight: rowsPerPage` even though there was usually fewer rows
- When there are > 10k cells, set rows per page to a lower options, e.g. 150 cols, 100 rows -> 150 cols, 25 rows. If the lowest option is too high (10 currently, meaning > 1000 columns...) It is set to 10k / columns.

I have tested with > 100 columns, but not > 1000.

### DIFF
--- a/src/__test__/handle-data.spec.js
+++ b/src/__test__/handle-data.spec.js
@@ -129,5 +129,14 @@ describe('handle-data', () => {
       expect(tableData).to.be.null;
       expect(setPageInfo).to.have.been.calledOnceWith({ ...pageInfo, rowsPerPage: 4 });
     });
+
+    it('should return null and call setPageInfo with rowsPerPage 4 when width > 10000', async () => {
+      pageInfo = { ...pageInfo, page: 0 };
+      layout = generateLayout(6000, 6000, 100);
+      const tableData = await manageData(model, layout, pageInfo, setPageInfo);
+
+      expect(tableData).to.be.null;
+      expect(setPageInfo).to.have.been.calledOnceWith({ ...pageInfo, rowsPerPage: 0 });
+    });
   });
 });
--- a/src/__test__/handle-data.spec.js
+++ b/src/__test__/handle-data.spec.js
@@ -7,7 +7,7 @@ describe('handle-data', () => {
   let colIdx;
 
   beforeEach(() => {
-    layout = generateLayout(2, 2, [1, 2, 0, 3]);
+    layout = generateLayout(2, 2, 200, [1, 2, 0, 3]);
   });
 
   describe('getColumnInfo', () => {
@@ -24,7 +24,7 @@ describe('handle-data', () => {
     });
 
     it('should return column info for dimension', () => {
-      const columnInfo = getColumnInfo(layout, colIdx);
+      const columnInfo = getColumnInfo(layout.qHyperCube, colIdx);
       expect(columnInfo).to.eql(getExpectedInfo(colIdx, true));
     });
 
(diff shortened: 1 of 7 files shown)


cbt1 (approved):
LGTM

src/handle-data.js
```diff
@@ -3,14 +3,20 @@ const directionMap = {
   D: 'desc',
 };
 
-export function getColumnOrder(layout) {
-  const { qColumnOrder, qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
+const MAX_CELLS = 10000;
```
LiKang6688, line 6: how about MAX_CELLS_PER_PAGE

LiKang6688: cells = rows per page x columns per page

haxxmaxx: But it is the max number of cells?

src/table/components/TableWrapper.jsx
```diff
@@ -14,6 +14,7 @@ import { handleTableWrapperKeyDown } from '../utils/handle-key-press';
 import { updateFocus, handleResetFocus, handleFocusoutEvent } from '../utils/handle-accessibility';
 import { handleScroll, handleNavigateTop } from '../utils/handle-scroll';
 import announcementFactory from '../utils/announcement-factory';
+// import { ROWS_PER_PAGE_OPTIONS } from '../../handle-data';
```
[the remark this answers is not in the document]
haxxmaxx: yes

src/handle-data.js
```diff
@@ -29,24 +35,30 @@ export function getColumnInfo(layout, colIndex) {
...
 export default async function manageData(model, layout, pageInfo, setPageInfo) {
-  const { page, rowsPerPage } = pageInfo;
-  const top = page * rowsPerPage;
+  const { page, rowsPerPage, rowsPerPageOptions } = pageInfo;
   const size = layout.qHyperCube.qSize;
-  const allRowsLength = size.qcy;
+  const top = page * rowsPerPage;
+  const width = size.qcx;
+  const totalHeight = size.qcy;
+  const height = Math.min(rowsPerPage, totalHeight - top);
```
LiKang6688, line 43: based on https://qlik.dev/apis/json-rpc/qix/genericobject#%23%2Fentries%2FGenericObject%2Fentries%2FGetHyperCubeData,

how about 
const qHyperCube = layout.qHyperCube;
const size = qHyperCube.qiszel
const top = page * rowsPerPage;
const width = size.qcx;
const height = size.qcy;
const actualHeight = Math.min(rowsPerPage, totalHeight - top);

src/__test__/handle-data.spec.js
```diff
@@ -7,7 +7,7 @@ describe('handle-data', () => {
   let colIdx;
 
   beforeEach(() => {
-    layout = generateLayout(2, 2, [1, 2, 0, 3]);
+    layout = generateLayout(2, 2, 200, [1, 2, 0, 3]);
```
LiKang6688, line 10: is it possible to have test cases to cover that the amount of cells  is larger than 10k?

src/handle-data.js
haxxmaxx: I don't see how the docs makes your suggestion make more sense. My thought was that width and height is what will be used for the page.

src/__test__/handle-data.spec.js
haxxmaxx: yes that is a good idea

haxxmaxx: or we have two of those actually

haxxmaxx: We could add one for columns > 10k. It will set rowsPerPage to 0, which is ok I think. We shouldn't support more than 10k columns, there is no useful use-case for that 😅

src/handle-data.js
LiKang6688: Screenshot 2021-11-30 at 09 41 43
I thought that width and height are what we got from the engine and make it clear that actualHeight is what is used for the page. now I think that both are fine.

LiKang6688 (approved):

PR #410, opened 2021-12-08
fix: remove hidden column from columnOrder

haxxmaxx:
- Remove the dependency of the `columnOrder` array outside `manageData`. Keep the "data" index of the column in the column info instead
- Use the column sort order from the properties instead of layout. `layout.qEffectiveInterColumnSortOrder` doesn't contain hidden column, thus it's ending the wrong thing in the patches. Fetching the properties makes the sorting function asynchronous, but that shouldn't matter

### DIFF
--- a/src/sorting-factory.js
+++ b/src/sorting-factory.js
@@ -1,7 +1,7 @@
 export default function sortingFactory(model, columnOrder) {
   return (layout, isDim, columnIndex) => {
     const sortIdx = columnOrder[columnIndex];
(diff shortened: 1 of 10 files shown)


src/sorting-factory.js
```diff
@@ -31,6 +33,7 @@ export default function sortingFactory(model, columnOrder) {
       });
     }
 
+    console.log(patches);
```
T-Wizard, line 36: remove

T-Wizard (approved):

LiKang6688 (approved):

PR #414, opened 2021-12-31
feat: apply theme from sense

LiKang6688:
Screenshot 2021-12-31 at 02 05 41


src/table/components/TableHeadWrapper.jsx
```diff
@@ -45,8 +46,11 @@ function TableHeadWrapper({
   focusedCellCoord,
   setFocusedCellCoord,
   keyboard,
+  backgroundColor,
 }) {
   const headStyle = useMemo(() => getHeadStyle(layout, theme), [layout, theme.name()]);
+  // transparent is not allowed
+  headStyle.backgroundColor = backgroundColor === 'transparent' ? 'inherent' : backgroundColor;
```
haxxmaxx, line 53: This should be done in `getHeadStyle`

src/table/components/TableWrapper.jsx
```diff
@@ -50,14 +50,19 @@ export default function TableWrapper(props) {
...
   const { size, rows, columns } = tableData;
   const { page, rowsPerPage, rowsPerPageOptions } = pageInfo;
   const [focusedCellCoord, setFocusedCellCoord] = useState([0, 0]);
   const shouldRefocus = useRef(false);
   const tableContainerRef = useRef();
   const tableWrapperRef = useRef();
-  const classes = useStyles();
+
+  const backgroundColorFromTheme = theme.getStyle('object', 'straightTable', 'backgroundColor');
+  const backgroundColor = !backgroundColorFromTheme ? 'inherit' : backgroundColorFromTheme;
```
haxxmaxx, line 63:
```suggestion
  const backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor') || 'inherit';
```

src/table/components/TableHeadWrapper.jsx
haxxmaxx, line 53: Is this only for the header? Can you have transparent on the rest of the table?

src/table/components/TableWrapper.jsx
```diff
@@ -50,14 +50,19 @@ export default function TableWrapper(props) {
...
   } = props;
   const { size, rows, columns } = tableData;
   const { page, rowsPerPage, rowsPerPageOptions } = pageInfo;
   const [focusedCellCoord, setFocusedCellCoord] = useState([0, 0]);
   const shouldRefocus = useRef(false);
   const tableContainerRef = useRef();
   const tableWrapperRef = useRef();
-  const classes = useStyles();
+
+  const backgroundColorFromTheme = theme.getStyle('object', 'straightTable', 'backgroundColor');
```
haxxmaxx, line 62: Unless `'straightTable'` is used for view data we might want to consider having a separate property for this table. Name TBD in that case I suppose

src/table/components/TableHeadWrapper.jsx
LiKang6688: yes, we can have transparent on the rest of the table

haxxmaxx: why not the header?

haxxmaxx:
what would the equivalent old table look like? For view data this might worsen the user experience, since the user can't control the font colors. In the image you can barely see the text on the black background

LiKang6688:
A new PR to replace this one will come soon

PR #416, opened 2022-01-03
feat: support right-to-left reading order

LiKang6688:


haxxmaxx (approved):
If it is this simple it is awesome. Make sure to try what happens when you have manually changed the alignment for different columns :)

LiKang6688:
> If it is this simple it is awesome. Make sure to try what happens when you have manually changed the alignment for different columns :)

Any suggestions of steps to change the alignment for different columns?

PR #417, opened 2022-01-04
refactor: make table pagination look better

LiKang6688:
Before:

Screenshot 2022-01-04 at 18 17 04

Screenshot 2022-01-04 at 18 15 35

After:

Screenshot 2022-01-04 at 18 11 24
Screenshot 2022-01-04 at 18 11 47


src/table/components/TableWrapper.jsx
```diff
@@ -54,10 +56,40 @@ export default function TableWrapper(props) {
...
+  const [paginationHeight, setPaginationHeight] = useState();
   const shouldRefocus = useRef(false);
   const tableContainerRef = useRef();
   const tableWrapperRef = useRef();
-  const classes = useStyles();
+  const tablePaginationSectionRef = useRef();
+
+  useEffect(() => {
+    return () => {};
+  }, []);
```
haxxmaxx, line 67: what is this?

src/table/components/TableWrapper.jsx
```diff
@@ -54,10 +56,40 @@ export default function TableWrapper(props) {
...
+
+  useEffect(() => {
```
haxxmaxx, line 69: Why move the useEffect?

haxxmaxx:
The weird thing to begin with is that the height of the pagination varies. We should be able to set it to a fixed value, instead of calculating the height of the table depending on what the pagination height ends up being

src/table/components/TableWrapper.jsx
LiKang6688: because of `const classes = useStyles({ paginationHeight }); ` want to have paginationHeight as soon as possible

LiKang6688:
> the height of the pagination varies

the height of the pagination varies is from https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/TablePagination/TablePagination.js#L31.
It makes sense to have different heights of the pagination section on different devices.

LiKang6688:
> The weird thing to begin with is that the height of the pagination varies. We should be able to set it to a fixed value, instead of calculating the height of the table depending on what the pagination height ends up being

the height of the pagination varies is from https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/TablePagination/TablePagination.js#L31.
It makes sense to have different heights of the pagination section on different devices.

haxxmaxx:
> > The weird thing to begin with is that the height of the pagination varies. We should be able to set it to a fixed value, instead of calculating the height of the table depending on what the pagination height ends up being
> 
> the height of the pagination varies is from https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/TablePagination/TablePagination.js#L31. It makes sense to have different heights of the pagination section on different devices.

Huh, didn't know it could have different sizes. Does the smaller once work with the custom components we have in there now?

src/table/components/TableWrapper.jsx
haxxmaxx: Does it make a difference? since it is a useEffect, does it matter where you define it? the function will run depending on the render cycle right?

LiKang6688: right, i had a try and there is no difference.

src/table/components/TableWrapper.jsx
```diff
@@ -85,16 +91,23 @@ export default function TableWrapper(props) {
   useEffect(() => {
     const memoedContainer = tableContainerRef.current;
     const memoedWrapper = tableWrapperRef.current;
-    if (!memoedContainer || !memoedWrapper) return () => {};
-
+    const tablePaginationSection = tablePaginationSectionRef.current;
+    if (!memoedContainer || !memoedWrapper || !tablePaginationSection) return () => {};
+    const handleResize = () => {
+      setPaginationHeight(tablePaginationSection.clientHeight);
+    };
```
haxxmaxx, line 98:
```suggestion
    const handleResize = () => setPaginationHeight(tablePaginationSection.clientHeight);
```

src/table/components/TableWrapper.jsx
```diff
@@ -85,16 +91,23 @@ export default function TableWrapper(props) {
...
+    handleResize();
```
haxxmaxx, line 99: should this run here?

LiKang6688: Yes, it needs to be run at the first time

src/mui-setup.js
```diff
@@ -19,6 +19,12 @@ export default function muiSetup() {
       boxShadow: '0 0 0 2px #3f8ab3 inset',
       outline: 'none',
     };
+    muiConfig.overrides.MuiToolbar = {
+      regular: {
+        minHeight: '49px !important',
```
LiKang6688, line 24: @haxxmaxx  I do not really want to use` !important`. But I do not understand that our own custom theme is overridden by the default theme.

PR #419, opened 2022-01-11
chore: fix package version

LiKang6688:
Removing the caret(ˆ) from the version of the packages, as we don’t know whether the new updates might bring breaking changes or not.
If there are vulnerabilities to the version we are using, we need to manually update that.


haxxmaxx:
Why? :)

haxxmaxx:
Are other repos doing this?

PR #420, opened 2022-01-11
chore: run `yarn spec` on npm version

haxxmaxx:
Based on @cbt1's changes for mekko, `yarn spec` is added to run at the same time as version. Build  runs before to make sure the build step wont fail on Circle. Also extending the instructions a bit.

### DIFF
--- a/.github/CONTRIBUTION.md
+++ b/.github/CONTRIBUTION.md
@@ -23,7 +23,8 @@ Run lint with:
 
 Currently only admins are able to publish.
 
-1. Check out master and make sure it is up to date
-2. Run `npm version minor/patch`
-3. Verify changes in `package.json`, commit and run `git push && git push --tags`
-4. Make sure all checks pass, then Circle automatically publishes to NPM.
+1. Check out master and run `git pull`.
+1. Run `git clean -dfx && yarn` to make sure depenencies are up-to-date.
+1. Run `npm version [major | minor | patch] -m "chore(release): v%s"`. Use semver string based on conventional commits since last release. Ex: `npm version patch -m "chore(release): v%s"`.
+1. Run `git push && git push --tags` to push commit and tag.
+1. Make sure all checks pass, then Circle CI automatically publishes to NPM.
--- a/.github/CONTRIBUTION.md
+++ b/.github/CONTRIBUTION.md
@@ -19,12 +19,25 @@ Run lint with:
 
 `yarn lint`
 
-## Publishing
+## Releasing
 
-Currently only admins are able to publish.
+Currently only admins are able to create a release. A release consists of the following:
+
+- Bumping the package version based on your commits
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTION.md
```diff
@@ -19,11 +19,25 @@ Run lint with:
...
+- Bumping the package version based on your commits
+- Updating the API specification
+- Creating a new commit with the changed files
+- Creating a tag with the new version
+- Pushing the release commit and tag to master
+
+### Step-By-Step
+
+1. Check out master and run `git pull`.
+1. Run `git clean -dfx && yarn` to make sure depenencies are up-to-date.
```
cbt1, line 35: Looks like the numbers are not getting incremented

cbt1 (approved):
Fix the number increment and you are good to go

.github/CONTRIBUTION.md
```diff
@@ -19,11 +19,25 @@ Run lint with:
...
+1. Run `npm version [major | minor | patch] -m "chore(release): v%s"`. Use semver string based on conventional commits since last release. Ex: `npm version patch -m "chore(release): v%s"`.
+1. Run `git push && git push --tags` to push commit and tag.
+1. Make sure all checks pass, then Circle CI automatically publishes to NPM.
+
+### On version command failure
```
LiKang6688, line 40: I do not understand this section

.github/CONTRIBUTION.md
haxxmaxx: No need to, that is a way of creating a numbered list with markdown. It is easier if you would add a step inbetween already existing steps for example.

.github/CONTRIBUTION.md
cbt1: Actually good catch. This section doesn't make sense in the context of failed npm version command. You should not end up in a scenario where a tag is created if `preversion` or `version` scripts fails.

But it should perhaps be updated to include instructions on how to revert `npm version` if one is for some reason not happy with the end result.

.github/CONTRIBUTION.md
cbt1: Cool! Did not know 👍

.github/CONTRIBUTION.md
haxxmaxx: For example, if a version number already exists on your machine, npm version will fail. Then you have to remove it and try again

LiKang6688: I see. Thanks. Could you also add those explanations into the docs?

haxxmaxx: OK I can clarify a bit.

haxxmaxx: fixed

LiKang6688 (approved):

PR #426, opened 2022-01-22
refactor: support style for right-to-left read direction

LiKang6688:
1. Following [mui recommended steps ](https://mui.com/guides/right-to-left/)to change the direction of reading as [never use CSS to apply the base direction.](https://www.w3.org/International/questions/qa-html-dir), using stylist-plugin-rtl with emotion. And the side effect: Two arrows for two dropdowns displayed on the wrong side of the text in RTL are fixed

2. Applying emotion with styled to replace makestyles

3. Learning from [RTL Styling 101](https://rtlstyling.com/posts/rtl-styling), some icons are universal, and it doesn't require us to flip them.  So the layout and direction of arrows do not need to change. But the order and action of arrows need to change: In RTL, the first button from the right is the first page, the second button is the previous page. Following https://github.com/mui-org/material-ui/blob/master/packages/mui-material/src/TablePagination/TablePaginationActions.js#L65 to do the fix

Before:
Screenshot 2022-01-22 at 19 30 53

After:
Screenshot 2022-01-24 at 22 28 08


src/locale/locales/en.json
```diff
@@ -20,19 +20,19 @@
     "comment": "Label for the dropdown to select a page."
   },
   "SNTable.Pagination.FirstPage": {
-    "value": "First page",
+    "value": "Go to first page",
```
haxxmaxx, line 23: If you change these you need to get a new approval from @tklkan

src/mui-setup.js
```diff
@@ -20,8 +20,12 @@ export default function muiSetup() {
       outline: 'none',
     };
   }
+  const themeConfig = {
+    ...adaptV4Theme(muiConfig),
+    ...{ direction },
```
haxxmaxx, line 25:
```suggestion
    direction,
```

src/table/components/TablePaginationActions.jsx
```diff
@@ -8,61 +9,50 @@ import FirstPageIcon from '@mui/icons-material/FirstPage';
...
+});
+
+const CustomFormControl = styled(FormControl)({
+  flexDirection: 'row',
+});
+
 export default function TablePaginationActions(props) {
-  const classes = useStyles();
-  const { page, lastPageIdx, onPageChange, keyboard, tableWidth, translator, isInSelectionMode } = props;
+  const { direction, page, lastPageIdx, onPageChange, keyboard, tableWidth, translator, isInSelectionMode } = props;
```
haxxmaxx, line 54: destruct in row above for consistency

haxxmaxx:
> After: Screenshot 2022-01-24 at 22 28 08

Evertyhing in the footer is strange except for the buttons

LiKang6688:
> > After: Screenshot 2022-01-24 at 22 28 08
> 
> Evertyhing in the footer is strange except for the buttons

I thought the same initially. 
But only 'of 132 100-2' is wrong. The error is from MUI itself. I will create a fix-PR for MUI.
The rest in the footer is correct.

src/table/components/TablePaginationActions.jsx
```diff
@@ -72,44 +62,54 @@ export default function TablePaginationActions(props) {
...
+      if (type === 'FirstPage') {
+        iconType = 'LastPage';
+      } else if (type === 'PreviousPage') {
+        iconType = 'NextPage';
+      } else if (type === 'NextPage') {
+        iconType = 'PreviousPage';
+      } else if (type === 'LastPage') {
+        iconType = 'FirstPage';
+      }
+    }
```
haxxmaxx, line 76: I would put this `icons` object. So you add `FirstPageRTL: LastPageIcon` and so on. The when you something like
```
iconType = `${type}${isRTL ? 'RTL' : ''}`
```

src/table/components/TableWrapper.jsx
```diff
@@ -15,33 +15,30 @@ import { updateFocus, handleResetFocus, handleFocusoutEvent } from '../utils/han
...
+  justifyContent: 'flex-end',
+});
+
+const TableContainerSection = styled(TableContainer)(({ containerMode }) => ({
+  height: containerMode ? '100%' : 'calc(100% - 52px)',
+  overflow: containerMode ? 'hidden' : 'auto',
+}));
+
+const CustomTablePagination = styled(TablePagination)(({ paginationHidden }) => ({
+  display: paginationHidden && 'none',
```
haxxmaxx, line 35: does this update when going back and forth between analysis/edit?

src/table/components/TableWrapper.jsx
```diff
@@ -147,9 +141,9 @@ export default function TableWrapper(props) {
       }
     >
       <AnnounceElements />
-      <TableContainer
+      <TableContainerSection
         ref={tableContainerRef}
-        className={classes[containerMode]}
+        containerMode={constraints.active}
```
haxxmaxx, line 146: Ah now I see how it updates, nice

src/table/components/TableWrapper.jsx
LiKang6688: yes, it does. I think that we have a unit test covering it. I gonna double check.

haxxmaxx:
> I thought the same initially. But only 'of 132 100-2' is wrong. The error is from MUI itself. I will create a fix-PR for MUI. The rest in the footer is correct.
> 
> BTW, in mui.com, we can set rtl in settings to check the looking of the table.

You're absolutely right, it just looks wrong in english 🙃

src/table/components/TablePaginationActions.jsx
haxxmaxx: isRTL is only used here so no need to create the constant

src/mui-setup.js
```diff
@@ -20,8 +20,12 @@ export default function muiSetup() {
       outline: 'none',
     };
   }
+  const themeConfig = {
+    ...adaptV4Theme(muiConfig),
```
haxxmaxx, line 24: Do we still need this? I thought it was to still be able to use makeStyles. Is it because muiConfig still is on mui v4 "format"?

LiKang6688: You are right. I do not think I am capable to update the format of muiConfig to use V5 theme yet. Side effects could happen that I am not aware of. 
I will work on it in the future on a separate branch.

tklkan (approved):

src/table/components/TablePaginationActions.jsx
LiKang6688: I think I understand what you mean now. Sure. Done.

src/table/components/TablePaginationActions.jsx
```diff
@@ -119,18 +119,25 @@ export default function TablePaginationActions(props) {
...
       {showFirstLast && getButton(onFirstPage, 0, 'FirstPage')}
       {getButton(onFirstPage, page - 1, 'PreviousPage')}
-      {getButton(onLastPage, page + 1, 'NextPage', !showFirstLast ? handleLastButtonTab : null)}
+      {getButton(
+        onLastPage,
+        page + 1,
+        'NextPage',
+
+        !showFirstLast ? handleLastButtonTab : null
+      )}
```
haxxmaxx, line 133: This doesn't look right

haxxmaxx (approved):

src/table/components/TablePaginationActions.jsx
LiKang6688: the format does not work in this case. Fixed. Thanks.

LiKang6688:
https://user-images.githubusercontent.com/4471489/151516781-6665596e-b753-436a-a0d0-4868bcec6a1e.mov

In this PR, using emotion as style engine (By default, MUI components come with emotion as their style engine. ) and creating a new cache instance that uses the stylis-plugin-rtl.

It causes an issue that CSS style will be duplicated as shown in the video.

haxxmaxx:
> It causes an issue that CSS style will be duplicated as shown in the video.

Hmm, I thought that it wouldn't do that anymore... Do you know if it is the same on main? I've seen the problem before, but maybe we have solved it for the table, don't rembember

LiKang6688:
> > It causes an issue that CSS style will be duplicated as shown in the video.
> 
> Hmm, I thought that it wouldn't do that anymore... Do you know if it is the same on main? I've seen the problem before, but maybe we have solved it for the table, don't rembember

It is caused by 
```
const cacheRtl = createCache({
  key: 'muirtl',
  stylisPlugins: [rtlPlugin],
});
<CacheProvider value={cacheRtl}></CacheProvider>
```

A new PR will come, which uses styled-components to avoid the issue.

LiKang6688:
> > > It causes an issue that CSS style will be duplicated as shown in the video.
> > 
> > 
> > Hmm, I thought that it wouldn't do that anymore... Do you know if it is the same on main? I've seen the problem before, but maybe we have solved it for the table, don't rembember
> 
> It is caused by
> 
> ```
> const cacheRtl = createCache({
>   key: 'muirtl',
>   stylisPlugins: [rtlPlugin],
> });
> <CacheProvider value={cacheRtl}></CacheProvider>
> ```
> 
> A new PR will come, which uses styled-components instead of emotion to avoid the issue.

haxxmaxx:
> > > It causes an issue that CSS style will be duplicated as shown in the video.
> > 
> > 
> > Hmm, I thought that it wouldn't do that anymore... Do you know if it is the same on main? I've seen the problem before, but maybe we have solved it for the table, don't remember
> 
> On the main branch, there is no such issue.
> 
> It is just caused by
> 
> ```
> const cacheRtl = createCache({
>   key: 'muirtl',
>   stylisPlugins: [rtlPlugin],
> });
> <CacheProvider value={cacheRtl}></CacheProvider>
> ```
> 
> A new PR will come, which uses styled-components instead of emotion to avoid the issue.

Ok so without cacheRtl it wouldn't be a problem? We are considering to use it in other projects
Go ahead and close this one then :)

LiKang6688:
> > > > It causes an issue that CSS style will be duplicated as shown in the video.
> > > 
> > > 
> > > Hmm, I thought that it wouldn't do that anymore... Do you know if it is the same on main? I've seen the problem before, but maybe we have solved it for the table, don't remember
> > 
> > 
> > On the main branch, there is no such issue.
> > It is just caused by
> > ```
> > const cacheRtl = createCache({
> >   key: 'muirtl',
> >   stylisPlugins: [rtlPlugin],
> > });
> > <CacheProvider value={cacheRtl}></CacheProvider>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > A new PR will come, which uses styled-components instead of emotion to avoid the issue.
> 
> Ok so without cacheRtl it wouldn't be a problem? We are considering to use it in other projects Go ahead and close this one then :)

Right, emotion does not cause such an issue. The problem comes from cacheRtl.

PR #427, opened 2022-01-22
fix: unable to scroll horizontally in right-to-left direction

LiKang6688:
need to consider scroll horizontally in right-to-left direction as well

### DIFF
--- a/src/table/components/TableWrapper.jsx
+++ b/src/table/components/TableWrapper.jsx
@@ -12,7 +12,7 @@ import TablePaginationActions from './TablePaginationActions';
 import useDidUpdateEffect from './useDidUpdateEffect';
 import { handleTableWrapperKeyDown } from '../utils/handle-key-press';
 import { updateFocus, handleResetFocus, handleFocusoutEvent } from '../utils/handle-accessibility';
-import { handleScroll, handleNavigateTop } from '../utils/handle-scroll';
+import { handleHorizontalScroll, handleNavigateTop } from '../utils/handle-scroll';
 import announcementFactory from '../utils/announcement-factory';
 
 const useStyles = makeStyles({
@@ -85,20 +85,28 @@ export default function TableWrapper(props) {
   };
 
   useEffect(() => {
-    const memoedContainer = tableContainerRef.current;
     const memoedWrapper = tableWrapperRef.current;
-    if (!memoedContainer || !memoedWrapper) return () => {};
+    if (!memoedWrapper) return () => {};
 
-    const scrollCallback = (evt) => handleScroll(evt, tableContainerRef);
     const focusOutCallback = (evt) => handleFocusoutEvent(evt, shouldRefocus, keyboard);
-    memoedContainer.addEventListener('wheel', scrollCallback);
     memoedWrapper.addEventListener('focusout', focusOutCallback);
 
     return () => {
-      memoedContainer.removeEventListener('wheel', scrollCallback);
       memoedWrapper.removeEventListener('focusout', focusOutCallback);
     };
   }, []);
+
+  useEffect(() => {
+    const memoedContainer = tableContainerRef.current;
+    if (!memoedContainer) return () => {};
+
+    const horizontalScrollCallback = (evt) => handleHorizontalScroll(evt, rtl, memoedContainer);
+    memoedContainer.addEventListener('wheel', horizontalScrollCallback);
+
+    return () => {
+      memoedContainer.removeEventListener('wheel', horizontalScrollCallback);
+    };
+  }, [rtl]);
 
   useEffect(
     () => handleNavigateTop({ tableContainerRef, focusedCellCoord, rootElement }),
--- a/src/table/utils/handle-scroll.js
+++ b/src/table/utils/handle-scroll.js
@@ -1,7 +1,7 @@
 export const handleHorizontalScroll = (evt, rtl, memoedContainer) => {
   evt.stopPropagation();
-  // scrollWidth is a measurement of the width of an element's content, including content not visible on the screen due to overflow.
-  // offsetWidth is a measurement in pixels of the element's CSS width, including any borders, padding, and vertical scrollbars
+  // scrollWidth is the width of an element's content, including content not visible on the screen due to overflow.
+  // offsetWidth is the element's CSS width, including any borders, padding, and vertical scrollbars
   // scrollLeft is the number of pixels scrolled from its left edge
   if (rtl) {
     // scrollLeft is 0 when the scrollbar is at its rightmost position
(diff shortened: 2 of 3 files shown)


src/table/components/TableWrapper.jsx
```diff
@@ -85,21 +85,30 @@ export default function TableWrapper(props) {
...
     memoedWrapper.addEventListener('focusout', focusOutCallback);
 
     return () => {
-      memoedContainer.removeEventListener('wheel', scrollCallback);
       memoedWrapper.removeEventListener('focusout', focusOutCallback);
     };
   }, []);
 
+  useEffect(() => {
+    window.console.log({ rtl });
```
haxxmaxx, line 100: remove

LiKang6688: thanks. Done

src/table/utils/handle-scroll.js
```diff
@@ -1,11 +1,21 @@
...
-  if (max > 0 && (newScrollLeft < 0 || newScrollLeft > max)) {
-    evt.preventDefault();
-    scrollLeft = Math.max(0, Math.min(max, newScrollLeft));
+  if (rtl) {
+    const scrollLeft = memoedContainer.scrollLeft + evt.deltaX;
+    let scrollRight = memoedContainer.scrollWidth - (scrollLeft + memoedContainer.clientWidth);
+    const max = memoedContainer.scrollWidth + memoedContainer.offsetWidth;
+    if (max > 0 && (scrollRight < 0 || scrollRight > max)) {
+      evt.preventDefault();
+      scrollRight = Math.max(0, Math.min(max, scrollRight));
```
haxxmaxx, line 9: How is this applied to anything? `scrollRight` isn't a property on `memoedContainer` or?

LiKang6688: Element only has [scrollLeft](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLefts) unfortunately. 
It is only applied in RTL. When the length of scrollRight is larger than the available scroll length (which is the `max` variable), preventDefault is run to prevent going to the previous page.

haxxmaxx: yeah but what happens when you actually set `scrollRight` in here, it is a local variable, then nothing is returned

LiKang6688: yes, good point. Something wrong here. It is weird that it worked before. let me fix it.

src/table/utils/handle-scroll.js
```diff
@@ -1,11 +1,23 @@
...
-  let { scrollLeft } = tableContainerRef.current;
-  const newScrollLeft = scrollLeft + evt.deltaX;
-  const max = tableContainerRef.current.scrollWidth - tableContainerRef.current.offsetWidth;
-  if (max > 0 && (newScrollLeft < 0 || newScrollLeft > max)) {
-    evt.preventDefault();
-    scrollLeft = Math.max(0, Math.min(max, newScrollLeft));
+  if (rtl) {
+    // in RTL, scrollLeft is actually scrollRight, scrollRight is scrollLeft
+    let { scrollLeft } = memoedContainer;
+    const newScrollLeft = scrollLeft + evt.deltaX;
```
haxxmaxx, line 6: break out these two lines, identitcal to else. So is `max` constant

src/table/utils/handle-scroll.js
```diff
@@ -1,11 +1,30 @@
...
+  // offsetWidth is a measurement in pixels of the element's CSS width, including any borders, padding, and vertical scrollbars
+  // scrollLeft is the number of pixels scrolled from its left edge
+  if (rtl) {
+    // scrollLeft is 0 when the scrollbar is at its rightmost position
+    // (at the start of the scrolled content),
+    // and then increasingly negative as you scroll towards (left) the end of the content .
+    // evt.deltaX increasingly negative as you scroll towards left,
+    // increasingly positive as you scroll towards right
+    let { scrollLeft } = memoedContainer;
+    const ScrollLeftWidth = scrollLeft + evt.deltaX;
```
haxxmaxx, line 13: These rows are still in both the if and the else. If you move the desctructuring to the commentes on lines 3-5 it makes more sense

src/table/utils/handle-scroll.js
```diff
@@ -1,11 +1,30 @@
...
   evt.stopPropagation();
-  let { scrollLeft } = tableContainerRef.current;
-  const newScrollLeft = scrollLeft + evt.deltaX;
-  const max = tableContainerRef.current.scrollWidth - tableContainerRef.current.offsetWidth;
-  if (max > 0 && (newScrollLeft < 0 || newScrollLeft > max)) {
-    evt.preventDefault();
-    scrollLeft = Math.max(0, Math.min(max, newScrollLeft));
+  // scrollWidth is a measurement of the width of an element's content, including content not visible on the screen due to overflow.
+  // offsetWidth is a measurement in pixels of the element's CSS width, including any borders, padding, and vertical scrollbars
+  // scrollLeft is the number of pixels scrolled from its left edge
```
haxxmaxx, line 5:
```suggestion
  // scrollWidth is the width of an element's content, including content not visible on the screen due to overflow.
  // offsetWidth is the element's CSS width, including any borders, padding, and vertical scrollbars
  // scrollLeft is the number of pixels scrolled from its left edge
```

src/table/utils/handle-scroll.js
LiKang6688: good point. Updated.

src/table/utils/handle-scroll.js
```diff
@@ -1,11 +1,25 @@
...
+export const handleHorizontalScroll = (evt, rtl, memoedContainer) => {
   evt.stopPropagation();
-  let { scrollLeft } = tableContainerRef.current;
-  const newScrollLeft = scrollLeft + evt.deltaX;
-  const max = tableContainerRef.current.scrollWidth - tableContainerRef.current.offsetWidth;
-  if (max > 0 && (newScrollLeft < 0 || newScrollLeft > max)) {
+  // scrollWidth is the width of an element's content, including content not visible on the screen due to overflow.
+  // offsetWidth is the element's CSS width, including any borders, padding, and vertical scrollbars
+  // scrollLeft is the number of pixels scrolled from its left edge
+  let { scrollLeft } = memoedContainer;
```
haxxmaxx, line 6: You can destruct `scrollWidth` and `offsetWidth` too if you want

haxxmaxx (approved):

src/table/utils/handle-scroll.js
LiKang6688: yes, I want that too. But eslint complaints that scrollWidth and offsetWidth should be const not let.

haxxmaxx: ah ok, then leave it

PR #428, opened 2022-01-26
fix: selected cell font color

LiKang6688:
Before:
Screenshot 2022-01-26 at 21 49 37

After:
Screenshot 2022-01-26 at 21 50 47


haxxmaxx:
I'm not sure about this. I consider always using black font color when selecting an improvement. That column effectively turns into a listbox, and should override the table colors.

If we can get UX eyes on this and they think this is better, then we should go for it. If not, it's better to wait until we get some proper designs, so we don't have to redo this later

LiKang6688:
> I'm not sure about this. I consider always using black font color when selecting an improvement. That column effectively turns into a listbox, and should override the table colors.
> 
> If we can get UX eyes on this and they think this is better, then we should go for it. If not, it's better to wait until we get some proper designs, so we don't have to redo this later

I agree with you

haxxmaxx:
Closing this since we are not doing it now. Keep the branch if we decide to do it later

PR #430, opened 2022-01-28
refactor: rename and fix warnings

LiKang6688:
Screenshot 2022-01-28 at 10 04 18

1. React does not like capital letters. Updating all props names into lowercase
2. Renaming `selDispatch` to `selectionDispatch`
3. `tableData.columns` to be `columns`
4. `onPageChange` is required in `TablePagination`

Screenshot 2022-01-28 at 10 07 05

5. Adding unique key for `<option>`

No solution for the last warning yet
Screenshot 2022-01-28 at 10 25 14

More to read:
https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html#changes-in-detail

### DIFF
--- a/src/table/components/withSelections.jsx
+++ b/src/table/components/withSelections.jsx
@@ -5,9 +5,9 @@ import { getSelectionStyle } from '../utils/styling-utils';
 
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
-    const { selectionState, cell, selDispatch, styling, announce } = props;
+    const { selectionstate: selectionState, cell, selectiondispatch: selectionDispatch, styling, announce } = props;
     const handleMouseUp = (evt) =>
-      cell.isDim && evt.button === 0 && selectCell({ selectionState, cell, selDispatch, evt, announce });
+      cell.isDim && evt.button === 0 && selectCell({ selectionState, cell, selectionDispatch, evt, announce });
     const selectionStyling = useMemo(
       () => getSelectionStyle(styling, cell, selectionState),
       [styling, cell, selectionState]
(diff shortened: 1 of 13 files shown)


src/table/components/withSelections.jsx
```diff
@@ -5,9 +5,9 @@ import { getSelectionStyle } from '../utils/styling-utils';
 
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
-    const { selectionState, cell, selDispatch, styling, announce } = props;
+    const { selectionstate: selectionState, cell, selectiondispatch: selectionDispatch, styling, announce } = props;
```
haxxmaxx, line 8: what, react complains when you use camel case??

LiKang6688: Not here, react complains there https://github.com/qlik-oss/sn-table/pull/430/files#diff-0d7173fd8da1e342bacca6f203bd521b0300c9ab25684eb6817b7e3d4be57a0cL90

haxxmaxx: Is there a good reason for it? I just think that it makes the variable names harder to read. What is the name of the rule?

LiKang6688: Screenshot 2022-01-28 at 12 07 13
https://reactjs.org/warnings/unknown-prop.html The fouth point. It is due to how react is working

LiKang6688: https://reactjs.org/warnings/unknown-prop.html

LiKang6688: the second point.
Screenshot 2022-01-28 at 12 14 33
 https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
 You are right, I will do the update.

LiKang6688: no, i do not find a better way.  `data-selection-state `is not converted to `selectionState`
Screenshot 2022-01-28 at 12 18 24

haxxmaxx: oh ok I think i understand now. CellRenderer isn't a real react component, but a standard html component. Then I guess it should be lower case

haxxmaxx (approved):

PR #431, opened 2022-01-28
refactor: unnecessary tablecell props cleanup

LiKang6688:
Screenshot 2022-01-28 at 16 18 54

a better way to fix the react warnings about dom attributes

`selectionState`, `selDispatch`, `announce`, `cell`, `column`, and `value` are not needed to be attached in the dom.

### DIFF
--- a/src/table/components/withSelections.jsx
+++ b/src/table/components/withSelections.jsx
@@ -9,6 +9,7 @@ export default function withSelections(CellComponent) {
 
     const handleMouseUp = (evt) =>
       cell.isDim && evt.button === 0 && selectCell({ selectionState, cell, selectionDispatch, evt, announce });
+
     const selectionStyling = useMemo(
       () => getSelectionStyle(styling, cell, selectionState),
       [styling, cell, selectionState]
--- a/src/table/components/withSelections.jsx
+++ b/src/table/components/withSelections.jsx
@@ -5,7 +5,8 @@ import { getSelectionStyle } from '../utils/styling-utils';
 
 export default function withSelections(CellComponent) {
   const HOC = (props) => {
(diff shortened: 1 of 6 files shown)


src/table/components/withSelections.jsx
```diff
@@ -5,23 +5,30 @@ import { getSelectionStyle } from '../utils/styling-utils';
...
     const selectionStyling = useMemo(
       () => getSelectionStyle(styling, cell, selectionState),
       [styling, cell, selectionState]
     );
 
-    return <CellComponent {...props} styling={selectionStyling} onMouseUp={handleMouseUp} />;
+    return <CellComponent {...passThroughProps} styling={selectionStyling} onMouseUp={handleMouseUp} />;
+  };
+
+  HOC.defaultProps = {
```
haxxmaxx, line 21: Why does only this HOC need defualtProps?

LiKang6688: Only this HOC has not required propTypes property `column`

haxxmaxx: I don't understand, we don't use it here, so why destruct it from `props`? Can't we just pass it through?

LiKang6688: The reason why we want to destruct it from props is that 
we do not want to pass the `column` to `withStyling`, as `column` is not needed in the DOM.
And When `hasColumnStyling` is true, `withColumnStyling` will not pass `column` to `withSelections`, 
so `column` could be null and should not be required.

haxxmaxx: that makes sense

haxxmaxx (approved):

PR #434, opened 2022-01-29
refactor: support right-to-left direction using styled-components as style engine

LiKang6688:


haxxmaxx:
ok so this is only to make rtl work? We are not replacing `makeStyles` here? Using this approach, can we still switch out `makeStyles`?

LiKang6688:
> ok so this is only to make rtl work? We are not replacing `makeStyles` here? Using this approach, can we still switch out `makeStyles`?

Yes, This PR only makes RTL work. I plan to replace `makestyles` and v4 theme together. Yes, we can.

haxxmaxx (approved):

PR #441, opened 2022-02-11
Feat: allow padding override of cells

skokenes:
Allow developers to provide custom padding for table headers and cells by overriding baseStyling in the same way color and fontSize can be overridden.

### DIFF
--- a/src/table/utils/styling-utils.js
+++ b/src/table/utils/styling-utils.js
@@ -36,7 +36,7 @@ export const getAutoFontColor = (backgroundColor) =>
 export const getBaseStyling = (styleObj, theme) => {
   let padding = STYLING_DEFAULTS.PADDING;
   if (styleObj.padding) {
-    padding = styleObj.padding;
+    ({ padding } = styleObj);
   } else if (styleObj.fontSize) {
     padding = `${styleObj.fontSize / 2}px ${styleObj.fontSize}px`;
   }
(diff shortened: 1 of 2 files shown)


LiKang6688 (approved):

src/table/utils/styling-utils.js
```diff
@@ -33,11 +43,13 @@ export function getColor(color = {}, defaultColor, theme) {
...
-  fontSize: styleObj.fontSize || STYLING_DEFAULTS.FONT_SIZE,
-  padding: styleObj.fontSize ? `${styleObj.fontSize / 2}px ${styleObj.fontSize}px` : STYLING_DEFAULTS.PADDING,
-});
+export const getBaseStyling = (styleObj, theme) => {
+  return {
+    color: getColor(styleObj.fontColor, STYLING_DEFAULTS.FONT_COLOR, theme),
+    fontSize: styleObj.fontSize || STYLING_DEFAULTS.FONT_SIZE,
+    padding: getPadding(styleObj, STYLING_DEFAULTS.PADDING),
+  };
+};
```
haxxmaxx, line 52: Why using a `return` when it is an arrow function?

haxxmaxx: Good question. Leave it for now, not sure why it is disabled.

haxxmaxx (approved):

src/table/utils/styling-utils.js
LiKang6688: My bad. I forgot to delete it before. A new fix PR will come.

haxxmaxx: ok then let's not change it

niekvanstaveren (commented):
How will this be supported? Only through options when rendering, or also through the Qlik Sense Theme?

PR #442, opened 2022-02-14
feat: apply theme from sense

LiKang6688:
1. Add themes into nebula.nebula.config.js, when running locally, in nebula dev envionment, we can select the theme 'Test' for testing purposes.
2. Refactor mui custom theme, converting string to number.
3. Make the border of table look perfect
4. Apply table header cell font color, size, padding and body cell font size, color, padding from sense theme 
Note:
the order of style
default (inl. sprout theme) < nebula theme < Sense theme < styling settings < column < selection (except the selected green) < hover < selected green

5. Apply table background color from sense theme
Note:
- when the table background color from sense theme is transparent, there is a default background color for the header 
to avoid seeing the table body through the header
- When a CSS property is not specified in the theme, like the color->header->straightTable is not specified, will check color->straghtTable->object, undefined, then color->->object, still undefined, will look for default theme - [nebula light](https://github.com/qlik-oss/nebula.js/blob/master/apis/theme/src/themes/light.json) and do the same.

Running in nebula dev environment (locally), the style from theme can be found in nebula.config.js -> serve -> themes.

Running as an extension in Qlik Sense, the theme is from there.  https://help.qlik.com/en-US/sense-developer/February2022/Subsystems/Extensions/Content/Sense_Extensions/CustomThemes/custom-themes-properties.htm

The default theme is from nebula theme: https://github.com/qlik-oss/nebula.js/tree/master/apis/theme/src/themes

In the next PRs:
#455  When the table background color from the sense theme is transparent, there is a header background color depending on the header font color to avoid seeing the table body through the table head.

#459 When you set the color of header content, the color of the sort label should be the same as that.
          When there is no header content color setting, the sort label color is kind of white in the dark background color and black in the light background color.

#446 Adjust the TablePagination font color based on the background color

#451 There is the border color depending on the background color

#453 Apply the body content hover background color and color from the theme


src/table/utils/styling-utils.js
```diff
@@ -25,58 +27,79 @@ export const SELECTION_STYLING = {
   },
 };
 
-export function getColor(color = {}, defaultColor, theme) {
+export function getColor(defaultColor, theme, color = {}) {
```
niekvanstaveren, line 30: Why change this order?

niekvanstaveren (commented):
There should probably be a default background color for the header. If we do not get anything from the theme (or an object override in the future) apply the background that exists now

nebula.config.js
```diff
@@ -12,5 +12,25 @@ module.exports = {
...
     keyboardNavigation: false,
+    themes: [
+      {
+        id: 'Test',
+        theme: {
+          straightTable: {
+            header: {
+              fontSize: '40px',
+              color: 'cyan',
+              padding: '10px',
```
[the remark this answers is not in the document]
LiKang6688: Speros team wants to have the option https://github.com/qlik-oss/sn-table/pull/441. They found that the default padding is too big.
As far as I know, it will only work for sn-table not straight-table.

src/index.js
```diff
@@ -40,6 +40,7 @@ export default function supernova(env) {
       const translator = useTranslator();
       const selectionsAPI = useSelections();
       const tableTheme = muiSetup(direction);
+      // the style which is from them can be found in nebula.config.js -> serve -> themes
```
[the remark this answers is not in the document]
LiKang6688: Yes, only for the nebula dev environment.

LiKang6688:
> There should probably be a default background color for the header. If we do not get anything from the theme (or an object override in the future) apply the background that exists now

I do not think it is a good idea to have a default background color for the header. When the default background color does not match the sheet color, it will look odd.

niekvanstaveren:
> > There should probably be a default background color for the header. If we do not get anything from the theme (or an object override in the future) apply the background that exists now
> 
> I do not think it is a good idea to have a default background color for the header. When the default background color does not match the sheet color, it will look odd.

Not as bad as seeing the table body through the header though....

LiKang6688:
> > > There should probably be a default background color for the header. If we do not get anything from the theme (or an object override in the future) apply the background that exists now
> > 
> > 
> > I do not think it is a good idea to have a default background color for the header. When the default background color does not match the sheet color, it will look odd.
> 
> Not as bad as seeing the table body through the header though....

I agree that. 
But I really want to have a better outcome. 
Though using the class selector to pick the color of the sheet is not good, it brings the outcome that we want...

nebula.config.js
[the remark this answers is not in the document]
niekvanstaveren: Yes, but that is through options not through the theme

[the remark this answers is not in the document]
LiKang6688: Right, as far as I know, they only need to have the padding from Properties, not from the Sense Theme. 
But at the same time, I see that it is a bonus to add more support for the theme and no harm.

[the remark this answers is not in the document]
niekvanstaveren: Except that everything in the Theme should also be supported in the old table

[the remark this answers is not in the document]
LiKang6688: Alright, we should not give users wrong expectations, I gonna remove them

nebula.config.js
```diff
@@ -12,5 +12,43 @@ module.exports = {
...
+        theme: {
+          straightTable: {
+            header: {
+              fontSize: '20px',
+              color: 'blue',
+              fontFamily: 'sans-serif',
+            },
+            content: {
+              fontSize: '16px',
+              color: '232323',
```
haxxmaxx, line 45: missing #?

src/index.js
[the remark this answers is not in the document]
haxxmaxx: then add that to the comment, only for dev env

src/table/components/TableHeadWrapper.jsx
```diff
@@ -34,15 +34,21 @@ function TableHeadWrapper({
...
 }) {
-  const headCellStyle = useMemo(() => getHeadStyle(layout, theme), [layout, theme.name()]);
-  const tableCellStyle = {
-    color: headCellStyle.color,
-    fontSize: headCellStyle.fontSize,
-    padding: headCellStyle.padding,
+  const tableBackgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
+  // When the table background color from sense theme is transparent, there is a default background color for the header
+  // to avoid seeing the table body through the header
+  const backgroundColor = tableBackgroundColor === 'transparent' ? '#FAFAFA' : tableBackgroundColor;
```
haxxmaxx, line 40: This should be done inside getHeadStyle IMO. Then you can create `headStyle` after that

haxxmaxx: actually would be nice if getHeadStyle returned both cell and head style

nebula.config.js
LiKang6688: Yes, added

src/table/components/TableHeadWrapper.jsx
LiKang6688: Okay, I will work on these in #454

haxxmaxx: Or just do it properly from the start? Since you are planning on merging this first?

src/table/utils/styling-utils.js
```diff
@@ -25,58 +27,78 @@ export const SELECTION_STYLING = {
...
+
+  const baseStyle = {
+    backgroundColor,
+    fontFamily,
+    color: styleObj?.fontColor ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj?.fontColor) : color,
+    fontSize: styleObj?.fontSize || fontSize,
+    padding: styleObj?.fontSize && `${styleObj?.fontSize / 2}px ${styleObj?.fontSize}px`,
+  };
+  // Remove all Undefined Values from an Object
+  Object.keys(baseStyle).forEach((key) => baseStyle[key] == null && delete baseStyle[key]);
```
haxxmaxx, line 52: use ===

src/table/utils/styling-utils.js
```diff
@@ -25,58 +27,78 @@ export const SELECTION_STYLING = {
...
+export const getBaseStyling = (styleObj, objetName, theme) => {
+  const backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
+  const fontFamily = theme.getStyle('object', `straightTable.${objetName}`, 'fontFamily');
+  const color = theme.getStyle('object', `straightTable.${objetName}`, 'color');
+  const fontSize = theme.getStyle('object', `straightTable.${objetName}`, 'fontSize');
+
+  const baseStyle = {
+    backgroundColor,
+    fontFamily,
+    color: styleObj?.fontColor ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj?.fontColor) : color,
```
haxxmaxx, line 47: can styleObj be undefined?

src/table/components/TableHeadWrapper.jsx
LiKang6688: Sorry, I realized that merging this first is not a good idea. We should be planning on merging the rest.

src/table/utils/styling-utils.js
```diff
@@ -25,58 +27,78 @@ export const SELECTION_STYLING = {
...
+  // when hovering, the hover font color becomes white when the hover
+  // background color is a dark color or the hover font color stays
+  // the same as whatever the font color is when the hover background
+  // color is a light color.
   // 2. There is a hover font color but no hover background color,
-  // when hovering, only a hover font color is applied and the hover background color disappears.
-  // 3. There is no hover font color and no hover background color, when hovering, the defalut hover effect (light gray backgournd)
-  // 4. There are both hover font and background colors, when hovering, the hover font and background colors take effect.
+  // when hovering, only a hover font color is applied and the hover
+  // background color disappears.
```
haxxmaxx, line 76: maybe say that the hover background color is transparent. disappears sounds a bit magic

src/table/utils/styling-utils.js
```diff
@@ -25,58 +27,78 @@ export const SELECTION_STYLING = {
...
+  const unsetHoverFontBackgroundColor = isUnset(content?.hoverFontColor) && unsetHoverBackgroundColor;
 
-  const hoverBackgroundColor = unsetHoverFontandBackgroundColor
-    ? STYLING_DEFAULTS.HOVER_BACKGROUND
-    : unsetHoverBackgroundColor
+  const setHoverBackgroundColor = unsetHoverBackgroundColor
     ? ''
-    : getColor(content.hoverColor, STYLING_DEFAULTS.HOVER_BACKGROUND, theme);
-  const hoverFontColor = unsetHoverFontandBackgroundColor
+    : getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, content?.hoverColor);
```
haxxmaxx, line 87: don't we know that `content.hoverColor` exists, since we checked with isUnset?

src/table/utils/styling-utils.js
```diff
@@ -25,58 +27,78 @@ export const SELECTION_STYLING = {
...
 
-  const unsetHoverBackgroundColor = isUnset(content.hoverColor);
-  const unsetHoverFontandBackgroundColor = isUnset(content.hoverFontColor) && unsetHoverBackgroundColor;
+  const unsetHoverBackgroundColor = isUnset(content?.hoverColor);
+  const unsetHoverFontBackgroundColor = isUnset(content?.hoverFontColor) && unsetHoverBackgroundColor;
 
-  const hoverBackgroundColor = unsetHoverFontandBackgroundColor
-    ? STYLING_DEFAULTS.HOVER_BACKGROUND
-    : unsetHoverBackgroundColor
+  const setHoverBackgroundColor = unsetHoverBackgroundColor
```
haxxmaxx, line 85: This name is confusing. since `unset...` is a bool, I would think this one is a bool as well. `hoverBackgroundColor` works then override it, or?

src/table/utils/styling-utils.js
```diff
@@ -96,26 +118,26 @@ export function getColumnStyle(styling, qAttrExps, stylingInfo) {
...
   const { colIdx, rows, api } = selectionState;
 
   if (api.isModal()) {
-    if (colIdx !== cell.colIdx) {
-      return { background: `${STYLING_DEFAULTS.EXCLUDED_BACKGROUND}, ${background}` };
-    }
+    if (colIdx !== cell.colIdx)
+      return {
+        background: `${STYLING_DEFAULTS.EXCLUDED_BACKGROUND}, ${
+          background || tableBackgroundColor || STYLING_DEFAULTS.WHITE
```
haxxmaxx, line 128: Can both `background` and `tableBackgroundColor` be undefined? also, which one is from the theme?

src/table/utils/styling-utils.js
```diff
@@ -96,26 +118,26 @@ export function getColumnStyle(styling, qAttrExps, stylingInfo) {
...
+        }`,
+      };
 
-    for (let i = 0; i < rows.length; i++) {
-      if (rows[i].qElemNumber === cell.qElemNumber) {
-        return SELECTION_STYLING.SELECTED;
-      }
-    }
+    const match = (row) => row.qElemNumber === cell.qElemNumber;
+    if (rows.some(match)) return SELECTION_STYLING.SELECTED;
```
haxxmaxx, line 133: Is this as fast? I used for loops for speed here

src/table/utils/styling-utils.js
haxxmaxx: isn't better to have a default value for the parameter?

src/table/components/TableHeadWrapper.jsx
LiKang6688: Done in #455

src/table/utils/styling-utils.js
LiKang6688: I mean == here,  it is true if a is null or undefined

src/table/utils/styling-utils.js
LiKang6688: it can. layout.components?.[0]?.header can be undefined

src/table/utils/styling-utils.js
LiKang6688: in this case, the value of hoverBackgroundColor is ''. I would not say it is transparent. 
How about 'There is a hover font color but no hover background color, when hovering, only the font color is applied and background color is the same as before'

src/table/utils/styling-utils.js
LiKang6688: yes, the array method is faster than for loops. Did a quick search https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

src/table/utils/styling-utils.js
LiKang6688: > yes, we know that. So `content?.hoverColor` should be `content.hoverColor` here.

src/table/utils/styling-utils.js
haxxmaxx: Didn't know about that, kind of confusing syntax. Can you use `!baseStyle[key]` instead?

src/table/utils/styling-utils.js
haxxmaxx: same as what? the background color or the default hover background color? `There is a hover font color but no hover background color, when hovering, only the font color is applied and no hover background is shown` or similar

src/table/utils/styling-utils.js
haxxmaxx: awesome

src/index.js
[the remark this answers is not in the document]
LiKang6688: Done in #446

src/table/utils/styling-utils.js
LiKang6688: that's better. Thanks. Updated.

src/table/utils/styling-utils.js
LiKang6688: in the future, it is possible to set the padding to be 0, right?

src/table/utils/styling-utils.js
LiKang6688: `setHoverBackgroundColor` is confusing? I do not follow. What do you mean?

src/table/utils/styling-utils.js
haxxmaxx: true, that is falsey... Then stick with this

src/table/utils/styling-utils.js
haxxmaxx: yes, I would not expect that to be the actual hover beckground color. Either a setter, or a boolean that means that the color should be set

src/table/utils/styling-utils.js
LiKang6688: Updated in #453

src/table/utils/styling-utils.js
LiKang6688: Okay, updated in #453

src/table/utils/styling-utils.js
LiKang6688: Yes, they can.
background is from qAttributeExpressions in qDimensions or qMeasures
tableBackgroundColor is from the theme
I am adding the comments to clarify.
Yes, it is better `tableBackgroundColor = STYLING_DEFAULTS.WHITE`

haxxmaxx: maybe call them `columnBackground` and `themeBackground` so you know which one is which? I know I'm being picky about variable names but I think we need to be consistent when we add this extra layer of theming to the styling.

LiKang6688: happy to hear that. I am doing a similar refactor.

LiKang6688: But I realized that renaming should be in a new PR otherwise conflicts everywhere. Sorry for so many PRs.

src/table/utils/styling-utils.js
haxxmaxx: But now this is back...

src/table/utils/styling-utils.js
haxxmaxx: this is also back

LiKang6688: because #453 is not merged yet

src/table/utils/styling-utils.js
LiKang6688: https://github.com/qlik-oss/sn-table/pull/453 is not merged yet

src/table/utils/styling-utils.js
niekvanstaveren: I think you got to a wrong conclusion from that page. When I read it it states that for loops are almost always better performing than the array methods

niekvanstaveren (commented):
Please don't mix refactoring and functional PRs. It gets almost impossible to do proper reviews.

src/index.js
```diff
@@ -39,8 +40,20 @@ export default function supernova(env) {
...
+      // Running in nebula dev environment (locally),
+      // the style from theme can be found in
+      // nebula.config.js -> serve -> themes.
+      // Get the value of a style attribute, starting in the given base path + path
+      // Ex: Base path: "object", Path: "straightTable", Attribute: "backgroundColor"
+      // Will search in, and fall back to:
+      // object - straightTable - backgroundColor
+      // straightTable - backgroundColor
+      // object - backgroundColor
+      // backgroundColor
```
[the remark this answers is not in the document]
LiKang6688: not on the dev portal, only in the nebula code comments. I do not think it is good to attach a link URL to the code comments URL, so I added those comments which are important to know in the first place.

src/index.js
```diff
@@ -39,8 +40,20 @@ export default function supernova(env) {
...
+      theme.backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
+      theme.isBackgroundDarkColor = isDarkColor(theme.backgroundColor);
```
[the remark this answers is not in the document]
LiKang6688: Yes, as component() reruns on a theme change.
Yes, I feel so as we do not have a state management tool in place yet.

src/table/components/TablePaginationActions.jsx
```diff
@@ -39,6 +40,9 @@ export default function TablePaginationActions({
   const handleSelectPage = (event) => onPageChange(+event.target.value);
   const handleLastButtonTab = keyboard.enabled ? (event) => handleLastTab(event, isInSelectionMode) : null;
 
+  const disabledIconButtonColor = theme.isBackgroundDarkColor ? 'rgba(255, 255, 255, 0.3)' : 'rgba(0, 0, 0, 0.3)';
+  const iconButtonColor = theme.isBackgroundDarkColor ? 'rgba(255, 255, 255, 0.9)' : 'rgba(0, 0, 0, 0.54)';
+
```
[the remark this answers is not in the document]
LiKang6688: I got them from mui table example
Screenshot 2022-03-01 at 19 13 49

src/table/components/TableWrapper.jsx
```diff
@@ -179,24 +200,14 @@ export default function TableWrapper(props) {
       </TableContainer>
       <Paper sx={paperTablePaginationStyle}>
         <TablePagination
-          sx={constraints.active && { display: 'none' }}
+          sx={tablePaginationStyle}
```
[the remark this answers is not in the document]
LiKang6688: As we decided to simplify the component, otherwise the return () section will become super long
Screenshot 2022-03-01 at 19 15 00

src/table/utils/styling-utils.js
LiKang6688: You are right. I misunderstood it as the x-axis is ops/sec! As the performance is more important here, I gonna revert it.

src/locale/src/index.js
```diff
@@ -7,7 +7,7 @@ export default function registerLocale(translator) {
     const g = translator.get(t);
 
     // if the translated string is different from its id,
-    // we assume the translations already exist for current locale
+    // you assume the translations already exist for current locale
```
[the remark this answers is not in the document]
LiKang6688: Fixed. 'the translations are assumed already existed for the current locale'

src/table/utils/styling-utils.js
LiKang6688: Had a test
Screenshot 2022-03-01 at 19 43 27
Screenshot 2022-03-01 at 19 44 57
for loop is faster

src/locale/src/index.js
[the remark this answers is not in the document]
haxxmaxx: Not really grammatically correct, use this: the translations are assumed to already exist for the current locale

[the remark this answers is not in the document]
LiKang6688: Updated

src/index.js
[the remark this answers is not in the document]
haxxmaxx: Did we do anything here?

src/table/components/TablePaginationActions.jsx
```diff
@@ -51,31 +55,42 @@ export default function TablePaginationActions({
...
       </IconButton>
     );
   };
 
+  const inputProps = {
+    'data-testid': 'pagination-dropdown',
+    tabIndex,
+    id: 'pagination-dropdown',
+    style: {
+      color: theme.isBackgroundDarkColor ? 'rgba(255, 255, 255, 0.9)' : '#404040',
```
haxxmaxx, line 71: You are doing this check three times. create a variable and do it once

src/table/components/TablePaginationActions.jsx
```diff
@@ -51,31 +55,42 @@ export default function TablePaginationActions({
...
+    },
+  };
+
+  const selectStyle = {
+    backgroundColor: 'inherit',
+    '& .MuiNativeSelect-icon': {
+      color: theme.isBackgroundDarkColor ? 'rgba(255, 255, 255, 0.9)' : 'rgba(0, 0, 0, 0.54)',
```
haxxmaxx, line 78: This one should be iconButtonColor right?

src/index.js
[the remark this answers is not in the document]
LiKang6688: Good point, I will first remove the comments here and add them to the dev portal.

src/table/components/TablePaginationActions.jsx
LiKang6688: Good point, I gonna do the refactor in another PR as I found more cases like this one.

haxxmaxx:
After #453 is merged, fix the two comments so we finally can get this in. We might have to revisit this, but it is good for now

src/table/components/TablePaginationActions.jsx
LiKang6688: Right, good find. Thanks.

LiKang6688: #510

src/table/components/TablePaginationActions.jsx
LiKang6688: #510

PR #447, opened 2022-02-17
feat: introduce react-native-sn-table

vcellu:
### This PR introduces the react-native-sn-table.
It uses all of the logic from the supernova, but renders the table in a native fashion.

The plan for releasing is to release two separate packages.  One for sn-table and one for react-native-sn-table. Nebula.js build has been updated to accept a flag "--reactNative".  This will signal to rollup to pick up all .native files in favour of the plain files.  For example, if there are two files a.jsx and a.native.jsx, passing the "--reactNative" flag will pick up the a.native.jsx.  Not passing the flag will ignore a.native.jsx


niekvanstaveren (commented):
I guess the setup works. However it does not look like there is any code shared between the desktop and native variant. If that stays the case what is the need of having this in the same repo?

package.json
```diff
@@ -68,11 +70,13 @@
     "@mui/material": "5.3.0",
     "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest",
     "@mui/styles": "5.3.0",
+    "@nebula.js/cli": "2.5.0",
+    "@nebula.js/cli-build": "2.5.0",
+    "@nebula.js/cli-sense": "2.5.0",
+    "@nebula.js/cli-serve": "2.5.0",
+    "@qlik/react-native-simple-grid": "^1.0.3",
+    "@react-native-async-storage/async-storage": "^1.16.1",
```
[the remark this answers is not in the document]
haxxmaxx: We've fixed this, this package.json is outdated

package.json
```diff
@@ -32,6 +32,8 @@
     "build": "yarn run locale:generate && node ./tools/build.js --core && shx cp assets/* dist",
     "build:dev": "yarn run locale:generate && node ./tools/build.js --core --mode development && shx cp assets/* dist",
     "build:watch": "yarn run locale:generate && node ./tools/build.js --mode development -w",
+    "build:rn": "yarn run locale:generate && node ./tools/build.js --carbon",
```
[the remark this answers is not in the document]
vcellu: Changed to reactNative

react-native/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@nebula.js/react-native-sn-table",
+  "version": "1.4.13",
```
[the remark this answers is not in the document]
vcellu: Done

niekvanstaveren:
@Caele Could you have a look at this as well?

Caele (approved):
I'm a bit rusty, but don't see anything strange here, looks good.

niekvanstaveren (approved):

haxxmaxx:
The tests have passed but are failing to upload the coverge report. This is due to being on a fork, we haven't fixed this yet. I can force merge this if you want

src/handle-data.js
```diff
@@ -45,12 +45,12 @@ export default async function manageData(model, layout, pageInfo, setPageInfo) {
   const height = Math.min(rowsPerPage, totalVerticalCount - top);
   // When the number of rows is reduced (e.g. confirming selections),
   // you can end up still being on a page that doesn't exist anymore, then go back to the first page and return null
-  if (page > 0 && top >= totalVerticalCount) {
+  if (page > 0 && top >= totalVerticalCount && pageInfo) {
```
LiKang6688, line 48: why `pageInfo` is needed here?

src/handle-data.js
```diff
@@ -45,12 +45,12 @@ export default async function manageData(model, layout, pageInfo, setPageInfo) {
...
     setPageInfo({ ...pageInfo, page: 0 });
     return null;
   }
   // If the number of cells exceeds 10k then we need to lower the rows per page to the maximum possible value
-  if (height * totalHorizontalCount > MAX_CELLS) {
+  if (height * totalHorizontalCount > MAX_CELLS && pageInfo) {
```
LiKang6688, line 53: same concern here

LiKang6688:
Hi @vcellu, I have the same question as Niek. I do not see that react native table shares any code with the sn-table. Why do we want to put them together? Please correct me if I was wrong.

LiKang6688:
> Hi @vcellu, I have the same question as Niek. I only see that react native table shares `sortingFactory` and `manageData`, these two functions with the sn-table. Why do we want to put them together? Please correct me if I was wrong.

LiKang6688:
Hi @vcellu , I think it would be very nice if you could add some docs, such as how do we run, build, test, deploy, and develop the react-native-sn-table. Something like https://github.com/qlik-oss/sn-table/blob/main/.github/CONTRIBUTION.md and https://github.com/qlik-oss/sn-table/blob/main/README.md.

src/handle-data.js
vcellu: I didn't add the code for the MAX_CELLS, I just added the code for && pageInfo.

src/handle-data.js
vcellu: This is an infinite scroll signal, that there is no data.  So pagInfo is set to null to avoid the re-renders.

LiKang6688: I see. Thanks.

LiKang6688 (approved):

vcellu:
> 

We are waiting for a build PR to get merged.  Once that is complete we will create a full document.   On how to build and deploy.  Currently you just run `"build:rn": "yarn run locale:generate && node ./tools/build.js --reactNative".  But the full documentation will be released when we have merged the build tools into nebula.js

PR #448, opened 2022-02-17
docs: update api specs

LiKang6688:


haxxmaxx:
Do we ever define what a stringExpression is?

spec-configs/props.conf.js
```diff
@@ -20,17 +20,35 @@ module.exports = {
...
+        url: 'https://qlik.dev/apis/json-rpc/qix/schemas/#%23%2Fdefinitions%2Fschemas%2Fentries%2FNxHypercubeMode',
+      },
+      NxAttrExprDef: {
+        url: 'https://qlik.dev/apis/json-rpc/qix/schemas#%23%2Fdefinitions%2Fschemas%2Fentries%2FNxAttrExprDef',
+      },
+      NxInlineDimensionDef: {
+        url: 'https://qlik.dev/apis/json-rpc/qix/schemas#%23%2Fdefinitions%2Fschemas%2Fentries%2FNxInlineDimensionDef',
+      },
+      NxInlineMeasureDef: {
+        url: 'https://qlik.dev/apis/json-rpc/qix/schemas#%23%2Fdefinitions%2Fschemas%2Fentries%2FNxInlineMeasureDef',
```
haxxmaxx, line 51: Are all of these used?

LiKang6688: Yes, when you run `yarn spec`, you will find that those types are missing

haxxmaxx: ok 👍

src/object-properties.js
```diff
@@ -1,29 +1,32 @@
...
     /** @type {MeasureProperties[]} */
     qMeasures: [],
+    /** @type {schemasNxHypercubeMode} */
     qMode: 'S',
     /** @type {boolean} */
     qSuppressZero: false,
     /** @type {boolean} */
     qSuppressMissing: true,
-    /** @type {Array} */
+    /** @type {Array<integer<int32>>} */
```
haxxmaxx, line 27: isn't int enough?

LiKang6688: I copied from somewhere else without thinking about it. Let me fix it.

haxxmaxx: I mean why int32? Oddly specific for being js

LiKang6688: aha, how about `Array<number>`? it is the way used in typescript.

haxxmaxx:
shouldn't you generate the actual specs?

src/object-properties.js
haxxmaxx: I usually use `number[]` in jsdoc

LiKang6688: aha, jsdoc3. https://jsdoc.app/tags-type.htm
Screenshot 2022-02-23 at 10 12 52
There is a dot between Array and <MyClass> which looks odd. I gonna use `number[]`.

LiKang6688: Done

LiKang6688:
> shouldn't you generate the actual specs?

No. Generating the actual specs fails the CI. I think that the actual specs can be generated when we bump the version.

haxxmaxx:
Isn't that weird? Better to update it as soon as we can?

LiKang6688:
> Isn't that weird? Better to update it as soon as we can?

I think it is correct that specs are not updated when did not update the version.

haxxmaxx (approved):

PR #450, opened 2022-02-21
refactor: remove the limitation that requires at least one dimension

LiKang6688:


haxxmaxx (approved):

src/object-properties.js
```diff
@@ -13,9 +13,15 @@ const properties = {
    * @extends {qae.HyperCubeDef}
    */
   qHyperCubeDef: {
-    /** @type {DimensionProperties[]} */
+    /**
+     * The maximum amount of  dimensions is 15
```
LiKang6688, line 17: @haxxmaxx  is it ok?

src/object-properties.js
```diff
@@ -13,9 +13,15 @@ const properties = {
    * @extends {qae.HyperCubeDef}
    */
   qHyperCubeDef: {
-    /** @type {DimensionProperties[]} */
+    /**
+     * The maximum amount of dimensions is 15
```
LiKang6688, line 17: @haxxmaxx  is it ok?

haxxmaxx: yup

PR #453, opened 2022-02-23
feat: apply body content hover background color and color from the theme

LiKang6688:
```
theme: {
  straightTable: {
    content: {
      hover: {
        backgroundColor: 'black',
        color: 'red',
      },
    },
  },
}
```


src/table/utils/styling-utils.js
```diff
@@ -65,6 +65,23 @@ export function getHeadStyle(layout, theme) {
...
+    contentHoverBackgroundColor === objectBackgroundColor
+  );
+
+  const contentHoverFontColor = theme.getStyle('object', 'straightTable.content.hover', 'color');
+  const contentFontColor = theme.getStyle('object', 'straightTable.content', 'color');
+  const tableFontColor = theme.getStyle('object', 'straightTable', 'color');
+  const isHoverFontColorFromTheme = !(
+    contentHoverFontColor === contentFontColor || contentHoverFontColor === tableFontColor
+  );
+
```
LiKang6688, line 84: To check that the contentHoverBackgroundColor and contentHoverFontColor are really from object->straightTable->content->hover->backgroundColor and color

src/table/utils/styling-utils.js
```diff
@@ -65,6 +66,23 @@ export function getHeadStyle(layout, theme) {
 }
 
 export function getBodyStyle(layout, theme) {
```
haxxmaxx, line 68: This function is very complex and not really self-explanatory. We need to look into if it can be simplified, or split up IMO

LiKang6688: Right, i gonna try to split it up

LiKang6688: Done. Any suggestion is welcome.

src/table/utils/styling-utils.js
haxxmaxx: What do the properties on theme mean? How is it that if the content or table color is the same as the hover color, then it is a valid color. Can't you set one color for the content, then a different hover color?

src/table/utils/styling-utils.js
```diff
@@ -64,8 +65,38 @@ export function getHeadStyle(layout, theme) {
...
+  const tableFontColor = theme.getStyle('object', 'straightTable', 'color');
+  const isHoverFontColorFromTheme = !(
+    contentHoverFontColor === contentFontColor || contentHoverFontColor === tableFontColor
+  );
+
+  return isHoverFontColorFromTheme ? contentHoverFontColor : undefined;
+}
+
 export function getBodyStyle(layout, theme) {
+  const contentHoverBackgroundColorFromTheme = getContentHoverBackgroundColor(theme);
```
haxxmaxx, line 94: SInce we know that we everything in this function is relevant to the body/content, you can remove content from all variable names, makes it easier to read

src/table/utils/styling-utils.js
```diff
@@ -74,27 +105,32 @@ export function getBodyStyle(layout, theme) {
...
-
-  const hoverBackgroundColor = unsetHoverFontBackgroundColor
-    ? STYLING_DEFAULTS.HOVER_BACKGROUND
-    : setHoverBackgroundColor;
+  // when hovering, the default hover effect (a light gray hover background
+  // color and no hover font color).
+  // 4. There are both a hover font and a hover background color,
+  // when hovering, the hover font and the hover background color take effect.
+  // 5. The hover font color and hover background color from properties have a
+  // higher priorities than those from theme.
```
haxxmaxx, line 116: This isn't really a 5th case, 
```suggestion
  // Note that hover colors from properties have a
  // higher priority than those from theme.
```

src/table/utils/styling-utils.js
LiKang6688: Sure

src/table/utils/styling-utils.js
```diff
@@ -74,27 +105,32 @@ export function getBodyStyle(layout, theme) {
...
+  const unsetHoverBackgroundColor = isUnset(contentHoverBackgroundColor) && !contentHoverBackgroundColorFromTheme;
+  const unsetHoverFontColor = isUnset(contentHoverFontColor) && !contentHoverFontColorFromTheme;
+  const unsetHoverFontBackgroundColor = unsetHoverBackgroundColor && unsetHoverFontColor;
+
+  let hoverBackgroundColor = isUnset(contentHoverBackgroundColor)
+    ? contentHoverBackgroundColorFromTheme
+    : getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, contentHoverBackgroundColor);
+  hoverBackgroundColor = unsetHoverBackgroundColor ? '' : hoverBackgroundColor;
+  hoverBackgroundColor = unsetHoverFontBackgroundColor ? STYLING_DEFAULTS.HOVER_BACKGROUND : hoverBackgroundColor;
```
haxxmaxx, line 125: Not a huge fan of setting `hoverBackgroundColor` three times, but it is fine as long as you comment which of the 4 cases this corresponds to
```suggestion
  let hoverBackgroundColor = isUnset(contentHoverBackgroundColor) // case 1 and 4
    ? contentHoverBackgroundColorFromTheme
    : getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, contentHoverBackgroundColor);
  hoverBackgroundColor = unsetHoverBackgroundColor ? '' : hoverBackgroundColor; // case 2
  hoverBackgroundColor = unsetHoverFontBackgroundColor ? STYLING_DEFAULTS.HOVER_BACKGROUND : hoverBackgroundColor; // case 3
```

src/table/utils/styling-utils.js
haxxmaxx: Wait there is a `!` in there. So if they are not the same, then it is a "real" hover color?

src/table/utils/styling-utils.js
LiKang6688: Okay, updated

src/table/utils/styling-utils.js
```diff
@@ -64,8 +65,36 @@ export function getHeadStyle(layout, theme) {
...
+  return isHoverFontColorFromTheme ? hoverFontColor : undefined;
+}
+
 export function getBodyStyle(layout, theme) {
+  const hoverBackgroundColorFromTheme = getHoverBackgroundColor(theme);
+  const hoverFontColorFromTheme = getHoverFontColor(theme);
+
   const content = layout.components?.[0]?.content;
+  const contentHoverBackgroundColor = content?.hoverColor;
+  const contentHoverFontColor = content?.hoverFontColor;
```
haxxmaxx, line 97: These still say content

haxxmaxx: ah nvm, I see why

src/table/utils/styling-utils.js
```diff
@@ -80,14 +80,14 @@ export function getHeaderStyle(layout, theme) {
...
-    hoverBackgroundColor === contentBackgroundColor ||
-    hoverBackgroundColor === tableBackgroundColor ||
-    hoverBackgroundColor === objectBackgroundColor
-  );
+  const isHoverBackgroundColorFromTheme =
+    !(hoverBackgroundColor === tableBackgroundColor || hoverBackgroundColor === objectBackgroundColor) ||
+    // hoverBackgroundColor is green
+    // tableBackgroundColor is red
+    // objectBackgroundColor is green
+    (hoverBackgroundColor !== tableBackgroundColor && hoverBackgroundColor === objectBackgroundColor);
```
haxxmaxx, line 90: I think it is better to not make it super complicated here. Better to fix `getStyles` and accept we hava a possible corner case for now

LiKang6688: By the way, https://github.com/qlik-oss/nebula.js/pull/776 is ready to be reviewed.

src/table/utils/styling-utils.js
```diff
@@ -97,27 +103,37 @@ export function getBodyCellStyle(layout, theme) {
...
+  // case 1 or 4
+  let hoverBackgroundColor = isUnset(contentHoverBackgroundColor)
+    ? hoverBackgroundColorFromTheme
+    : getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, contentHoverBackgroundColor);
+  hoverBackgroundColor = unsetHoverBackgroundColor
+    ? '' // case 2
+    : hoverBackgroundColor;
+  hoverBackgroundColor = unsetHoverFontBackgroundColor
+    ? STYLING_DEFAULTS.HOVER_BACKGROUND // case 3
+    : hoverBackgroundColor;
```
haxxmaxx, line 128: Something like this should work if I understand it correctly:
1. there is a content hover color set -> use that
2. there is a hover color set on theme -> use that
3. If both 1 and 2 are false, no background is set. so now if a font color is set -> set to '' (no hover bg color)
4. else set to default color
```suggestion
  const unsetHoverFontColor = isUnset(contentHoverFontColor) && !hoverFontColorFromTheme;

  const  hoverBackgroundColor = !isUnset(contentHoverBackgroundColor)
    ? getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, contentHoverBackgroundColor) // case 1 and 4
    : hoverBackgroundColorFromTheme 
    ? hoverBackgroundColorFromTheme // case 1 and 4
    : !unsetHoverFontColor
    ? '' // case 2
    : STYLING_DEFAULTS.HOVER_BACKGROUND; // case 3

```

haxxmaxx: Also I would change isUnset to isSet to remove the negation. Now we would be checking if stuff exists

haxxmaxx (changes requested):

src/table/utils/styling-utils.js
```diff
@@ -97,27 +103,37 @@ export function getBodyCellStyle(layout, theme) {
...
-    : setHoverBackgroundColor;
+  // when hovering, the default hover effect (a light gray hover background
+  // color and no hover font color).
+  // 4. There are both a hover font and a hover background color,
+  // when hovering, the hover font and the hover background color take effect.
+  // Note that hover colors from properties have a
+  // higher priority than those from theme.
+  const unsetHoverBackgroundColor = isUnset(contentHoverBackgroundColor) && !hoverBackgroundColorFromTheme;
+  const unsetHoverFontColor = isUnset(contentHoverFontColor) && !hoverFontColorFromTheme;
+  const unsetHoverFontBackgroundColor = unsetHoverBackgroundColor && unsetHoverFontColor;
```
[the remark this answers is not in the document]
haxxmaxx: ignore this see my other comment

src/table/utils/styling-utils.js
haxxmaxx: So without negation it should be like this
```suggestion
  const isHoverFontColorSet = isSet(contentHoverFontColor) || !!hoverFontColorFromTheme;
  const  hoverBackgroundColor = isSet(contentHoverBackgroundColor)
    ? getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, contentHoverBackgroundColor) // case 1 and 4
    : hoverBackgroundColorFromTheme 
    ? hoverBackgroundColorFromTheme // case 1 and 4
    : isHoverFontColorSet
    ? '' // case 2
: STYLING_DEFAULTS.HOVER_BACKGROUND; // case 3
```

src/table/utils/styling-utils.js
```diff
@@ -54,7 +55,7 @@ export const getBaseStyling = (styleObj, objetName, theme) => {
   const baseStyle = {
     backgroundColor,
     fontFamily,
-    color: styleObj?.fontColor ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj?.fontColor) : color,
+    color: styleObj?.fontColor ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj.fontColor) : color,
```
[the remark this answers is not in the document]
LiKang6688: right, should skip it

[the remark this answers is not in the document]
LiKang6688: If we call `getColor`, we know that `fontColor` exists, right?

src/table/utils/styling-utils.js
LiKang6688: Learning from best practice, we should extract this nested ternary operation into an independent statement.

More to read:
https://medium.com/@benlmsc/stop-using-nested-ternary-operators-heres-why-53e7e078e65a

I gonna look into your 4 points and suggestion to change isUnset to isSet.

haxxmaxx: I wouldn't call it best practice, there are pros and cons with both: https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340.
I'd say if statements are a bit easier to read, but I am fine with both as long as the logic is as simple as possible

src/table/utils/styling-utils.js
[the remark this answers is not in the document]
haxxmaxx: ahhh yes, of coure my bad

src/table/utils/styling-utils.js
LiKang6688: My code can also write like 
```
const hoverBackgroundColor = unsetHoverFontBackgroundColor
    ? STYLING_DEFAULTS.HOVER_BACKGROUND
    : unsetHoverBackgroundColor
    ? ''
    : isUnset(hoverBackgroundColorFromLayout)
    ? hoverBackgroundColorFromTheme
    : getColor(STYLING_DEFAULTS.HOVER_BACKGROUND, theme, hoverBackgroundColorFromLayout);
```
I think it is the same as yours except that isUnset is replaced with isSet.

LiKang6688: But for good readability, I dislike **nested** ternary operation

haxxmaxx: I'm not saying you need to use nested terneries. I think it should be set only once. That you can do with if else and it would be easy to read as well

haxxmaxx (approved):
🚢

PR #454, opened 2022-02-23
feat: when the backgroundColor is transparent, adjust that based on the header content color

LiKang6688:


LiKang6688:
replaced by #455

PR #455, opened 2022-02-24
feat: when the backgroundColor is transparent, adjust that based on the header content color

LiKang6688:


src/table/utils/styling-utils.js
```diff
@@ -60,8 +60,14 @@ export const isUnset = (prop) => !prop || JSON.stringify(prop) === JSON.stringif
 
 export function getHeadStyle(layout, theme) {
   const header = layout.components?.[0]?.header;
-
-  return getBaseStyling(header, 'header', theme);
+  const headStyle = getBaseStyling(header, 'header', theme);
+  const setBackgroundColor = isDarkColor(headStyle.color) ? '#FAFAFA' : '#323232';
```
haxxmaxx, line 64: rename to `backgroundColor`, this sounds like a setter, not the value of the background Color

LiKang6688: yup

haxxmaxx (approved):

PR #456, opened 2022-02-24
feat: Add option to render pagination content inside specified element.

T-Wizard:
For use in Analytics view where pagination is combined with extra content

Example Image:
table

### DIFF
--- a/src/table/components/TableWrapper.jsx
+++ b/src/table/components/TableWrapper.jsx
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { useEffect, useState, useMemo, useRef } from 'react';
+import ReactDOM from 'react-dom';
 import Paper from '@mui/material/Paper';
 import Table from '@mui/material/Table';
 import TableContainer from '@mui/material/TableContainer';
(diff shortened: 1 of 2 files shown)


LiKang6688:
Can you post a screenshot here, please?

streamside:
I suggest to use `footer` as name for the option. Over time I suppose the content below the table may change and have other things than only pagination features.

LiKang6688:
Hi @T-Wizard, I agree with @streamside. Could you rename it?

T-Wizard:
> Hi @T-Wizard, I agree with @streamside. Could you rename it?

do you want the name to be just `footer`?
or should it be `footerContainer` to indicate that the footer will be placerd inside it. (and not an option to replace the footer content)

LiKang6688:
> > Hi @T-Wizard, I agree with @streamside. Could you rename it?
> 
> do you want the name to be just `footer`? or should it be `footerContainer` to indicate that the footer will be placerd inside it. (and not an option to replace the footer content)

footerContainer looks good to me.

LiKang6688 (approved):
LGTM

T-Wizard:
@haxxmaxx @niekvanstaveren any comment? (Or should i merge it)

src/table/components/TableWrapper.jsx
```diff
@@ -234,6 +249,7 @@ TableWrapper.propTypes = {
   selectionsAPI: PropTypes.object.isRequired,
   keyboard: PropTypes.object.isRequired,
   rect: PropTypes.object.isRequired,
+  footerContainer: PropTypes.object.isRequired,
```
niekvanstaveren, line 252: Should this be required?

niekvanstaveren (approved):
Looks good. I don't think the footerContainer should be required though

src/table/components/TableWrapper.jsx
T-Wizard: I will change it to not required.

haxxmaxx (approved):

PR #459, opened 2022-02-25
feat: adjust the sort label color based on the header background color

LiKang6688:
When you set the color of header content, the color of the sort label should be the same as that.
When there is no header content color setting, the sort label color is kind of white in the dark background color and black in the light background color.

### DIFF
--- a/src/table/components/TableHeadWrapper.jsx
+++ b/src/table/components/TableHeadWrapper.jsx
@@ -40,7 +40,15 @@ function TableHeadWrapper({
     },
   };
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme.name()]);
-
+  // When you set the header font color,
+  // the sort label color should be same.
+  // When there is no header content color setting,
+  // the sort label color is depending on the header background color.
+  const tableSortLabelStyle = {
+    '&.Mui-active .MuiTableSortLabel-icon': {
+      color: headerStyle.color ? headerStyle.color : headerStyle.sortLabelColor,
+    },
+  };
   return (
     <TableHead>
       <TableRow sx={headRowStyle} className="sn-table-row">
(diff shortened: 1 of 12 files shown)


src/table/components/TableHeadWrapper.jsx
```diff
@@ -40,7 +40,15 @@ function TableHeadWrapper({
...
-
+  // When you set the header font color,
+  // the sort label color should be same.
+  // When there is no header content color setting,
+  // the sort label color is depending on the header background color.
+  const tableSortLabelStyle = {
+    '&.Mui-active .MuiTableSortLabel-icon': {
+      color: headerStyle.color ? headerStyle.color : headerStyle.sortLabelColor,
+    },
+  };
```
haxxmaxx, line 51: Move to styling-utils

src/table/components/TableHeadWrapper.jsx
```diff
@@ -40,7 +40,15 @@ function TableHeadWrapper({
...
   };
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme.name()]);
-
+  // When you set the header font color,
+  // the sort label color should be same.
+  // When there is no header content color setting,
+  // the sort label color is depending on the header background color.
+  const tableSortLabelStyle = {
+    '&.Mui-active .MuiTableSortLabel-icon': {
+      color: headerStyle.color ? headerStyle.color : headerStyle.sortLabelColor,
```
haxxmaxx, line 49:
```suggestion
      color: headerStyle.color ?? headerStyle.sortLabelColor,
```

src/table/components/TableHeadWrapper.jsx
LiKang6688: Done

src/table/components/TableHeadWrapper.jsx
```diff
@@ -40,6 +40,11 @@ function TableHeadWrapper({
     },
   };
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme.name()]);
+  const tableSortLabelStyle = {
+    '&.Mui-active .MuiTableSortLabel-icon': {
+      color: headerStyle.sortLabelColor,
+    },
+  };
```
haxxmaxx, line 47: This is fine, but I have a refactor idea where we move all classes made for sx to styling-utils. So you get everything already packaged, and only updated when needed

haxxmaxx (approved):

PR #462, opened 2022-02-26
chore: run unit test and rendering test when commit changes

LiKang6688:


haxxmaxx:
Not a fan of this. I checkout the branch and tried it. First it failed since the locales where not generated. Then the rendering tests timed out, not sure why. If I where to change something irrelevant to the rendering tests, or just fix a small typo, I think it is overkill to run all tests, especially since it takes more than a minute to run. Sometimes you also want to commit something, then fix the unit tests. You can't do that now

LiKang6688:
> Not a fan of this. I checkout the branch and tried it. First it failed since the locales where not generated. Then the rendering tests timed out, not sure why. If I where to change something irrelevant to the rendering tests, or just fix a small typo, I think it is overkill to run all tests, especially since it takes more than a minute to run. Sometimes you also want to commit something, then fix the unit tests. You can't do that now

Right, sometimes we want to commit first then fix the unit tests later.

PR #465, opened 2022-03-01
docs: test extension against the Saas Edition of Qlik Sense

LiKang6688:


haxxmaxx:
This applies to all charts right? (The exact commands will differ of course.) I'd say then we want it documented somewhere general

LiKang6688:
> This applies to all charts right? (The exact commands will differ of course.) I'd say then we want it documented somewhere general

Should be when a chart works correctly. Any place do you recommend?

haxxmaxx:
here maybe? https://qlik.dev/libraries-and-tools/nebulajs/nebula-serve

LiKang6688:
> here maybe? https://qlik.dev/libraries-and-tools/nebulajs/nebula-serve

Ok, will do that.

PR #467, opened 2022-03-02
chore: make cc-test-reporter after-build do not run for forked PRs

LiKang6688:
When there is a secret for cc-test-reporter, make it run.
reference: https://circleci.com/blog/triggering-trusted-ci-jobs-on-untrusted-forks/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,7 +70,9 @@ jobs:
             chmod +x ./cc-test-reporter
             ./cc-test-reporter before-build
             yarn test:unit -- --mocha.bail false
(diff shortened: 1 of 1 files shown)


niekvanstaveren:
What about the diff coverage? That one is required it seems

niekvanstaveren:
> > What about the diff coverage? That one is required it seems
> 
> Which line of command in the circleci do you refer to?

Not sure where that is configured, but I see this 
image

LiKang6688:
Right, looked into it https://github.com/codeclimate/test-reporter/issues/452, it seems that the required codecliomate/diff-coverage needs cc-test-reporter.

PR #468, opened 2022-03-03
docs(example): use custom theme in mashup

LiKang6688:
Reference:
https://github.com/qlik-oss/nebula.js-examples/tree/main/examples/themes


haxxmaxx:
This is not a feature. I thought it had something to do with getting themes to work in a mashup, but it is just adding themes to the example

LiKang6688:
> This is not a feature. I thought it had something to do with getting themes to work in a mashup, but it is just adding themes to the example

docs? i am not sure about it.

LiKang6688:
Screenshot 2022-03-08 at 09 49 31
@haxxmaxx which one do you prefer?

haxxmaxx:
good question... docs(example) maybe. Want to avoid feat at least

haxxmaxx (approved):
I saw that codeclimate is picking this up, we probably want to ignore the whole folder

LiKang6688:
> I saw that codeclimate is picking this up, we probably want to ignore the whole folder

I will ignore the sense-dark-horizon.js file if possible. I see the benefit to keep the quality of mashup examples in place.

PR #477, opened 2022-03-12
fix: the case at the end of a horizontal scroll

LiKang6688:
fix the possibility of being not able to scroll vertically at the end of a horizontal scroll
as `evt.preventDefault()` is run at the wrong case


src/table/utils/handle-scroll.js
```diff
@@ -13,11 +13,11 @@ export const handleHorizontalScroll = (evt, isRTL, memoedContainer) => {
     // evt.deltaX increasingly negative as you scroll towards left,
     // increasingly positive as you scroll towards right
     const scrollRight = maxScrollableWidth + ScrollLeftWidth;
-    if (maxScrollableWidth > 0 && (scrollRight <= 0 || scrollRight > maxScrollableWidth)) {
+    if (maxScrollableWidth > 0 && (scrollRight < 0 || scrollRight > maxScrollableWidth + 1)) {
```
haxxmaxx, line 16: codeclimate does have a point here. The only difference is using `scrollRight` instead of `ScrollLeftWidth`, and `min` instead of `max`. Could refactor this if it doesn't hurt the readability. You could also early return when `maxScrollableWidth > 0` is false

src/table/utils/__tests__/handle-scroll.spec.js
```diff
@@ -20,14 +20,14 @@ describe('handle-scroll', () => {
       isRTL = true;
     });
 
-    it('should run preventDefault when scroll to leftmost place in ltr direction', () => {
+    it('should run preventDefault when scroll to the leftmost place and keep scrolling left in LTR direction', () => {
```
haxxmaxx, line 23: This is actually what happens right? Since you only scroll once
```suggestion
    it('should run preventDefault when you are at leftmost place and keep scroll left in LTR direction', () => {
```

src/table/utils/__tests__/handle-scroll.spec.js
```diff
@@ -53,22 +53,33 @@ describe('handle-scroll', () => {
...
+
+    it('should not run preventDefault when scroll to the rightmost place but not keep scrolling right in LTR direction', () => {
+      evt = {
+        ...evt,
+        deltaX: 0,
       };
       memoedContainer = {
         ...memoedContainer,
-        scrollLeft: 400,
+        scrollLeft: 1369,
```
haxxmaxx, line 74: No point of having uneven numbers here, it just makes the tests harder to understand.

src/table/utils/handle-scroll.js
```diff
@@ -1,25 +1,28 @@
 export const handleHorizontalScroll = (evt, isRTL, memoedContainer) => {
   evt.stopPropagation();
+
   // scrollWidth is the width of an element's content, including content not visible on the screen due to overflow.
   // offsetWidth is the element's CSS width, including any borders, padding, and vertical scrollbars
+  const maxScrollableWidth = memoedContainer.scrollWidth - memoedContainer.offsetWidth;
+  window.console.log(memoedContainer.scrollWidth);
+  window.console.log(memoedContainer.offsetWidth);
```
haxxmaxx, line 8: remove

LiKang6688: Done

PR #482, opened 2022-03-16
feat: auto hide pagination when it is not necessary

LiKang6688:
Pagination auto-hides when it is not needed
totalVerticalCount <= 10

raname totalHeight to totalVerticalCount
width to totalHorizontalCount

Screenshot 2022-03-16 at 21 58 59


haxxmaxx:
Code looks fine, but I don't see the renaming you are mentioning?
Also there seem to be screenshots that hide the pagination even though they have more than 10 rows, should it be like that?

LiKang6688:
> Code looks fine, but I don't see the renaming you are mentioning? Also there seem to be screenshots that hide the pagination even though they have more than 10 rows, should it be like that?

yes, i thought to do the refactoring but found that it is not related to this PR anymore. So I gonna remove the renaming thing in the description.

test/rendering/__fixtures__/scenario_2.fix.js
```diff
@@ -0,0 +1,1231 @@
...
+          qType: 'sn-table',
+        },
+        qMeta: {
+          privileges: ['read', 'update', 'delete', 'exportdata'],
+        },
+        qSelectionInfo: {},
+        qHyperCube: {
+          qSize: {
+            qcx: 1,
+            qcy: 6,
```
LiKang6688, line 17: you are right, there are more than 10 rows. But only qSize is checked which is below 10 @haxxmaxx

LiKang6688:
> 



> Code looks fine, but I don't see the renaming you are mentioning? Also there seem to be screenshots that hide the pagination even though they have more than 10 rows, should it be like that?

renaming is done now

test/rendering/__fixtures__/scenario_2.fix.js
haxxmaxx: why is qcy < 10 here? I think it is misleading that the code implies that the pagination should be hidden there are less than 10 rows, but the screenshots show otherwise

LiKang6688: just for the purpose of testing. The problem is that the  `qHyperCube` does not match with `getHyperCubeData` as they are from different places. Gonna work on it.

LiKang6688: Done

src/table/components/TableWrapper.jsx
```diff
@@ -123,7 +123,7 @@ export default function TableWrapper(props) {
   };
 
   const tableContainerStyle = {
-    height: constraints.active || footerContainer ? '100%' : 'calc(100% - 52px)',
+    height: constraints.active || footerContainer || totalVerticalCount <= 10 ? '100%' : 'calc(100% - 52px)',
```
haxxmaxx, line 126: I would do something like `const noPaginationNeeded = totalVerticalCount <= 10` since you use it in three different places

haxxmaxx (approved):
LGTM, the name changes makes the the code easier to read, good stuff

src/table/components/TableWrapper.jsx
LiKang6688: Right. Thanks. Updated.

PR #485, opened 2022-03-22
chore: run yarn build to generate sn-table-ext

LiKang6688:
miss sn-table-ext in the packing
https://unpkg.com/browse/@nebula.js/sn-table@1.9.0/

add `module": "core/esm/index.js"` to fix Error: "export 'useId' (imported as 'React') was not found in 'react'" after upgrading" (https://github.com/mui/material-ui/issues/29860)


haxxmaxx:
we need this included in the package right? So we need to include the sense script somewhere where we build?

LiKang6688:
> we need this included in the package right? So we need to include the sense script somewhere where we build?

Yes, correct. I will do that. Thanks

haxxmaxx:
@Caele would be great if you can have a look so we get the bundling correct once and for all :)

package.json
```diff
@@ -22,19 +22,22 @@
...
     "api-specifications",
-    "core"
+    "core",
+    "sn-table-ext"
   ],
   "engines": {
     "node": ">=16"
   },
   "main": "dist/sn-table.js",
+  "module": "core/esm/index.js",
```
LiKang6688, line 32: To fix error: "export 'useId' (imported as 'React') was not found in 'react'" after upgrading" (https://github.com/mui/material-ui/issues/29860)

Caele (approved):
Looks correct, but it is the published output that matters.

PR #490, opened 2022-03-28
docs(readme): load the sn-table through the script tag from unpkg

LiKang6688:
<script src="https://unpkg.com/@nebula.js/sn-table"></script>


haxxmaxx (approved):
LGTM as long you've tested it

PR #503, opened 2022-04-20
chore: add a script to copy sn-table-ext into local extension folder

LiKang6688:


haxxmaxx:
why do we need this?

LiKang6688:
> why do we need this?

To manually verify that the sn-table-ext works correctly.

haxxmaxx (approved):

PR #508, opened 2022-04-29
feat: shift + arrow up/down to select multiple values

LiKang6688:


haxxmaxx:
I am doing some refactoring for the selection-utils for performance so this will have to wait a bit

src/table/components/TableBodyWrapper.jsx
```diff
@@ -98,6 +98,8 @@ function TableBodyWrapper({
                       selectionState,
                       cell,
                       selectionDispatch,
+                      tableRows: rows,
```
haxxmaxx, line 101: I think this is not a good idea. You are basically putting all the table data on every cell. I see you need qElemNumber for the cell above and bellow. If you have to pass multiple cells, I would prefer only sending those two cells extra cells that are needed

LiKang6688: I agree that putting all the table data on every cell is not good. It is a bit overkill and would be good to pass one-column data.

But putting all the table data on every cell does not impact performance. It is a copy by reference.

And getting two cells extra cells for every cell impacts performance. As it needs to do operations on the array.

src/table/utils/selections-utils.js
```diff
@@ -95,7 +104,17 @@ export function selectCell({ selectionState, cell, selectionDispatch, evt, annou
     return;
   }
 
-  selectedRows = getSelectedRows({ selectedRows, qElemNumber, rowIdx, evt });
+  if (isSelectMultiCells) {
```
haxxmaxx, line 107: should be able to check what keys are pressed on `evt` and get rid of this argument

src/table/components/TableBodyWrapper.jsx
haxxmaxx: what about putting the previous and next cell `qElemNumber` on the cell in `handleData`? That way you don't have to add any arguments at all.

LiKang6688:
a new one gonna come.

PR #511, opened 2022-05-03
refactor: set theme color in a central place

LiKang6688:
- fix icon color for rows per page in select icon
- fix sort label color
- set theme color in a central place - src/table/utils/table-theme-colors.js. Motivation: easy to find and modify.


src/table/utils/styling-utils.js
```diff
@@ -69,19 +68,22 @@ export const getBaseStyling = (styleObj, objetName, theme) => {
 export function getHeaderStyle(layout, theme) {
   const header = layout.components?.[0]?.header;
   const headerStyle = getBaseStyling(header, 'header', theme);
+  headerStyle.borderWidth = '1px 1px 1px 0px';
 
   // When the table background color from the sense theme is transparent,
   // there is a header background color depending on the header font color
   // to avoid seeing the table body through the table head.
   const headerBackgroundColor = isDarkColor(headerStyle.color) ? '#FAFAFA' : '#323232';
```
haxxmaxx, line 76: This one you could move as well right?

src/table/utils/styling-utils.js
```diff
@@ -69,19 +68,22 @@ export const getBaseStyling = (styleObj, objetName, theme) => {
...
+  headerStyle.backgroundColor =
+    theme.table.backgroundColor === 'transparent' ? headerBackgroundColor : theme.table.backgroundColor;
+
   // When you set the header font color,
   // the sort label color should be same.
   // When there is no header content color setting,
   // the sort label color is depending on the header background color.
   headerStyle.sortLabelColor =
-    headerStyle.color ?? isDarkColor(headerStyle.backgroundColor) ? 'rgba(255,255,255,0.9)' : 'rgba(0, 0, 0, 0.54)';
+    headerStyle.color ?? (isDarkColor(headerStyle.backgroundColor) ? 'rgba(255,255,255,0.9)' : 'rgba(0, 0, 0, 0.54)');
```
haxxmaxx, line 85: This is the same as the icon color, not sure if it makes sense to use it here?

src/table/utils/styling-utils.js
LiKang6688: `headerStyle.borderWidth = '1px 1px 1px 0px';`
and
`contentStyle.borderWidth = '0px 1px 1px 0px';`
so I did not move borderWidth to `getBaseStyling`

haxxmaxx: I mean headerBackgroundColor

src/table/utils/styling-utils.js
LiKang6688: they are differeent. 
this one is checking `isDarkColor(headerStyle.backgroundColor)`
the icon color is checking `isDarkColor(backgroundColor)`

src/table/utils/styling-utils.js
LiKang6688: I had a try, and we need to move both the headerStyle and getBaseStyling then, which it is better to be in styling-utils.js not in table-theme-colors.js

LiKang6688: came up with a way:
```
const header = (color) => {
    return {
      iconColor: isDarkColor(color) ? 'rgba(255,255,255,0.9)' : 'rgba(0, 0, 0, 0.54)',
      backgroundColor: isDarkColor(color) ? '#FAFAFA' : '#323232',
    };
  };
```
maybe it is a bit overkill. But the idea is to center all those colors.

src/table/utils/styling-utils.js
haxxmaxx: ok good

src/table/utils/styling-utils.js
haxxmaxx: yeah I think it is a bit overkill. I don't like that we add functions to the theme like this. I missed that in this case we are dependent on `getBaseStyling(header, 'header', theme)`, not on the `theme.backgroundColor`

haxxmaxx: It made more sense before

LiKang6688: sure, i gonna revert it back

haxxmaxx (approved):

PR #514, opened 2022-05-05
fix: check is color set

LiKang6688:


src/table/utils/styling-utils.js
```diff
@@ -54,7 +57,7 @@ export const getBaseStyling = (styleObj, objetName, theme) => {
   const baseStyle = {
     backgroundColor: theme.table.backgroundColor,
     fontFamily,
-    color: styleObj?.fontColor ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj.fontColor) : color,
+    color: isSet(styleObj?.fontColor) ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj.fontColor) : color,
```
haxxmaxx, line 60: Is there a test on these two cases?

LiKang6688: Good question. Do not think there is any. Let me add some.

LiKang6688: Done

src/table/utils/__tests__/styling-utils.spec.js
```diff
@@ -113,6 +113,29 @@ describe('styling-utils', () => {
...
     });
+    it('should return styling with fontSize and padding when the index for font color is -1 and color is null', () => {
+      styleObj.fontColor = { index: -1, color: null };
+      const resultStyling = getBaseStyling(styleObj, objetName, theme);
+      expect(resultStyling).toEqual({
+        backgroundColor: '#323232',
+        borderColor: '#D9D9D9',
+        borderStyle: 'solid',
+        fontSize: 12,
+        padding: '6px 12px',
```
haxxmaxx, line 124: Is this the real case scenario? that you don't get a color set?

haxxmaxx: only if there is something set on the theme?

LiKang6688: You are right. When there is no color from Layout and color from the theme, you get color from the theme. 
In the test case, there is no color from either Layout or theme, so you don't get any color.

LiKang6688: Adding a new test case addressing it.

haxxmaxx (approved):

PR #515, opened 2022-05-05
Fix: table boder without pagination

LiKang6688:
before:
Screenshot 2022-05-05 at 02 58 49

after:
Screenshot 2022-05-05 at 02 47 02


src/table/components/TableWrapper.jsx
```diff
@@ -238,9 +238,15 @@ export default function TableWrapper(props) {
...
-          <TableHeadWrapper {...props} setFocusedCellCoord={setFocusedCellCoord} focusedCellCoord={focusedCellCoord} />
+          <TableHeadWrapper
+            {...props}
+            paginationNeeded={paginationNeeded}
+            setFocusedCellCoord={setFocusedCellCoord}
+            focusedCellCoord={focusedCellCoord}
+          />
           <TableBodyWrapper
             {...props}
+            paginationNeeded={paginationNeeded}
```
haxxmaxx, line 249: I don't think we should pass another prop here. I had a look at all the how we set this variable up top. We calculate it every time the TableWrapper re-renders. But it will only change if the data changes. So we might as well calculate it at that point and put it on tableData. Let me do a quick refactor and you can use that

LiKang6688: Sounds good

src/table/components/TableBodyWrapper.jsx
```diff
@@ -49,7 +49,10 @@ function TableBodyWrapper({
 
   const bodyRowAndCellStyle = {
     'tr :last-child': {
-      borderRight: 0,
+      borderRight: paginationNeeded && 0,
+    },
+    'tr :first-child': {
+      borderLeft: !paginationNeeded && '1px solid rgb(217, 217, 217)',
```
haxxmaxx, line 55: We should look into color constants, but that is for a different pr

haxxmaxx (approved):

PR #517, opened 2022-05-06
fix: background color from theme with opacity

LiKang6688:
- a table background color has the highest priority
```
"object":{
  "straightTable": {
     backgroundColor:''
   }
}
```
- when there is no table background color, the CSS color for .qv-inner-object overrides the object and sheet background color from the theme.
```
.qv-object .qv-inner-object {
    background-color: rgba(254, 222, 235, 0.7) !important;
}
```

```
{
  backgroundColor:''
  "object":{
    backgroundColor:''
  }
}
```
- when the applied background color has opacity(like 0.2, not 0), the header background color should remove that opacity value (alpha value in rgba)

Screenshot 2022-05-07 at 01 33 44


src/table-theme-colors.js
```diff
@@ -1,7 +1,13 @@
 import { isDarkColor } from './table/utils/color-utils';
 
 export default function tableThemeColors(theme) {
-  const backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
+  const qvInnerObject = document.querySelector('.qv-object .qv-inner-object');
+  const backgroundColorFromCSS = qvInnerObject && window.getComputedStyle?.(qvInnerObject).backgroundColor;
```
LiKang6688, line 5: Reference:
https://zellwk.com/blog/css-values-in-js/

src/table-theme-colors.js
```diff
@@ -1,7 +1,13 @@
 import { isDarkColor } from './table/utils/color-utils';
 
 export default function tableThemeColors(theme) {
-  const backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
+  const qvInnerObject = document.querySelector('.qv-object .qv-inner-object');
```
haxxmaxx, line 4: So this is where the theme where we found the bug adds the color. But I think it could on both a higher and a lower level. We need to make sure it works for other use cases

src/table/components/TableWrapper.jsx
```diff
@@ -130,7 +130,7 @@ export default function TableWrapper(props) {
     borderStyle: 'solid',
     borderColor: theme.table.borderColor,
     height: '100%',
-    backgroundColor: theme.table.backgroundColor,
+    backgroundColor: theme.table.tableBackgroundColor ? theme.table.backgroundColor : 'inherit',
```
haxxmaxx, line 133: if `tableBackgroundColor` exists, use `backgroundColor`?

src/table/utils/color-utils.js
```diff
@@ -46,13 +47,25 @@ export function isDarkColor(color) {
...
     r = parseInt(matches[1], 16);
     g = parseInt(matches[2], 16);
     b = parseInt(matches[3], 16);
   }
+  let HSP = 0.299 * r + 0.587 * g + 0.114 * b;
+  HSP = a ? HSP / a : HSP;
   // Using the HSP (Highly Sensitive Poo) value, determine whether the color is light or dark
   // HSP < 125, the color is dark, otherwise, the color is light
-  return 0.299 * r + 0.587 * g + 0.114 * b < 125;
+  return a !== 0 && HSP < 125;
```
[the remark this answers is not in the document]
haxxmaxx: the whole concept of a color is bright or not is derived from humans perceive color. That is where the different factors for red, green and blue come from. I don't think we should add something similar for alpha without any theory behind it. Also, if something is transparent, the resulting color depends on what is behind it

src/table/components/TableWrapper.jsx
LiKang6688: Although it does not affect the outcome, I will fix it.

src/table-theme-colors.js
haxxmaxx: I mean technically you can set any color using css. But when users do that, they set it to inner-object, then this is fine

src/table/components/TableWrapper.jsx
LiKang6688: Fixed

src/table/utils/styling-utils.js
```diff
@@ -77,8 +76,8 @@ export function getHeaderStyle(layout, theme) {
   // there is a header background color depending on the header font color
   // to avoid seeing the table body through the table head.
   const headerBackgroundColor = isDarkColor(headerStyle.color) ? '#FAFAFA' : '#323232';
-  headerStyle.backgroundColor =
-    theme.table.backgroundColor === 'transparent' ? headerBackgroundColor : theme.table.backgroundColor;
+  const RGB = removeOpacity(theme.table.backgroundColor);
+  headerStyle.backgroundColor = theme.table.backgroundColor === 'transparent' ? headerBackgroundColor : RGB;
```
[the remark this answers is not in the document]
LiKang6688: Fixed

src/table/utils/color-utils.js
```diff
@@ -34,25 +43,33 @@ export function resolveExpression(input) {
...
+  const matches =
+    /^rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/i.exec(rgba) ||
+    /^rgba\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d(\.\d+)?)\s*\)$/i.exec(rgba);
+  const r = +matches?.[1];
+  const g = +matches?.[2];
+  const b = +matches?.[3];
+
   // Using the HSP (Highly Sensitive Poo) value, determine whether the color is light or dark
+  // the equation is from http://alienryderflex.com/hsp.html
+  const HSP = Math.sqrt(0.299 * (r * r) + 0.587 * (g * g) + 0.114 * (b * b));
```
haxxmaxx, line 56: Let's not change this. This is shared between charts, so even if this is more accurate, we should review this when we move this to nebula or similar

src/table/utils/color-utils.js
```diff
@@ -34,25 +43,33 @@ export function resolveExpression(input) {
...
-  let matches;
-  if (
-    (matches =
-      /^rgb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/i.exec(color) ||
-      /^rgba\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d(\.\d+)?)\s*\)$/i.exec(color))
-  ) {
-    r = parseInt(matches[1], 10);
-    g = parseInt(matches[2], 10);
-    b = parseInt(matches[3], 10);
-  } else if ((matches = /^#([A-f0-9]{2})([A-f0-9]{2})([A-f0-9]{2})$/i.exec(color))) {
```
haxxmaxx, line 49: what happened to this case?

src/table/utils/color-utils.js
LiKang6688: Sure

src/table/utils/color-utils.js
LiKang6688: const rgba = resolveExpression(color);
color was converted first, so no need to handle the hex code

haxxmaxx:
We need to actually decide a what we need to solve here. I'm starting a conversation in the channel

src/table/utils/styling-utils.js
```diff
@@ -73,12 +72,14 @@ export function getHeaderStyle(layout, theme) {
   const headerStyle = getBaseStyling(header, 'header', theme);
   headerStyle.borderWidth = '1px 1px 1px 0px';
 
-  // When the table background color from the sense theme is transparent,
+  // To avoid seeing the table body through the table head:
+  // - When the table background color from the sense theme is transparent,
   // there is a header background color depending on the header font color
```
haxxmaxx, line 77: when did we decide to do this? I don't remember

src/table/utils/styling-utils.js
```diff
@@ -73,12 +72,14 @@ export function getHeaderStyle(layout, theme) {
...
-  // to avoid seeing the table body through the table head.
+  // - When the table background color from the sense theme has opacity,
+  // removing that.
   const headerBackgroundColor = isDarkColor(headerStyle.color) ? '#FAFAFA' : '#323232';
-  headerStyle.backgroundColor =
-    theme.table.backgroundColor === 'transparent' ? headerBackgroundColor : theme.table.backgroundColor;
+  const RGB = removeOpacity(theme.table.backgroundColor);
+  headerStyle.backgroundColor = theme.table.isBackgroundTransparentColor ? headerBackgroundColor : RGB;
```
haxxmaxx, line 82: No need to run removeOpacity if it is not needed
```suggestion
  headerStyle.backgroundColor = theme.table.isBackgroundTransparentColor ? headerBackgroundColor : removeOpacity(theme.table.backgroundColor);
```

haxxmaxx: wait shouldn't it be the other way around?

src/table-theme-colors.js
```diff
@@ -1,8 +1,23 @@
-import { isDarkColor } from './table/utils/color-utils';
+import { isDarkColor, isTransparentColor } from './table/utils/color-utils';
 
 export default function tableThemeColors(theme) {
-  const backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
-  const isBackgroundDarkColor = isDarkColor(backgroundColor);
+  const qvInnerObject = document.querySelector('.qv-object .qv-inner-object');
+  const objectBackgroundColorFromCSS = qvInnerObject && window.getComputedStyle?.(qvInnerObject).backgroundColor;
```
haxxmaxx, line 5: why wouldn't getComputedStyle exist?

src/table-theme-colors.js
```diff
@@ -1,8 +1,23 @@
...
+  const qvPanelSheet = document.querySelector('.qv-panel-sheet');
```
haxxmaxx, line 6: We need to sync this with the sheet background color project so this is the same element they are applying the color too

src/table-theme-colors.js
```diff
@@ -1,8 +1,23 @@
...
+  const sheetBackgroundColorFromCSS = qvPanelSheet && window.getComputedStyle?.(qvPanelSheet).backgroundColor;
+  const opacityFromCSS = qvInnerObject && window.getComputedStyle?.(qvInnerObject).opacity;
+
+  const tableBackgroundColorFromTheme = theme.getStyle('', '', 'object.straightTable.backgroundColor');
+  const backgroundColorFromTheme = theme.getStyle('object', 'straightTable', 'backgroundColor');
+
+  const backgroundColor = tableBackgroundColorFromTheme || objectBackgroundColorFromCSS || backgroundColorFromTheme;
+  const isBackgroundTransparentColor = isTransparentColor(
+    backgroundColor,
+    tableBackgroundColorFromTheme ?? opacityFromCSS
```
haxxmaxx, line 16: this is a bit weird that you would pass `tableBackgroundColorFromTheme` if it exists, it is not opacity. If we would convert to ts, it would not like this.

src/table/utils/color-utils.js
haxxmaxx: The thing is that this file exists in multiple repos. Optimally they should be moved to a separate package. Before that is done, we want to keep it as similar as possible, to easily extract it. If I understand what you are trying to do, you need the color on a rgba format for `isTransparentColor` and `removeOpacity`. The thing is that hex will never have transparancy so you can early return in that case

src/table/utils/styling-utils.js
LiKang6688: Here https://github.com/qlik-oss/sn-table/pull/442/files#diff-03c15d1e38a398c39952d4fe961d27a310bf4d1959f3b86b92a0cd7d5721f846R74

src/table-theme-colors.js
LiKang6688: just in case that https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle#browser_compatibility

src/table/utils/color-utils.js
LiKang6688: the hex has transparency when there are eight digits https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4
So in https://github.com/qlik-oss/sn-table/pull/517/files#diff-116e1caa786a873fe51a85a99bf012cc7b016e631b783c26e99cac13ee0eb1e5R9, we should get the alpha value from the hex code.

src/table-theme-colors.js
haxxmaxx: yeah it is supported by everything, so should never be a problem

src/table/utils/styling-utils.js
LiKang6688: No, fixed

src/table-theme-colors.js
LiKang6688: fixed

src/table/utils/color-utils.js
haxxmaxx: Ahh that is new to me. ok so what resolve expression should do is return a value that is valid css. Hmm this is tricky. So if you really want to do it this way, I think you should rewrite it to a `resolveToRGBA` function. Or just add a case to `isTransparentColor` and `removeOpacity` that covers hexcolors with eight digits

src/table-theme-colors.js
LiKang6688: sure, fixed

src/table/utils/color-utils.js
LiKang6688: okay, I will rename the resolveExpression to resolveToRGBAorRGB

src/table/utils/styling-utils.js
haxxmaxx: Looked at the old PR and it makes sense

src/table/utils/color-utils.js
LiKang6688: Done

src/table-theme-colors.js
haxxmaxx: nvm the selector here can easily be changed. In any case, I think we should only look in one place for sheet color and one for object color, so this is good

src/table-theme-colors.js
```diff
@@ -1,8 +1,23 @@
...
-  const backgroundColor = theme.getStyle('object', 'straightTable', 'backgroundColor');
-  const isBackgroundDarkColor = isDarkColor(backgroundColor);
+  const qvInnerObject = document.querySelector('.qv-object .qv-inner-object');
+  const objectBackgroundColorFromCSS = qvInnerObject && window.getComputedStyle(qvInnerObject).backgroundColor;
+  const opacityFromCSS = qvInnerObject && window.getComputedStyle(qvInnerObject).opacity;
+  const qvPanelSheet = document.querySelector('.qv-panel-sheet');
+  const sheetBackgroundColorFromCSS = qvPanelSheet && window.getComputedStyle(qvPanelSheet).backgroundColor;
+
+  const tableBackgroundColorFromTheme = theme.getStyle('', '', 'object.straightTable.backgroundColor');
+  const backgroundColorFromTheme = theme.getStyle('object', 'straightTable', 'backgroundColor');
```
haxxmaxx, line 11: ok more questions... Isn't this already applied to `qvInnerObject`? And what happens outside the client? do we have access to the theme at all?

LiKang6688: inside the client, yes, it gets the same backgroundColor. But in the mashup, we can only get the backgroundColor from the theme, not from the qvInnerObject.
Yes, we do have access to the theme https://github.com/qlik-oss/sn-table/blob/main/mashup-example/using-build-tool/qlik-sense-desktop/configure.js#L17

haxxmaxx: ok that makes sense. The only thing I am missing now is some tests on this file

haxxmaxx: this isn't resolved. I don't see any tests on this file at all. Please add some, since this logic is not that straight forward

LiKang6688: do you mean unit tests for table-theme-colors.js?

haxxmaxx: yes

LiKang6688: okay, gonna try to add some

LiKang6688: Done

src/__test__/table-theme-colors.spec.js
```diff
@@ -3,42 +3,479 @@ import tableThemeColors from '../table-theme-colors';
...
-      },
+      global.window.getComputedStyle = jest.fn((selector) => {
+        if (selector === '.qv-panel-sheet') {
+          return { backgroundColor: '#000' };
+        }
+        return undefined;
+      });
+      result = tableThemeColors(theme);
+      expect(result).toEqual(defaultValue);
+    });
```
haxxmaxx, line 80: Is this suppose to be two tests?

src/__test__/table-theme-colors.spec.js
```diff
@@ -3,42 +3,479 @@ import tableThemeColors from '../table-theme-colors';
...
+        expect(result).toEqual({
+          ...defaultValue,
+          backgroundColor: '#11111100',
+          isBackgroundTransparentColor: true,
+        });
+      });
+    });
+
+    describe('when there is background color on object and background color on sheet', () => {
+      it('should return the when the background color on object is dark and not transparent', () => {
```
haxxmaxx: should return the...?

src/locale/locales/de.json
```diff
@@ -56,7 +56,7 @@
   },
   "SNTable.SelectionLabel.DeselectedValue": {
     "value": "Die Werteauswahl ist aufgehoben.",
-    "comment": "announced by screen readers  when a value is deselected. (tew 210929)",
+    "comment": "announced by screen readers when a value is deselected. (tew 210929)",
```
haxxmaxx, line 59: please don't do irrelevant changes like this to big PRs

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,481 @@
...
+          return style;
+        });
+        const result = tableThemeColors(theme);
+        expect(result).toEqual({
+          ...darkBackgroundColorValue,
+          backgroundColor: '#000',
+        });
+      });
+
+      it('should return the defaultValue and backgroundColor and isBackgroundTransparentColor to be true when the background color is transparent', () => {
```
haxxmaxx, line 120: So for all the transparent tests, there is no need to do more than one type of transparent. You already added tests for the isTransparent function, so you can trust that it will return the correct thing

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,481 @@
...
+        expect(result).toEqual({
+          ...defaultValue,
+          backgroundColor: '#11111100',
+          tableBackgroundColorFromTheme: '#11111100',
+          isBackgroundTransparentColor: true,
+        });
+      });
+    });
+
+    describe('when there is background color on table and background color on object', () => {
```
haxxmaxx, line 322: I think you can scratch all of these. This should behave exactly as the prev describe, since it always picks the table color instead of the object, and the object color has no affect, right?

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,481 @@
...
+        });
+        const result = tableThemeColors(theme);
+        expect(result).toEqual({
+          ...darkBackgroundColorValue,
+          backgroundColor: 'rgba(0,0,0,0)',
+          isBackgroundTransparentColor: true,
+        });
+      });
+    });
+  });
```
haxxmaxx, line 256: You need to check  `isBackgroundDarkColor` otherwise the test check the same thing

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,481 @@
...
+      expect(result).toEqual(defaultValue);
+
+      global.window.getComputedStyle = jest.fn((selector) => {
+        if (selector === '.qv-panel-sheet') {
+          return { backgroundColor: '#000' };
+        }
+        return undefined;
+      });
+      result = tableThemeColors(theme);
+      expect(result).toEqual(defaultValue);
```
haxxmaxx, line 79: Is this what we want in this case? the sheet background is dark, but `isBackgroundDarkColor` is false?

src/locale/locales/de.json
haxxmaxx: it is fine for now, but for next time :)

haxxmaxx:
I personally don't think it is a good idea to do multiple tests per it. I think it should be one setup, run the function once (maybe some other steps to, but not rerunning the same function) and check what happened.

src/__test__/table-theme-colors.spec.js
LiKang6688: Sorry, removed it and will be up in another PR.

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,481 @@
...
+    });
+  });
+
+  describe('when there is background color on object', () => {
+    describe('when there is only background color on object', () => {
```
haxxmaxx, line 84: Is this even possible by the way? I mean if you are in the client, both the sheet and the object selectors exist? and then there should be a computed style for each of them?

src/locale/locales/de.json
LiKang6688: Sorry, removed it, and will be up in another PR.

src/__test__/table-theme-colors.spec.js
LiKang6688: we also need to consider cases in mashup, so it is possible.

haxxmaxx: ok 👍

LiKang6688:
> I personally don't think it is a good idea to do multiple tests per it. I think it should be one setup, run the function once (maybe some other steps to, but not rerunning the same function) and check what happened.

I agree. Let me think about it.

src/__test__/table-theme-colors.spec.js
LiKang6688: good question. 
I realized that in client, objectBackgroundColorFromCSS will at least be rgba(0, 0, 0, 0) which is impossible to be undefine. 
I cannot figure out why.

src/__test__/table-theme-colors.spec.js
```diff
@@ -3,42 +3,479 @@ import tableThemeColors from '../table-theme-colors';
...
+      it('should return the default value and backgroundColor when the background color is light', () => {
+        global.document.querySelector = jest.fn((selector) => {
+          if (selector === '.qv-object .qv-inner-object') return selector;
+          return undefined;
+        });
+        global.window.getComputedStyle = jest.fn((selector) => {
+          let style;
+          if (selector === '.qv-object .qv-inner-object') {
+            style = { backgroundColor: '#fff' };
+          }
```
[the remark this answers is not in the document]
LiKang6688: Good idea. Working on it.

src/__test__/table-theme-colors.spec.js
LiKang6688: Oki

src/__test__/table-theme-colors.spec.js
LiKang6688: Right, fixed those

haxxmaxx:
Screenshot 2022-05-23 at 12 56 35
I don't see the point of having this many describes if there is only one test inside them

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,298 @@
...
+    jest.fn((base, path, attribute) => {
+      let backgroundColor;
+      if (attribute === 'backgroundColor') {
+        backgroundColor = objectBackgroundColor;
+      } else if (attribute === 'object.straightTable.backgroundColor') {
+        backgroundColor = tableBackgroundColor;
+      }
+      return backgroundColor;
+    });
+  const setqvPanelSheetAndqvInnerObject = jest.fn((selector) => selector);
```
haxxmaxx, line 17: why do you need jest.fn? Isn't this just a function?

LiKang6688:
> Screenshot 2022-05-23 at 12 56 35
> 
> I don't see the point of having this many describes if there is only one test inside them

I use that to separate test cases and see that it is easy to read.

src/__test__/table-theme-colors.spec.js
LiKang6688: yes, good point. Got it upated.

src/__test__/table-theme-colors.spec.js
```diff
@@ -0,0 +1,296 @@
+import tableThemeColors from '../table-theme-colors';
+
+describe('table-theme-colors', () => {
+  const spyQuerySelector = jest.fn();
+  const spyGetComputedStyle = jest.fn();
+  const spyGetStyle = jest.fn();
+  const setObjectAndTableBackgroundColorInTheme = (objectBackgroundColor, tableBackgroundColor) =>
+    jest.fn((base, path, attribute) => {
```
haxxmaxx, line 8: same thing here or does it have a purpose?

LiKang6688: same thing. No. 
I missed it. But now it is covered.

haxxmaxx (approved):
I think this is fine, we might have to come back if there are other quirly theme cases

PR #523, opened 2022-05-13
fix: use CreateRoot

haxxmaxx:


src/table/Root.jsx
```diff
@@ -1,27 +1,31 @@
-import ReactDOM from 'react-dom';
+import { createRoot } from 'react-dom/client';
 import React from 'react';
 import { StyleSheetManager } from 'styled-components';
 import { ThemeProvider } from '@mui/material/styles';
 import rtlPluginSc from 'stylis-plugin-rtl-sc';
 import TableWrapper from './components/TableWrapper';
 
+let reactRoot;
```
haxxmaxx, line 8: tried to save as a state in index, but seems like it wont be the same instance. This works fine if it is not considered to hacky

LiKang6688 (approved):

PR #524, opened 2022-05-13
refactor: format code and install deps react-window

LiKang6688:


haxxmaxx:
Are we sure we will use react-window? Don't see the point of importing it if it is not used yet

LiKang6688:
> Are we sure we will use react-window? Don't see the point of importing it if it is not used yet

Not yet.

LiKang6688:
Since we can only vertically visualize the table to avoid breaking a11y, we decided not to adopt react-window.

PR #525, opened 2022-05-16
fix: remove a mounted React component from the DOM

LiKang6688:
https://reactjs.org/docs/react-dom.html#unmountcomponentatnode


haxxmaxx (approved):
ahh that got lost in some stash or something, good find

PR #526, opened 2022-05-17
refactor: put select logic on reducer

haxxmaxx:
This makes way more sense to me, instead of having a seperate function that we need to pass both the state and the dispatcher to.


src/table/components/__tests__/withSelections.spec.jsx
```diff
@@ -68,16 +67,9 @@ describe('withSelections', () => {
     );
     fireEvent.mouseUp(queryByText(value));
 
-    expect(selectionsUtils.selectCell).toHaveBeenCalledTimes(1);
-    expect(selectionsUtils.selectCell).toHaveBeenCalledWith(
```
haxxmaxx, line 72: I don't see the point of checking what has been called with if it is anything. The main point here is to test whether it has been called or not

PR #527, opened 2022-05-17
Crl/selection context

haxxmaxx:


niekvanstaveren (approved):
The code looks good to me. I don't fully understand the logic in the providers, but maybe we can look at that a bit next week? Or maybe even do a show & tell around it

LiKang6688:
https://user-images.githubusercontent.com/4471489/169680994-c08a192d-1b21-46b2-8df3-735e0447f4e5.mov

I came across the issue that when selecting the first-row cell, the color is not correct. The same behavior in the client.

haxxmaxx:
> The code looks good to me. I don't fully understand the logic in the providers, but maybe we can look at that a bit next week? Or maybe even do a show & tell around it

Neither do I. I should probably understand it before merging

haxxmaxx:
> table-selection.mov
> 
> I came across the issue that when selecting the first-row cell, the color is not correct. The same behavior in the client.

Good find, will have a look

src/table/context/TableContext.jsx
```diff
@@ -0,0 +1,42 @@
+import React, { useState, useReducer, createContext } from 'react';
+import PropTypes from 'prop-types';
+import { createSelectorProvider } from './createSelectorProvider';
+import { reducer } from '../utils/selections-utils';
+
+export const TableContext = createContext();
+
+const ProviderWithSelector = createSelectorProvider(TableContext);
+
+export const TableContextProvider = ({ children, selectionsAPI, cellCoordMock, selectionDispatchMock }) => {
```
LiKang6688, line 10: why cellCoordMock and selectionDispatchMock are needed?

haxxmaxx: for testing, need to spy on the selectionDispatch and need to set focusedCellCoord to a different value for a test in the header

src/table/components/PaginationContent.jsx
```diff
@@ -160,11 +159,13 @@ export default function PaginationContent({
       {showFirstAndLast && getButton(onFirstPage, 0, 'FirstPage')}
       {getButton(onFirstPage, page - 1, 'PreviousPage')}
       {getButton(onLastPage, page + 1, 'NextPage', !showFirstAndLast ? handleLastButtonTab : null)}
-      {showFirstAndLast && getButton(onLastPage, lastPageIdx, 'LastPage', handleLastButtonTab)}
+      {showFirstAndLast && getButton(onLastPage, totalPages - 1, 'LastPage', handleLastButtonTab)}
     </>
   );
 }
 
+export default memo(PaginationContent);
```
LiKang6688, line 167: in other files, you put this line below defaultProps, right?

src/table/context/TableContext.jsx
LiKang6688: oki

src/table/components/PaginationContent.jsx
haxxmaxx: you are right, I'll move it

haxxmaxx:
a11y seems to be as before. That being said, there were some quirky stuff before this change. We need to to a proper a11y review, both focus and screen reader

src/table/components/withSelections.jsx
```diff
@@ -1,27 +1,20 @@
...
 
-    const selectionStyling = useMemo(
-      () => getSelectionStyle(styling, cell, selectionState, themeBackgroundColor),
-      [styling, cell, selectionState, themeBackgroundColor]
-    );
+    const handleMouseUp = (evt) => {
+      cell.isDim && evt.button === 0 && selectionDispatch({ type: 'select', payload: { cell, evt, announce } });
+    };
+
+    const selectionStyling = getSelectionStyle(styling, cellSelectionState, themeBackgroundColor);
```
LiKang6688, line 17: why not to use useMemo here?

src/table/utils/handle-key-press.js
```diff
@@ -133,46 +132,49 @@ export const bodyHandleKeyPress = ({
...
       announce({ keys: 'SNTable.SelectionLabel.SelectionsConfirmed' });
       break;
     }
     // Esc: Cancels selections. If no selections, do nothing and handleTableWrapperKeyDown should catch it
     case 'Escape': {
-      if (!isAnalysisMode || !selectionState.api.isModal()) break;
+      if (!isAnalysisMode || !isInSelectionMode) break;
       preventDefaultBehavior(evt);
-      selectionState.api.cancel();
+      selectionsAPI.cancel();
```
LiKang6688, line 171: selectionsAPI?.cancel();

src/table/utils/handle-key-press.js
```diff
@@ -133,46 +132,49 @@ export const bodyHandleKeyPress = ({
...
     case 'ArrowLeft': {
-      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, selectionState, setFocusedCellCoord, announce);
+      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, setFocusedCellCoord, announce, isInSelectionMode);
       break;
     }
     // Space bar: Selects value.
     case ' ': {
       preventDefaultBehavior(evt);
-      cell?.isDim && isAnalysisMode && selectCell({ selectionState, cell, selectionDispatch, evt, announce });
+      cell?.isDim && isAnalysisMode && selectionDispatch({ type: 'select', payload: { cell, evt, announce } });
```
LiKang6688, line 156: why sometimes `selectCell` and sometimes `selectionDispatch({ type: 'select', payload: { cell, evt, announce } })`?

src/table/components/TableWrapper.jsx
```diff
@@ -29,28 +30,26 @@ export default function TableWrapper(props) {
...
-  const setShouldRefocus = () => {
+  const setShouldRefocus = useCallback(() => {
     shouldRefocus.current = rootElement.getElementsByTagName('table')[0].contains(document.activeElement);
-  };
-
-  const handleChangePage = (pageIdx) => {
-    setPageInfo({ ...pageInfo, page: pageIdx });
-    announce({ keys: [['SNTable.Pagination.PageStatusReport', [pageIdx + 1, totalPages]]], politeness: 'assertive' });
-  };
+  }, [rootElement]);
```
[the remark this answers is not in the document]
haxxmaxx: exactly. If these are created each time `TableWrapper` renders, the `TableBody` and `PaginatioContent` can't be memoed.

src/table/components/withSelections.jsx
haxxmaxx: because it is low cost to run, only a switch returning the style. We shouldn't use useMemo if it doesn't give any value

src/table/utils/handle-key-press.js
```diff
@@ -133,46 +132,49 @@ export const bodyHandleKeyPress = ({
...
       break;
     }
     // Enter: Confirms selections.
     case 'Enter': {
       preventDefaultBehavior(evt);
-      if (!selectionState.api.isModal()) break;
-      selectionState.api.confirm();
+      if (!isInSelectionMode) break;
+      selectionsAPI.confirm();
```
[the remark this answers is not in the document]
haxxmaxx: In that case `isInSelectionMode` is false and it would break on line 162

src/table/utils/handle-key-press.js
haxxmaxx: same  here, would break earlier

src/table/utils/handle-key-press.js
haxxmaxx: `selectCell` is only called inside the reducer. That logic could live in the reducer itself, but I kept the separate function to not clutter the reducer

src/table/utils/selections-utils.js
```diff
@@ -63,52 +71,67 @@ export const handleAnnounceSelectionStatus = ({ announce, rowsLength, isAddition
...
 
-export const getSelectedRows = ({ selectedRows, qElemNumber, rowIdx, evt }) => {
+export const getSelectedRows = ({ selectedRows, cell, evt }) => {
+  const { qElemNumber, rowIdx } = cell;
   if (evt.ctrlKey || evt.metaKey) {
     // if the ctrl key or the ⌘ Command key (On Macintosh keyboards) or the ⊞ Windows key is pressed
-    // get the last clicked item
-    return [{ qElemNumber, rowIdx }];
+    // get the last clicked item (single select)
+    return { [qElemNumber]: rowIdx };
```
LiKang6688, line 79: `[qElemNumber]: rowIdx` is new to me. what will it look like?

src/table/components/withSelections.jsx
haxxmaxx: It might even be the case that the checks react needs to do for the dependency array (at least 2 checks) is more calculations than actually running `getSelectionStyling`

LiKang6688: oki

src/table/utils/handle-key-press.js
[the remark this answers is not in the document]
LiKang6688: right

LiKang6688:
Screenshot 2022-05-24 at 11 35 36
I tested for theme color, in the mashup, the background color is not correct when doing the selection.

LiKang6688:
> Screenshot 2022-05-24 at 11 35 36
> 
> I tested for theme color, in the mashup, the background color is not correct when doing the selection.

NVM, it is not correct in the main branch.

LiKang6688 (approved):
Awesome 👍🚀

PR #529, opened 2022-05-18
fix: avoid conflicts when using multiple roots on the same page

LiKang6688:
https://reactjs.org/docs/react-dom-client.html#createroot

identifierPrefix: optional prefix React uses for ids generated by React.useId. Useful to avoid conflicts when using multiple roots on the same page. Must be the same prefix used on the server.

We either revert back to react 17 or add useId in nebula.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,6 +14,7 @@ import {
   useKeyboard,
   useRect,
   useApp,
+  useId,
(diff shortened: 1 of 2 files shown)


LiKang6688:
an easier way to fix the issue on https://github.com/qlik-oss/sn-table/pull/530

PR #530, opened 2022-05-18
fix: avoid conflicts when using multiple roots on the same page

LiKang6688:
save reactRoot as a state 
create a new reactRoot when there is a new element to render on


src/index.js
```diff
@@ -76,8 +81,9 @@ export default function supernova(env) {
         if (layout && tableData && !env.carbon) {
           registerLocale(translator);
           const changeSortOrder = sortingFactory(model);
-          render(rootElement, {
+          render({
             rootElement,
+            reactRoot,
```
haxxmaxx, line 86: Don't put it in this object, no need to pass it to the rest of the app

src/index.js
```diff
@@ -25,8 +27,12 @@ import { mount, render, teardown } from './table/Root';
 import muiSetup from './mui-setup';
 import tableThemeColors from './table-theme-colors';
 
+const initialPageInfo = () => ({
```
haxxmaxx, line 30: This could just be an object

src/table/Root.jsx
```diff
@@ -1,19 +1,16 @@
-import { createRoot } from 'react-dom/client';
 import React from 'react';
 import { StyleSheetManager } from 'styled-components';
 import { ThemeProvider } from '@mui/material/styles';
 import rtlPluginSc from 'stylis-plugin-rtl-sc';
 import TableWrapper from './components/TableWrapper';
 
-let reactRoot;
+let root;
```
haxxmaxx, line 7: Do we need this? Can't we pass it to teardown as well? Might as well avoid global variables if we can

src/index.js
LiKang6688: Fixed

src/index.js
LiKang6688: Sure. Done

src/table/Root.jsx
LiKang6688: No, we do not. Good idea. Done

src/index.js
```diff
@@ -95,6 +99,7 @@ export default function supernova(env) {
           });
         }
       }, [
+        rootElement,
```
haxxmaxx, line 102: If anything, this should have the reactRoot as dependency, otherwise we could still be passing the old reactRoot

LiKang6688: good find. I mistakenly remove the line.

haxxmaxx: you wont need rootElement though since if `rootElement` changes `reactRoot` will change too

LiKang6688: that's true. Updated.

haxxmaxx (approved):

PR #531, opened 2022-05-19
fix: unmount react root

LiKang6688:


src/index.js
```diff
@@ -117,7 +117,7 @@ export default function supernova(env) {
 
       useEffect(
         () => () => {
-          teardown(reactRoot);
+          teardown();
         },
         []
       );
```
LiKang6688, line 123: it is mind blowing that
```
 useEffect(() => {
        const root = reactRoot;

        return () => {
          teardown(root);
        };
      }, [reactRoot]);
```
though reactRoot is a global variable, but it is undefined here.

LiKang6688: Nope
```
 useEffect(() => {
        let root;
        if (reactRoot) root = reactRoot;

        return () => {
          teardown(root);
        };
      }, [reactRoot]);
```

LiKang6688: Now I know that reactRoot is undefined because 
```
return () => {
          teardown(root);
        };
```
not only runs when sn-table wants to unmount, but also runs before the execution of the next scheduled effect.
so reactRoot is unmounted and then is undefined
Reference:
https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/#:~:text=What%20is%20the%20useEffect%20cleanup,itself%20using%20the%20cleanup%20function.

PR #535, opened 2022-05-23
refactor: unmount react root

LiKang6688:


haxxmaxx:
I thought this didn't work?

LiKang6688:
> I thought this didn't work?



```
 useEffect(() => {
        let root;
        if (reactRoot) root = reactRoot;

        return () => {
          teardown(root);
        };
      }, [reactRoot]);
```
did not work, it should be
```
useEffect(() => {
        return () => {
         reactRoot && teardown(root);
        };
      }, [reactRoot]);
```

haxxmaxx (approved):

LiKang6688:
reactRoot is undefined because
 
return () => {
          teardown(root);
        };
 
not only runs when sn-table wants to unmount, but also runs before the execution of the next scheduled effect.
so reactRoot is unmounted and then is undefined
Reference:
https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/#:~:text=What%20is%20the%20useEffect%20cleanup,itself%20using%20the%20cleanup%20function.

PR #536, opened 2022-05-23
feat: add totals function to sn-table

Donya-qlik:
Add totals functions to sn-table:
- totalsAggrFunction ( Auto, min, max, ...)
- showtotals ( none, top, buttom ) + total label
- Unit tests
(includes DQ unsupported 'totals' feature check)
image
image


src/ext.js
```diff
@@ -97,230 +99,349 @@ const textAlignItems = {
...
+              },
+              {
+                ref: 'totals.label',
+                translation: 'properties.totals.label',
+                type: 'string',
+                expression: 'optional',
+                defaultValue: 'Totals',
+              },
+            ],
+            show: !env?.anything?.sense?.isUnsupportedFeature?.('totals'),
```
Donya-qlik, line 282: These properties are the new changes for totals aggr functions. The other changes ( except label that I've marked it) are for refactoring purposes and there is no specific change on the properties.

src/ext.js
```diff
@@ -97,230 +99,349 @@ const textAlignItems = {
...
+              },
             },
+            show: !env?.anything?.sense?.isUnsupportedFeature?.('totals'),
           },
         },
       },
     },
-    settings: {
-      uses: 'settings',
+  };
```
Donya-qlik, line 418: The presentation, total functions. They are added via the current PR, the other changes are mostly for refactoring purposes which was required.

src/ext.js
Donya-qlik, line 282: I refactored the current file because it was hard to follow the order of properties, to make it more readable but my changes are basically focused only on `totals-show `( lines 234 - 282) and `totals aggregation function` ( lines 355- 413).

src/table/utils/styling-utils.js
```diff
@@ -150,6 +150,20 @@ export function getBodyCellStyle(layout, theme) {
...
+};
+
+export function showTotals(layout) {
+  if (isTotalModeAuto(layout) && hasOnlyMeasure(layout)) return false;
+  return hasTotalsMeasures(layout) && layout.totals && (layout.totals.show || layout.totals.position !== 'noTotals');
+}
+
+export function showTotalsAtTop(layout) {
+  return !(showTotals(layout) && layout.totals && !layout.totals.show && layout.totals.position === 'bottom');
+}
```
haxxmaxx, line 166: This doesn't really relate to styling. I think it would make sense to be a seperate file.

src/ext.js
haxxmaxx: Try to keep big refactors like this out of feature PRs. Plus this adds three issues on codeclimate. This could still be a good idea but lets keep it in a seperate PR

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,107 @@
...
+      totalsHeader = layout.qHyperCube.qGrandTotalRow[measureIndex].qText;
+      measureIndex += 1;
+    }
+    return totalsHeader;
+  };
+
+  return (
+    <TableRow sx={totalRowStyle()} className="sn-table-row">
+      {columns.map((column, columnIndex) => {
+        const tabIndex = columnIndex === 0 && !keyboard.enabled ? 0 : -1;
```
haxxmaxx, line 65: This will cause some issues. It is based on the table header only being one row, so it should add tabindex 0 only to the top left cell. This will add two, so you can set the tabindex to always be -1

src/table/components/TableHeadWrapper.jsx
```diff
@@ -100,6 +103,7 @@ function TableHeadWrapper({
           );
         })}
       </TableRow>
+      {showTotals(layout) && showTotalsAtTop(layout) && <TableTotals {...props} />}
```
haxxmaxx, line 106: there is a nice trick you can do to avoid the props drilling here. Bit hard to explain I can show you tomorrow

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,107 @@
...
+      borderRight: paginationNeeded && 0,
+    },
+    'th:first-of-type': {
+      borderLeft: !paginationNeeded && '1px solid rgb(217, 217, 217)',
+    },
+  };
+
+  const totalRowStyle = () => {
+    if (showTotalsAtTop(layout)) return [headRowStyle, { position: 'sticky', top: 35 }];
+    return [headRowStyle, { position: 'sticky', bottom: 0 }];
```
haxxmaxx, line 41: Why an array with two objects? btw do you need to add sticky here? how does the normal header stay sticky?

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,107 @@
...
+          ? layout.totals.label !== undefined
+            ? layout.totals.label
+            : translator.get('Object.Table.Totals')
+          : '\u00A0';
+    }
+    if (!column.isDim) {
+      totalsHeader = layout.qHyperCube.qGrandTotalRow[measureIndex].qText;
+      measureIndex += 1;
+    }
+    return totalsHeader;
```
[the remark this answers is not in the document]
haxxmaxx: you could also put a `headerPosition` prop on the `tableData` object. That way we can skip the function calls from the components as well

LiKang6688:
Where do you include DQ unsupported 'totals' feature check?

src/ext.js
```diff
@@ -97,230 +99,349 @@ const textAlignItems = {
...
+                          translation: 'Object.Table.AggrFunc.None',
+                        },
+                      ]);
+                    },
+                    defaultValue: env?.anything?.sense?.isUnsupportedFeature?.('totals') ? 'None' : 'Expr',
+                  },
+                },
+              },
             },
+            show: !env?.anything?.sense?.isUnsupportedFeature?.('totals'),
```
Donya-qlik, line 413: Here @LiKang6688, and line 408 and one at line 282.

LiKang6688: I see. Thanks.

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,107 @@
...
+        const tabIndex = columnIndex === 0 && !keyboard.enabled ? 0 : -1;
+
+        return (
+          <TableCell
+            sx={totalCellStyle()}
+            key={column.id}
+            align={column.align}
+            tabIndex={tabIndex}
+            onKeyDown={(e) =>
+              headHandleKeyPress(
```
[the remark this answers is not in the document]
Donya-qlik: Should I keep onKeyDown for Totals at all ?

src/ext.js
Donya-qlik: Sure! Agree!  It's a quite big refactoring so I'll revert it to the original orders as soon as I finish with the other comments.

src/table/utils/handle-key-press.js
```diff
@@ -129,6 +130,27 @@ export const headHandleKeyPress = (
...
+//   switch (evt.key) {
+//     case 'ArrowRight':
+//     case 'ArrowLeft': {
+//       !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, false, setFocusedCellCoord);
+//       break;
+//     }
+//     default:
+//       break;
+//   }
+// };
```
Donya-qlik, line 152: I'll remove this...

src/table/components/TableTotals.jsx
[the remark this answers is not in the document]
haxxmaxx: well yes you would need to be able to navigate from and to these cells. You might need to create a new function for it thought, since the only thing you are suppose to do is move, they can't be used for selecting or sorting

src/handle-data.js
```diff
@@ -31,11 +39,28 @@ export function getColumnInfo(qHyperCube, colIndex) {
...
+  return true;
+}
+
+export function getTotalPosition(layout) {
+  if (!showTotals(layout)) return 'noTotals';
+  const isTotalModeAuto = () => layout.totals?.show === true;
+  if (isTotalModeAuto() || (!isTotalModeAuto() && layout.totals.position === 'top')) return 'top';
+  if (!isTotalModeAuto() && layout.totals.position === 'bottom') return 'bottom';
+  return 'noTotals';
+}
```
haxxmaxx, line 61: THis could be one function since you don't need showTotals anywhere

src/handle-data.js
```diff
@@ -31,11 +39,28 @@ export function getColumnInfo(qHyperCube, colIndex) {
...
+      totalInfo,
     }
   );
 }
 
-export default async function manageData(model, layout, pageInfo, setPageInfo) {
+export function showTotals(layout) {
+  const hasOnlyMeasure = () => layout.qHyperCube.qDimensionInfo.length === 0;
+  const hasTotalsMeasures = () =>
+    layout.qHyperCube.qMeasureInfo.length > 0 && layout.qHyperCube.qGrandTotalRow.length > 0;
```
haxxmaxx, line 50: No need for these to be functions, should just get the value

src/handle-data.js
```diff
@@ -31,11 +39,28 @@ export function getColumnInfo(qHyperCube, colIndex) {
...
+  if (hasOnlyMeasure() || !hasTotalsMeasures()) return false;
+  return true;
+}
+
+export function getTotalPosition(layout) {
+  if (!showTotals(layout)) return 'noTotals';
+  const isTotalModeAuto = () => layout.totals?.show === true;
```
haxxmaxx, line 57: this shouldn't be a function

haxxmaxx: also a comment about that totals.true is actually the auto setting, this confused me a lot

src/handle-data.js
```diff
@@ -15,12 +15,20 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
...
   const { qDimensionInfo, qMeasureInfo } = qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIndex < numDims;
   const info = isDim ? qDimensionInfo[colIndex] : qMeasureInfo[colIndex - numDims];
   const isHidden = info.qError?.qErrorCode === 7005;
+  const totalInfo = isDim
+    ? colIndex === 0
+      ? layout.totals.label !== undefined
+        ? layout.totals.label
+        : translator.get('Object.Table.Totals')
```
haxxmaxx, line 28: this will be translated later
```suggestion
        : 'Object.Table.Totals'
```

src/table/utils/styling-utils.js
```diff
@@ -150,6 +150,21 @@ export function getBodyCellStyle(layout, theme) {
   };
 }
 
+export function getTotalsCellStyle(headerStyle, layout) {
+  if (getTotalPosition(layout) === 'noTotals') return false;
+  headerStyle.borderColor = 'rgb(166, 166, 166)';
+  headerStyle.fontWeight = 'bold';
```
haxxmaxx, line 156: Not a great idea changing properties on an input argument, make sure you desctruct so you don't change the original `headerStyle`

src/table/components/TableWrapper.jsx
```diff
@@ -25,6 +28,7 @@ export default function TableWrapper(props) {
     translator,
     selectionsAPI,
     theme,
+    layout,
```
haxxmaxx, line 31: make sure to remove layout here

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,74 @@
...
+                rootElement,
+                [0, columnIndex],
+                column,
+                changeSortOrder,
+                layout,
+                !constraints.active,
+                setFocusedCellCoord
+              )
+            }
+            onMouseDown={() => handleClickToFocusHead(columnIndex, rootElement, setFocusedCellCoord, keyboard)}
```
haxxmaxx, line 52: clicking the head will sort the table. Clicking the totals should just focus it

src/handle-data.js
```diff
@@ -77,5 +104,5 @@ export default async function manageData(model, layout, pageInfo, setPageInfo) {
     return row;
   });
 
-  return { totalColumnCount, totalRowCount, paginationNeeded, rows, columns };
+  return { totalColumnCount, totalRowCount, paginationNeeded, rows, columns, translator };
```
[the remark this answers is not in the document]
Donya-qlik: Good idea! 👍

src/table/components/TableHeadWrapper.jsx
```diff
@@ -21,19 +23,20 @@ const VisuallyHidden = styled('span')({
...
-  changeSortOrder,
-  constraints,
-  translator,
-  selectionsAPI,
-  focusedCellCoord,
-  setFocusedCellCoord,
-  keyboard,
-}) {
+function TableHeadWrapper(props) {
+  const {
```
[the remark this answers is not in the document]
Donya-qlik: ...but I'm passing it to TableTotal down in the same file, should I pass them as an object argument instead ?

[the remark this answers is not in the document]
haxxmaxx: ahh sorry my bad, yes spread props as you do now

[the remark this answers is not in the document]
haxxmaxx: now I remember, I suggested this before I think. You don't need to add `TableTotals` in here. you can do the same as with `FooterWrapper`. So you do
```
          <TableHeadWrapper {...props}>
            {totalsPosition === 'top' && <TableTotals {...props} />}
          </TableHeadWrapper>
```
inside `TableWrapper` and then just put `{children}` on line 106 in here

src/table/components/TableTotals.jsx
haxxmaxx: not true, sorry. This will only handle focus, but the row index will be wrong. The row index for focus in general will be a bit tricky since it depends on how many header rows there are, that needs to be taken into account

src/table/utils/styling-utils.js
Donya-qlik: To not to get any conflict I created specific function for totalstyle

src/table/components/TableHeadWrapper.jsx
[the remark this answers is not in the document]
Donya-qlik: I ended up to add a `TableHead` and define both headerWrapper and `Tabletotals `as the children. It's more readable now.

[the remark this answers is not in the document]
Donya-qlik: I added a `TableHead` and set these two as its children, I think it is more readable now.

src/handle-data.js
```diff
@@ -15,12 +15,20 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
   return qColumnOrder?.length === columnsLength ? qColumnOrder : [...Array(columnsLength).keys()];
 }
 
-export function getColumnInfo(qHyperCube, colIndex) {
+export function getColumnInfo(qHyperCube, colIndex, layout) {
```
haxxmaxx, line 18: `qHyperCube` exists on `layout`, only pass layout

src/handle-data.js
```diff
@@ -31,10 +39,31 @@ export function getColumnInfo(qHyperCube, colIndex) {
...
+  const hasOnlyMeasure = layout.qHyperCube.qDimensionInfo.length;
+  const hasGrandTotal = layout.qHyperCube.qGrandTotalRow.length;
+  const hasMeasure = layout.qHyperCube.qMeasureInfo.length;
+  const isTotalModeAuto = layout.totals?.show;
+
+  if (hasOnlyMeasure === 0 || !(hasMeasure > 0 && hasGrandTotal > 0)) showTotal = false;
+  if (!showTotal) return 'noTotals';
+  if (isTotalModeAuto || (!isTotalModeAuto && layout.totals.position === 'top')) return 'top';
+  if (!isTotalModeAuto && layout.totals.position === 'bottom') return 'bottom';
+  return 'noTotals';
```
haxxmaxx, line 64: The names doesn' really make sense here. `has...` variables should be booleans by them selves. try this:

```suggestion
  const hasOnlyMeasure = layout.qHyperCube.qDimensionInfo.length === 0;
  const hasGrandTotal = layout.qHyperCube.qGrandTotalRow.length > 0;
  const hasMeasure = layout.qHyperCube.qMeasureInfo.length > 0;
  const isTotalModeAuto = layout.totals?.show;

  if (hasOnlyMeasure || !(hasMeasure && hasGrandTotal )) return 'noTotals';
  if (isTotalModeAuto || (!isTotalModeAuto && layout.totals.position === 'top')) return 'top';
  if (!isTotalModeAuto && layout.totals.position === 'bottom') return 'bottom';
  return 'noTotals';
```

src/object-properties.js
```diff
@@ -55,6 +55,15 @@ const properties = {
...
   footnote: '',
+  /**
+   * totals settings
+   * @type {object}
+   */
+  totals: {
+    show: true,
+    position: 'noTotals',
+    label: 'Totals',
+  },
```
haxxmaxx, line 66: @niekvanstaveren How do we want to do this? If we add theme here for now, do we need to add it to the spec as well, at least before releasing?

src/table/components/TableWrapper.jsx
```diff
@@ -152,13 +154,23 @@ export default function TableWrapper(props) {
...
         <Table stickyHeader aria-label={tableAriaLabel}>
-          <TableHeadWrapper {...props} />
+          <TableHead>
+            <TableHeadWrapper
+              {...props}
+              setFocusedCellCoord={setFocusedCellCoord}
+              focusedCellCoord={focusedCellCoord}
+            />
+            {totalsPosition === 'top' && <TableTotals {...props} setFocusedCellCoord={setFocusedCellCoord} />}
+          </TableHead>
```
haxxmaxx, line 164: Sorry I don't like this, having the `TableHead` component wrap the `TableHeadWrapper` doesn't make sense. I'd say move  `TableHead` back to the other file

src/table/components/TableWrapper.jsx
```diff
@@ -152,13 +154,23 @@ export default function TableWrapper(props) {
...
           <TableBodyWrapper
             {...props}
             announce={announce}
             setShouldRefocus={setShouldRefocus}
             tableWrapperRef={tableWrapperRef}
           />
+          <TableFooter>
+            {totalsPosition === 'bottom' && <TableTotals {...props} setFocusedCellCoord={setFocusedCellCoord} />}
+          </TableFooter>
```
haxxmaxx, line 173: We don't want the footer at all if there is no totals here
```suggestion
          {totalsPosition === 'bottom' && (<TableFooter>
            <TableTotals {...props} setFocusedCellCoord={setFocusedCellCoord} />
          </TableFooter>)}
```

src/table/components/TableWrapper.jsx
```diff
@@ -152,13 +154,23 @@ export default function TableWrapper(props) {
...
       >
         <Table stickyHeader aria-label={tableAriaLabel}>
-          <TableHeadWrapper {...props} />
+          <TableHead>
+            <TableHeadWrapper
+              {...props}
+              setFocusedCellCoord={setFocusedCellCoord}
+              focusedCellCoord={focusedCellCoord}
+            />
+            {totalsPosition === 'top' && <TableTotals {...props} setFocusedCellCoord={setFocusedCellCoord} />}
```
haxxmaxx, line 163: setFocusedCellCoord should not be passed as a props since it exists on the `useContextSelector`. use That instead inside the totals function

src/table/utils/handle-key-press.js
```diff
@@ -76,7 +76,6 @@ export const getRowAndColumnCount = (rootElement) => {
 
   const headCellElements = rootElement.getElementsByClassName('sn-table-head-cell');
   const columnCount = headCellElements.length;
-
```
haxxmaxx, line 79: why?

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,60 @@
...
+import TableRow from '@mui/material/TableRow';
+import TableCell from '@mui/material/TableCell';
+import { useContextSelector, TableContext } from '../context';
+import { getHeaderStyle, getTotalsCellStyle } from '../utils/styling-utils';
+import { totalHeadHandleKeyPress } from '../utils/handle-key-press';
+import { handleClickToFocusHead } from '../utils/handle-accessibility';
+
+function TableTotals({ rootElement, tableData, theme, layout, translator, keyboard }) {
+  const { columns, paginationNeeded, totalsPosition } = tableData;
+  const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
```
haxxmaxx, line 12: Ah you are already using it here, good :) just remove the pass you prop in tableWrapper then

src/table/utils/styling-utils.js
```diff
@@ -153,6 +153,28 @@ export function getBodyCellStyle(layout, theme) {
...
 
+/**
+ * Get total head cell style
+ *
+ * @param {Object} totalHeaderStyle
+ * @param {Object} layout
+ * @returns total cell style regarding the total head position
+ */
+export function getTotalsCellStyle(totalHeaderStyle, layout) {
+  if (getTotalPosition(layout) === 'noTotals') return false;
```
haxxmaxx, line 164: better to pass the position from totals component, you should have access to it there

haxxmaxx: then you don't need layout

src/table/components/TableTotals.jsx
haxxmaxx: This one sets the rowIdx to 0, that would be wrong. So need to pass the actual cell to `clickToFocusBody`. Should proably be renamed `clickToFocusCell` if it is used here

src/handle-data.js
Donya-qlik: Ah, you're right! It was there,  I see I missed to remove it after adding layout

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,60 @@
...
+            key={column.id}
+            align={column.align}
+            className="sn-table-cell"
+            tabIndex={-1}
+            onKeyDown={(e) => {
+              let cellCoord;
+              if (totalsPosition === 'bottom') {
+                const rowIndex = rootElement.getElementsByClassName('sn-table-row').length;
+                cellCoord = [rowIndex - 1, columnIndex];
+              } else cellCoord = [1, columnIndex];
```
haxxmaxx, line 38: the amount of rows already exist on pageInfo. it should be the `rowsPerPage` property. You can use that to simplify this:
```suggestion
              const cellCoord = [totalsPosition === 'bottom' ? rowsPerPage : 1, columnIndex];
```
Might need `rowsPerPage + 1` to  or something for the number to be correct

src/object-properties.js
niekvanstaveren: Yes I think so

src/table/utils/handle-key-press.js
```diff
@@ -128,6 +127,28 @@ export const headHandleKeyPress = (
...
 
+/**
+ * Handle key press for Totals in table head
+ *
+ * @param {event} evt
+ * @param {Object} rootElement
+ * @param {Array} cellCoord
+ * @param {Function} setFocusedCellCoord
+ */
+export const totalHeadHandleKeyPress = (evt, rootElement, cellCoord, setFocusedCellCoord) => {
```
haxxmaxx, line 138: should probably be `totalHandleKeyPress` since it is not always in the header

src/handle-data.js
haxxmaxx: btw `if (hasOnlyMeasure || !(hasMeasure && hasGrandTotal )) return 'noTotals';` is the same as `if (hasOnlyMeasure || !hasMeasure || !hasGrandTotal) return 'noTotals';`

src/handle-data.js
```diff
@@ -31,10 +39,31 @@ export function getColumnInfo(qHyperCube, colIndex) {
...
   );
 }
 
+/**
+ * Get the total head position of the table
+ *
+ * @param {Object} layout
+ * @returns the position as a string, it can be any of top, bottom or noTotals
+ */
+export function getTotalPosition(layout) {
```
haxxmaxx, line 53: this function should have unit tests btw

src/handle-data.js
```diff
@@ -15,12 +15,20 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
...
 }
 
-export function getColumnInfo(qHyperCube, colIndex) {
+export function getColumnInfo(qHyperCube, colIndex, layout) {
   const { qDimensionInfo, qMeasureInfo } = qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIndex < numDims;
   const info = isDim ? qDimensionInfo[colIndex] : qMeasureInfo[colIndex - numDims];
   const isHidden = info.qError?.qErrorCode === 7005;
+  const totalInfo = isDim
```
haxxmaxx, line 24: you need tests for this part as well

src/table/utils/styling-utils.js
```diff
@@ -153,6 +153,28 @@ export function getBodyCellStyle(layout, theme) {
...
+  totalHeaderStyle.borderColor = 'rgb(166, 166, 166)';
```
haxxmaxx, line 165: I don't think you want to change this, since the border color depends on the theme

haxxmaxx: no not if you do it at this stage. What comes from `getHeaderStyle` is the final style, theme taken into account

src/handle-data.js
haxxmaxx: great :)

src/table/components/TableTotals.jsx
haxxmaxx: yes, you are right sorry. But the information is on `tableData`. rows.length should give you the current amount of rows showing

src/table/components/TableTotals.jsx
haxxmaxx: when you click, it sets the focus to the cell you clicked. it takes a coordinate, blurs the previous focused cell, and focus the one at given coordinate + updates the focus state. What you need to do is to figure out the correct coordinate as you do for `totalHandleKeyPress`

Tabbing has a role in the table. If you have focus on a header or body cell, tab will take you to the pagination controls, the you tab through those since the are buttons/selects.

src/table/components/TableWrapper.jsx
haxxmaxx: Let me check out your branch and see if my approach works

src/table/components/TableWrapper.jsx
```diff
@@ -152,13 +154,25 @@ export default function TableWrapper(props) {
         data-testid="table-container"
       >
         <Table stickyHeader aria-label={tableAriaLabel}>
-          <TableHeadWrapper {...props} />
+          <TableHead>
+            <TableHeadWrapper
+              {...props}
+              setFocusedCellCoord={setFocusedCellCoord}
+              focusedCellCoord={focusedCellCoord}
```
haxxmaxx, line 161: These should be removed as well

src/table/components/TableWrapper.jsx
haxxmaxx: It works, I have a commit I can push if you want. Sorry if I didn't explain it well enough. Yes, there is a reason. `TableHeader` adds no information to `TableTotals` component. Everything it needs exists inside `TableWrapper`. So, to reduce prop drilling (we want to keep it at the `TableWrapper` level, not lower) and keep the show logic in TableWrapper, for both top and bottom.

This is a bit long, but explains prop drilling and ways to solve it. What we are doing here is Container components, the second way described in the article

src/table/utils/handle-key-press.js
```diff
@@ -128,6 +128,28 @@ export const headHandleKeyPress = (
   }
 };
 
+/**
+ * Handle key press for Totals in table head
```
haxxmaxx, line 132: description is still wrong

src/table/utils/styling-utils.js
```diff
@@ -153,6 +152,27 @@ export function getBodyCellStyle(layout, theme) {
   };
 }
 
+/**
+ * Get total head cell style
```
haxxmaxx, line 156:
```suggestion
 * Get style for total cell
```

src/table/components/TableTotals.jsx
haxxmaxx: yeah but you know that you have one header row and one totals row, so just add those

src/table/components/TableWrapper.jsx
haxxmaxx: yes and then you add children (which is on the props now) at the end of the `TableHeadWrapper` component here
```
      </TableRow>
      {children}
    </TableHead>

File: src/handle-data.js [hunk 37]
```diff
@@ -31,10 +39,31 @@ export function getColumnInfo(qHyperCube, colIndex) {
...
+  const hasOnlyMeasure = layout.qHyperCube.qDimensionInfo.length === 0;
+  const hasGrandTotal = layout.qHyperCube.qGrandTotalRow.length > 0;
+  const hasMeasure = layout.qHyperCube.qMeasureInfo.length > 0;
+  const isTotalModeAuto = layout.totals?.show;
+
+  if (hasOnlyMeasure || !(hasMeasure && hasGrandTotal)) showTotal = false;
+  if (!showTotal) return 'noTotals';
+  if (isTotalModeAuto || (!isTotalModeAuto && layout.totals.position === 'top')) return 'top';
+  if (!isTotalModeAuto && layout.totals.position === 'bottom') return 'bottom';
+  return 'noTotals';
```
haxxmaxx, line 64: as I said, you don't need the showTotal variable at all. This would be even niced actually:
```suggestion
  const hasDimension = layout.qHyperCube.qDimensionInfo.length > 0;
  const hasGrandTotal = layout.qHyperCube.qGrandTotalRow.length > 0;
  const hasMeasure = layout.qHyperCube.qMeasureInfo.length > 0;
  const isTotalModeAuto = layout.totals?.show;

  if (hasDimension && hasMeasure && hasGrandTotal) {
    if (isTotalModeAuto || (!isTotalModeAuto && layout.totals.position === 'top')) return 'top';
    if (!isTotalModeAuto && layout.totals.position === 'bottom') return 'bottom';
  }
  return 'noTotals';
```

src/table/components/TableTotals.jsx
haxxmaxx: and this would be the same rowIdx for both `totalHandleKeyPress` and `handleClickToFocusCell`

src/table/utils/handle-accessibility.js
```diff
@@ -38,8 +38,20 @@ export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, k
   removeAndFocus([rawRowIdx + 1, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
 };
 
-export const handleClickToFocusHead = (columnIndex, rootElement, setFocusedCellCoord, keyboard) => {
-  removeAndFocus([0, columnIndex], rootElement, setFocusedCellCoord, keyboard);
+export const handleClickToFocusCell = (
```
haxxmaxx, line 41: hmm sorry if I was unclear, it is the body `handleClickToFocusBody` that we want to rewrite. But I just realized `removeAndFocus` does exactly what you want. If you calculate the `newCoord` inside the component you can use it as is

src/handle-data.js
Donya-qlik: Yes, it is not needed also something I've changed here that seems I was wrong about it. When the table has only measure and it is not in auto mode the totals still are shown. This is how it works on the old table. I've fixed it in my coming commit.

haxxmaxx: ok, good find!

src/table/utils/handle-accessibility.js
Donya-qlik: Yes removeAndFocus is the one has been used in all of them. I'll refactor this one by rows.length in Total component.

src/table/components/TableWrapper.jsx
Donya-qlik: @Max now I'm confused... Where exactly you want this component to be? inside `TableHeadWrapper.jsx` or inside `TableWrapper.jsx` ?
I know it works if I set it at the end of `TableHeadWrapper` as a child, it's where I had put it from the very beginning but then you said it is preferred to be set inside `TableWrapper.jsx` instead or I got it totally wrong?

src/table/utils/styling-utils.js
```diff
@@ -153,6 +152,27 @@ export function getBodyCellStyle(layout, theme) {
...
 
+/**
+ * Get style for total cell
+ *
+ * @param {Object} totalHeaderStyle
+ * @param {String} totalsPosition
+ * @returns total cell style regarding the total head position
+ */
+export function getTotalsCellStyle(totalHeaderStyle, totalsPosition) {
+  if (totalsPosition === 'noTotals') return false;
```
niekvanstaveren, line 163: Is this possible?

niekvanstaveren (commented):
Nice work @Donya-qlik !
Would like to see some screen shots of this attached and we should definitely test the accessibility parts of this. Will leave it to Max And Li to do a full review

src/handle-data.js
Donya-qlik: I'm not fan of them either... but why not ?

niekvanstaveren: The moment you start nesting them it becomes very hard to read.

src/table/utils/styling-utils.js
Donya-qlik: Are you asking about the condition ? If so, yes. There are some conditions that total row is not shown ( when it is set to be none or table has only dimension and so on)

src/handle-data.js
```diff
@@ -15,12 +15,20 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
   return qColumnOrder?.length === columnsLength ? qColumnOrder : [...Array(columnsLength).keys()];
 }
 
-export function getColumnInfo(qHyperCube, colIndex) {
-  const { qDimensionInfo, qMeasureInfo } = qHyperCube;
+export function getColumnInfo(colIndex, layout) {
```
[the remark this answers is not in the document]
Donya-qlik: It's fixed now.

src/handle-data.js
```diff
@@ -5,6 +5,21 @@ const directionMap = {
...
+ * @param {Object} layout
+ * @param {Number} colIndex
+ * @param {Number} numDims
+ * @returns dimensions and measures total cell values as strings
+ */
+export function getTotalInfo(isDim, layout, colIndex, numDims) {
+  if (!isDim) return layout.qHyperCube.qGrandTotalRow[colIndex - numDims]?.qText;
+  if (colIndex !== 0) return '\u00A0';
+  if (layout.totals.label !== undefined && layout.totals.label) return layout.totals.label;
+  return 'Totals';
```
Donya-qlik, line 20: @Max when the condition reaches to line 20, it supposed to return `translator.get('Object.Tabl.Totals')` , but without translator it just returns the input as a text. I just replace it with Totals but not sure if it's a good idea ....

haxxmaxx: It should be `'Object.Tabl.Totals'` here, then it will be translated later in the totals component

src/table/utils/styling-utils.js
haxxmaxx: this is inconsistent with what your description says. And we would have to have two different return types when converting to typescript. You could just return `totalHeaderStyle` as is insteaad

src/table/utils/handle-accessibility.js
```diff
@@ -38,8 +38,9 @@ export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, k
   removeAndFocus([rawRowIdx + 1, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
 };
 
-export const handleClickToFocusHead = (columnIndex, rootElement, setFocusedCellCoord, keyboard) => {
-  removeAndFocus([0, columnIndex], rootElement, setFocusedCellCoord, keyboard);
+export const handleClickToFocusCell = (columnIndex, rootElement, setFocusedCellCoord, keyboard, cellCoord) => {
+  const isTotalCell = rootElement.getElementsByClassName('sn-table-total-cell').length > 0;
+  removeAndFocus((isTotalCell && cellCoord) || [0, columnIndex], rootElement, setFocusedCellCoord, keyboard);
```
haxxmaxx, line 43: you shouldn't change this function at all. You can use as is in your component

src/handle-data.js
```diff
@@ -5,6 +5,21 @@ const directionMap = {
...
+ * @param {Boolean} isDim
+ * @param {Object} layout
+ * @param {Number} colIndex
+ * @param {Number} numDims
+ * @returns dimensions and measures total cell values as strings
+ */
+export function getTotalInfo(isDim, layout, colIndex, numDims) {
+  if (!isDim) return layout.qHyperCube.qGrandTotalRow[colIndex - numDims]?.qText;
+  if (colIndex !== 0) return '\u00A0';
+  if (layout.totals.label !== undefined && layout.totals.label) return layout.totals.label;
```
haxxmaxx, line 19: you are checking if it is not undefined then checking if it exists at all? one should be enough

src/table/components/TableHeadWrapper.jsx
```diff
@@ -2,13 +2,14 @@ import React, { memo, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { styled } from '@mui/system';
 import TableCell from '@mui/material/TableCell';
-import TableHead from '@mui/material/TableHead';
 import TableRow from '@mui/material/TableRow';
 import TableSortLabel from '@mui/material/TableSortLabel';
+import { TableHead } from '@mui/material';
```
haxxmaxx, line 7: @LiKang6688 I think we should get the component explicitly from material to not load everything, do you remember if that is why we do that?

src/table/components/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -44,8 +44,17 @@ describe('<TableBodyWrapper />', () => {
...
     jest.spyOn(selectionsUtils, 'addSelectionListeners').mockImplementation(() => jest.fn());
+    const getLayout = generateLayout(1, 1, 2);
+    getLayout.totals = {
+      label: 'Totals',
+    };
+    getLayout.qHyperCube.qGrandTotalRow = [
+      {
+        qText: '100',
+      },
+    ];
```
haxxmaxx, line 55: I see you adding this to at least two test files. Does it work to add this to generateLayout?

src/table/components/__tests__/TableTotals.spec.jsx
```diff
@@ -0,0 +1,215 @@
...
+    keyboard = {
+      enabled: false,
+    };
+    translator = { get: (s) => s };
+    selectionsAPI = {
+      isModal: () => false,
+    };
+  });
+
+  describe('Table total position:', () => {
```
haxxmaxx, line 90: This should be the same as the function name
```suggestion
  describe('getTotalPosition', () => {
```

src/table/utils/__tests__/handle-accessiblity.spec.js
```diff
@@ -149,11 +149,11 @@ describe('handle-accessibility', () => {
     });
   });
 
-  describe('handleClickToFocusHead', () => {
+  describe('handleClickToFocusCell', () => {
```
haxxmaxx, line 152: don't forget to revert this

src/table/utils/handle-key-press.js
```diff
@@ -128,6 +128,28 @@ export const headHandleKeyPress = (
...
 
+/**
+ * Handle totals row key press
+ *
+ * @param {event} evt
+ * @param {Object} rootElement
+ * @param {Array} cellCoord
+ * @param {Function} setFocusedCellCoord
+ */
+export const totalHandleKeyPress = (evt, rootElement, cellCoord, setFocusedCellCoord) => {
```
haxxmaxx, line 139: No coverage here, needs a couple of tests

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
+  const totalHeaderStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme.name()]);
+  const headRowStyle = {
+    '& :last-child': {
+      borderRight: paginationNeeded && 0,
+    },
+    'th:first-of-type': {
+      borderLeft: !paginationNeeded && '1px solid rgb(217, 217, 217)',
+    },
+  };
```
haxxmaxx, line 21: fyi I have merged a change that changes how we style things, so this wont be needed when you merge main :)

src/table/utils/handle-key-press.js
haxxmaxx: yes they should be moved

src/table/components/__tests__/TableTotals.spec.jsx
```diff
@@ -0,0 +1,185 @@
...
+    //         qText: '-',
+    //       },
+    //     ],
+    //   },
+    //   totals: {
+    //     show: false,
+    //     position: 'top',
+    //     label: 'Totals',
+    //   },
+    // };
```
Donya-qlik, line 81: I'll remove it

src/table/utils/styling-utils.js
haxxmaxx: This statement will never be true, since we catch that earlier, so you can remove it

src/table/components/TableTotals.jsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <StyledHeadRow paginationNeeded={paginationNeeded} className="sn-table-row">
+      {columns.map((column, columnIndex) => {
+        const cellCoord = [totalsPosition === 'bottom' ? rows.length + 1 : 1, columnIndex];
+        return (
+          <TableCell
+            sx={getTotalsCellStyle(totalCellStyle, totalsPosition)}
+            key={column.id}
+            align={column.align}
+            className="sn-table-total-cell sn-table-cell"
```
haxxmaxx, line 24: remove unused class

src/locale/locales/en.json
```diff
@@ -86,5 +86,9 @@
   "SNTable.SelectionLabel.selected": {
     "value": "Selected",
     "comment": "announced by screen readers for a selected value. (tew 210929)"
+  },
+  "SNTable.Object.Table.Totals": {
```
haxxmaxx, line 90: Remove for now. we might need a default string that is translated later, but it could be hard-coded for now

src/table/components/__tests__/TableBodyWrapper.spec.jsx
```diff
@@ -44,8 +44,9 @@ describe('<TableBodyWrapper />', () => {
 
   beforeEach(async () => {
     jest.spyOn(selectionsUtils, 'addSelectionListeners').mockImplementation(() => jest.fn());
+    const getLayout = generateLayout(1, 1, 2, [], [{ qText: '100' }]);
```
haxxmaxx, line 47: this is not a getter, just call it layout

src/table/components/__tests__/TableTotals.spec.jsx
```diff
@@ -0,0 +1,152 @@
...
+    keyboard = {
+      enabled: false,
+    };
+    translator = { get: (s) => s };
+    selectionsAPI = {
+      isModal: () => false,
+    };
+  });
+
+  describe('getTotalPosition:', () => {
```
haxxmaxx, line 58: move to handle-data.spec and just test the logic

src/table/components/__tests__/TableTotals.spec.jsx
```diff
@@ -0,0 +1,152 @@
...
+      });
+    });
+  });
+
+  describe('Handle keys call when TableTotals is rendered', () => {
+    beforeEach(() => {
+      layout.totals.position = 'top';
+    });
+    it('should call totalHandleKeyPress when keyDown on a total cell', () => {
+      jest.spyOn(handleKeyPress, 'totalHandleKeyPress').mockImplementation(() => jest.fn());
```
haxxmaxx, line 139: make sure you clear this in an afterEach: `afterEach(() => jest.clearAllMocks());`

src/table/utils/handle-key-press.js
```diff
@@ -121,6 +121,28 @@ export const headHandleKeyPress = ({
...
+  switch (evt.key) {
+    case 'ArrowUp':
+    case 'ArrowDown':
+    case 'ArrowRight':
+    case 'ArrowLeft': {
+      !isCtrlShift(evt) && moveFocus(evt, rootElement, cellCoord, setFocusedCellCoord);
+      break;
+    }
+    default:
+      break;
```
haxxmaxx, line 142: add test cases for when isCtrlShift is true and the default case

src/table/components/__tests__/TableBodyWrapper.spec.jsx
Donya-qlik: It is because I have another layout at line 60 but I can change it, it will not make any conflict if I call it layout here too.

src/__test__/handle-data.spec.js
```diff
@@ -154,4 +154,86 @@ describe('handle-data', () => {
       expect(setPageInfo).toHaveBeenCalledWith({ ...pageInfo, rowsPerPage: 0 });
     });
   });
+
+  describe('getTotalInfo:', () => {
+    beforeEach(() => {
+      numDims = layout.qHyperCube.qDimensionInfo.length;
+    });
```
haxxmaxx, line 161: you never change this so no need to update it everytime

src/table/components/TableWrapper.jsx
```diff
@@ -2,9 +2,11 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useRef, useCallback } from 'react';
 import Table from '@mui/material/Table';
 
+import { TableFooter } from '@mui/material';
```
haxxmaxx, line 5:
```suggestion
import TableFooter from '@mui/material/TableFooter';
```

haxxmaxx (approved):
Great work, the unit tests looks nice as well. Just fix those minor comments and you are done!

LiKang6688:
Two issues to be fixed:
- When you change header font size, the looking is not correct doing scroll
Screenshot 2022-06-20 at 14 53 10
Screenshot 2022-06-20 at 14 52 30

- When you delete 
```
"totals": {
    "show": true,
    "position": "noTotals",
    "label": "Totals"
  },

``` 
in the properties, you will get undefined `tableData`, and the table becomes blank.
Screenshot 2022-06-20 at 15 44 01

src/table/utils/styling-utils.js
```diff
@@ -154,6 +153,26 @@ export function getBodyCellStyle(layout, theme) {
...
+ * Get style of total cell
+ *
+ * @param {Object} totalHeaderStyle
+ * @param {String} totalsPosition
+ * @returns total cell style regarding the total head position
+ */
+export function getTotalsCellStyle(totalCellStyle, totalsPosition) {
+  totalCellStyle.fontWeight = 'bold';
+  if (totalsPosition === 'top') {
+    totalCellStyle.borderWidth = '0px 1px 2px 0px';
```
LiKang6688, line 166: would be good to cover here as well.

haxxmaxx:
>     * When you change header font size, the looking is not correct doing scroll
As we discussed on planning, this would likely be found but not fixed in this PR
>     * When you delete stuff from object props...
My idea was that they should be mandatory. But not sure about this and we would do another PR where you also fix the API specs. I haven't tried, what happens to extensions created before this change? Do they crash? if not I think it is fine. And then in a later PR we fix the sanity checks properly, default props and API specs. We should not release in between this PR and those fixes though

src/table/utils/styling-utils.js
haxxmaxx: wait what I never pushed that... I'll fix that

src/table/components/TableWrapper.jsx
```diff
@@ -2,9 +2,11 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useRef, useCallback } from 'react';
```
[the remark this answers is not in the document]
haxxmaxx: technically they are not included in the coverage, but we might want a test on if the totals are on top or bottom

PR #548, opened 2022-06-07
refactor: custom hooks for announce, sorting and theme

haxxmaxx:
Since there was no need to declare the announce function in the TableWrapper, I moved it and created a hook for it. Did the same for sorting and theme. This way we can keep the index file cleaner and reduce calculations to only when needed. Also
- created hooks for adding listeners
- started typing the announcer so decided to simplify the input. Now it has to be an array, if the keys have arguments you create an array for that as before. Then all the arguments has to be strings, that is actually what the translator wants


src/index.js
```diff
@@ -104,17 +102,17 @@ export default function supernova(env) {
...
+        announce,
+        changeSortOrder,
       ]);
 
       // this is the one we want to use for carbon
       useEffect(() => {
-        renderWithCarbon({ env, translator, rootElement, model, theme, selectionsAPI, app, rect, layout });
-      }, [layout, model, selectionsAPI.isModal(), theme.name(), manageData, translator.language(), app]);
+        renderWithCarbon({ env, rootElement, model, theme, selectionsAPI, app, rect, layout });
+      }, [layout, model, selectionsAPI.isModal(), theme, translator.language(), app, changeSortOrder]);
```
haxxmaxx, line 115: @vcellu Why is managedData passed to the native render? It is a pure function, you can just import it where needed?

src/index.js
```diff
@@ -10,34 +10,30 @@ import {
...
   rowsPerPageOptions: [10, 25, 100],
 };
 const nothing = async () => {};
-const renderWithCarbon = ({ env, translator, rootElement, model, theme, selectionsAPI, app, rect, layout }) => {
-  if (env.carbon) {
-    registerLocale(translator);
-    const changeSortOrder = sortingFactory(model);
+const renderWithCarbon = ({ env, rootElement, model, theme, selectionsAPI, app, rect, layout, changeSortOrder }) => {
+  if (env.carbon && changeSortOrder && theme) {
     render(rootElement, { layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect });
```
LiKang6688, line 37: @vcellu  is the render from Root.jsx or Root.native.jsx?

src/mui-setup.native.js
```diff
@@ -1 +0,0 @@
-export default function muiSetup() {}
```
LiKang6688, line 1: @vcellu Can this file be deleted?

src/nebula-hooks/use-announce-and-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+// eslint-disable-next-line no-shadow
+enum AnnouncerElements {
+  FIRST = 'first-announcer-element',
+  SECOND = 'second-announcer-element',
+}
+
+type AnnounceArgs = {
+  keys: Array<string | Array<string>>;
+  shouldBeAtomic: boolean;
+  politeness: string;
```
LiKang6688, line 14: how about `politeness: 'polite' | 'assertive' | 'off';`? it is more specific.

src/table/components/TableWrapper.jsx
```diff
@@ -1,19 +1,21 @@
...
 import { StyledTableContainer, StyledTableWrapper } from '../styles';
+
+import PaginationContent from './PaginationContent';
+import useDidUpdateEffect from '../hooks/use-did-update-effect';
+import useFocusListener from '../hooks/use-focus-listener';
+import useScrollListener from '../hooks/use-scroll-listener';
 import { handleTableWrapperKeyDown } from '../utils/handle-key-press';
-import { updateFocus, handleResetFocus, handleFocusoutEvent, getCellElement } from '../utils/handle-accessibility';
-import { handleHorizontalScroll, handleNavigateTop } from '../utils/handle-scroll';
-import announcementFactory from '../utils/announcement-factory';
```
LiKang6688, line 16: the file announcement-factory.ts in utils can be deleted, right?

src/nebula-hooks/use-announce-and-translations.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Translator } from '../types';
+
+// eslint-disable-next-line no-shadow
+enum AnnouncerElements {
+  FIRST = 'first-announcer-element',
+  SECOND = 'second-announcer-element',
+}
+
+type AnnounceArgs = {
+  keys: Array<string | Array<string>>;
```
LiKang6688, line 12: Nice! I think it is better

src/nebula-hooks/use-sorting.js
```diff
@@ -35,4 +37,12 @@ export default function sortingFactory(model) {
 
     model.applyPatches(patches, true);
   };
-}
+};
+
+const useSorting = (model) => {
+  const changeSortOrder = useMemo(() => sortingFactory(model), [model]);
```
LiKang6688, line 43: as `const model: EngineAPI.IGenericObject | undefined`, the model can be undefined. I assume that `if(model)` is needed here

src/table/components/TableWrapper.jsx
```diff
@@ -63,38 +68,15 @@ export default function TableWrapper(props) {
...
   useEffect(
     () => handleNavigateTop({ tableContainerRef, focusedCellCoord, rootElement }),
-    [tableContainerRef, focusedCellCoord]
+    [tableContainerRef, focusedCellCoord, rootElement]
   );
 
   useDidUpdateEffect(() => {
-    if (!keyboard.enabled) return;
-
     updateFocus({ focusType: keyboard.active ? 'focus' : 'blur', cell: getCellElement(rootElement, focusedCellCoord) });
```
LiKang6688, line 80: Could you add a comment to explain it a bit?

src/nebula-hooks/use-announce-and-translations.ts
haxxmaxx: yes, and they should be optional as well

LiKang6688: RIght

src/table/components/TableWrapper.jsx
```diff
@@ -63,38 +68,15 @@ export default function TableWrapper(props) {
...
+  useScrollListener(tableContainerRef, direction);
 
   useEffect(
     () => handleNavigateTop({ tableContainerRef, focusedCellCoord, rootElement }),
-    [tableContainerRef, focusedCellCoord]
+    [tableContainerRef, focusedCellCoord, rootElement]
   );
 
   useDidUpdateEffect(() => {
-    if (!keyboard.enabled) return;
```
LiKang6688, line 96: when Nebula does not handle keyboard navigation, 
`updateFocus({ focusType: keyboard.active ? 'focus' : 'blur', cell: getCellElement(rootElement, focusedCellCoord) });` still needs to run?

src/table/components/TableWrapper.jsx
haxxmaxx: when keyboard.enabled is not true, keyboard.active never changes, so this hook will never be triggered. No need for the check

src/table/components/TableWrapper.jsx
haxxmaxx: it wont, see other comment

src/nebula-hooks/use-sorting.js
haxxmaxx: hmm maybe... I'll add a sanity check in the factory function

src/index.js
haxxmaxx: yes it is, we need to sync on how this works. Will setup a meeting, let's merge this for now

src/table/components/TableWrapper.jsx
LiKang6688: no, this hook will never be triggered after renders succeed. It at least runs one time although not necessary.

src/nebula-hooks/use-sorting.js
haxxmaxx: done

src/table/components/TableWrapper.jsx
haxxmaxx: done

src/table/components/TableWrapper.jsx
haxxmaxx: well, that is what `useDidUpdateEffect` should take care of. I just tested and it is broken bc of the useEffect running twice....... Even if keyboard.enabled is true this shouldn't run the first time. And it should be trouble for the other use-case of useDidUpdateEffect. I specifically added the hook since we needed it to not run on first render

LiKang6688: I type wrong. I mean this hook will be triggered after renders succeed. It at least runs one time although not necessary.

LiKang6688: aha, right, I recalled. So the useDidUpdateEffect is broken...

haxxmaxx: I could add a counter so it makes sure it skips the first to times it is triggered but it is dirty... I think I read somewhere that it is the react strict mode messing this up, why do we have that? I think you added it? Could we remove it once we convert to typescript?

haxxmaxx: I agree. I think we would get a lot of the same warnings etc from typscript

haxxmaxx: I'll leave this for now since this change doesn't really cause the problem

LiKang6688 (approved):

src/nebula-hooks/use-announce-and-translations.ts
```diff
@@ -0,0 +1,64 @@
...
+  shouldBeAtomic?: boolean;
+  politeness?: 'polite' | 'assertive' | 'off';
+};
+
+/* creates the function for announcement */
+export const announcementFactory = (rootElement: Element, translator: Translator, prevAnnounceEl?: string) => {
```
LiKang6688, line 18: @haxxmaxx I just realized that prevAnnounceEl never gets a value which does not look correct.

haxxmaxx: from a typing perspective?

PR #549, opened 2022-06-07
feat:  when a dimension is locked, disable selections on that dimension

LiKang6688:
- check qLock in each qDimensionInfo
- when a cell is locked or not a dimension, the selection should not work

### DIFF
--- a/src/table/components/TableBodyWrapper.jsx
+++ b/src/table/components/TableBodyWrapper.jsx
@@ -67,7 +67,7 @@ function TableBodyWrapper({
       {rows.map((row, rowIndex) => (
         <TableRow hover={hoverEffect} tabIndex={-1} key={row.key} sx={rowCellStyle} className="sn-table-row">
           {columns.map((column, columnIndex) => {
-            const { id, isLocked, align } = column;
+            const { id, align } = column;
             const cell = row[id];
             const CellRenderer = columnRenderers[columnIndex];
 
@@ -83,7 +83,6 @@ function TableBodyWrapper({
                   styling={{ color: bodyCellStyle.color, backgroundColor: theme.table.backgroundColor }}
                   tabIndex={-1}
                   announce={announce}
-                  isLocked={isLocked}
                   onKeyDown={(evt) =>
                     bodyHandleKeyPress({
                       evt,
@@ -96,7 +95,6 @@ function TableBodyWrapper({
                       setFocusedCellCoord,
                       announce,
                       keyboard,
-                      isLocked,
                     })
                   }
                   onMouseDown={() => handleClickToFocusBody(cell, rootElement, setFocusedCellCoord, keyboard)}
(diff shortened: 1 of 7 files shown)


haxxmaxx:
put it on the cell as well, as we do with isDim

LiKang6688:
> put it on the cell as well, as we do with isDim in handleData

Good point. Gonna do.

haxxmaxx:
actually we only use isDim on the cell to determine if a cell is selectable or not, so this would be even cleaner
`isSelectable: c.isDim && !c.isLocked,`

src/table/components/TableBodyWrapper.jsx
```diff
@@ -67,17 +67,19 @@ function TableBodyWrapper({
       {rows.map((row, rowIndex) => (
         <TableRow hover={hoverEffect} tabIndex={-1} key={row.key} sx={rowCellStyle} className="sn-table-row">
           {columns.map((column, columnIndex) => {
-            const cell = row[column.id];
+            const { id, align } = column;
```
haxxmaxx, line 70: I'd say this is a bit overkill but fine with me

src/table/utils/__tests__/handle-key-press.spec.js
```diff
@@ -177,7 +177,7 @@ describe('handle-key-press', () => {
         cancel: jest.fn(),
         isModal: () => isModal,
       };
-      cell = { qElemNumber: 1, colIdx: 1, rowIdx: 1, isDim: true };
+      cell = { qElemNumber: 1, colIdx: 1, rowIdx: 1, isDim: true, isSelectable: true };
```
haxxmaxx, line 180: you can remove isDim here

src/table/utils/__tests__/handle-key-press.spec.js
```diff
@@ -457,8 +457,8 @@ describe('handle-key-press', () => {
         announce,
         keyboard,
       });
-      expect(evt.preventDefault).not.toHaveBeenCalledTimes(1);
-      expect(evt.stopPropagation).not.toHaveBeenCalledTimes(1);
+      expect(evt.preventDefault).not.toHaveBeenCalled();
```
haxxmaxx, line 460: isn't this less specific?

haxxmaxx (approved):

src/table/utils/__tests__/handle-key-press.spec.js
LiKang6688: if it calls two times, `not.toHaveBeenCalledTimes(1)` is still true, right?

src/table/utils/__tests__/handle-key-press.spec.js
LiKang6688: right, gonna do

src/table/utils/__tests__/handle-key-press.spec.js
haxxmaxx: ah sorry it is a `.not.` check, yes your change is better

PR #550, opened 2022-06-08
refactor: updateFocus

haxxmaxx:


src/table/utils/__tests__/handle-key-press.spec.js
```diff
@@ -124,24 +123,6 @@ describe('handle-key-press', () => {
     });
   });
 
-  describe('getRowAndColumnCount', () => {
```
haxxmaxx, line 127: This has no logic and runs in other tests

src/table/components/TableWrapper.jsx
```diff
@@ -83,11 +83,7 @@ export default function TableWrapper(props) {
   useDidUpdateEffect(() => {
     if (!keyboard.enabled) return;
 
-    updateFocus({
-      focusType: keyboard.active ? 'focus' : 'blur',
-      rowElements: rootElement.getElementsByClassName('sn-table-row'),
-      cellCoord: focusedCellCoord,
-    });
+    updateFocus({ focusType: keyboard.active ? 'focus' : 'blur', cell: getCellElement(rootElement, focusedCellCoord) });
```
LiKang6688, line 86: Is it possible to have a unit test for this part, if not, probably we need to go for rendering testing for the focus thing.

src/table/utils/handle-accessibility.js
```diff
@@ -1,7 +1,11 @@
+export const getCellElement = (rootElement, cellCoord) =>
+  rootElement.getElementsByClassName('sn-table-row')[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[
+    cellCoord[1]
+  ];
+
 export const findCellWithTabStop = (rootElement) => rootElement.querySelector("td[tabindex='0'], th[tabindex='0']");
 
-export const updateFocus = ({ focusType, rowElements = [], cellCoord = [], providedCell = undefined }) => {
-  const cell = providedCell || rowElements[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[cellCoord[1]];
+export const updateFocus = ({ focusType, cell }) => {
```
LiKang6688, line 8: how about 
`export const updateFocus = ( focusType, cell ) => {...}`

haxxmaxx: hmm actually I think it helps with the readability, since there are multiple places where the call is nested with another function call, e.g. `updateFocus({ focusType: keyboard.active ? 'focus' : 'blur', cell: getCellElement(rootElement, focusedCellCoord) });`

LiKang6688: aha, i see.

LiKang6688 (approved):

haxxmaxx:
> Looks good to me. Have you tested in client with a screenreader?

No. I'll try with all three refactor PRs that I have up tomorrow

LiKang6688:
> > Looks good to me. Have you tested in client with a screenreader?
> 
> No. I'll try with all three refactor PRs that I have up tomorrow

Good good

PR #551, opened 2022-06-08
chore(typescript): setup to enable typescript

LiKang6688:


src/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface Translator {
+  get: (str: string, args?: string[]) => string;
```
haxxmaxx, line 2: this should be a type in nebula?

LiKang6688: unfortunately, no. nebula only exports galaxy, not Translator specifically.

haxxmaxx: I see this is the way pivot does it as well

LiKang6688: good

PR #552, opened 2022-06-08
refactor: move component styling to styled

haxxmaxx:
This gets the styling out of the component files. Styled should be faster as well. There is some logic in there but this should be covered by rendering tests. If we like, the styling can be on top in each compnent file, but outside the component fn.

### DIFF
--- a/src/table/components/TableBodyWrapper.jsx
+++ b/src/table/components/TableBodyWrapper.jsx
@@ -32,6 +32,7 @@ function TableBodyWrapper({
   );
   const bodyCellStyle = useMemo(() => getBodyCellStyle(layout, theme), [layout, theme.name()]);
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
+  const cellStyle = { color: bodyCellStyle.color, backgroundColor: theme.table.backgroundColor };
 
   useEffect(() => {
     addSelectionListeners({ api: selectionsAPI, selectionDispatch, setShouldRefocus, keyboard, tableWrapperRef });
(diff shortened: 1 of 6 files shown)


src/table/styles.js
```diff
@@ -0,0 +1,135 @@
...
+  overflow: 'hidden',
+  position: 'absolute',
+  whiteSpace: 'nowrap',
+  width: '1px',
+});
+
+// ---------- FooterWrapper ----------
+
+export const StyledFooterWrapper = styled(Paper, {
+  shouldForwardProp: (prop) => prop !== 'tableTheme',
```
LiKang6688, line 18: should be `shouldForwardProp: (prop) => prop !== 'tableTheme' && prop !== 'theme', `?

src/table/styles.js
```diff
@@ -0,0 +1,135 @@
...
+    borderLeft: !paginationNeeded && '1px solid rgb(217, 217, 217)',
+  },
+  '& td, th': {
+    fontSize: bodyCellStyle.fontSize,
+    padding: bodyCellStyle.padding,
+  },
+}));
+
+export const StyledBodyRow = styled(TableRow, {
+  shouldForwardProp: (prop) => prop !== 'bodyCellStyle',
```
LiKang6688, line 69: should be `shouldForwardProp: (prop) => prop !== 'hover && prop !== 'bodyCellStyle'',`?

src/table/styles.js
```diff
@@ -0,0 +1,135 @@
+import { styled, Paper, TableContainer, TableRow, TableSortLabel, TableBody, Select, IconButton } from '@mui/material';
+
+// ---------- AnnounceWrapper ----------
```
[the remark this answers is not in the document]
haxxmaxx: alphabetically

src/table/styles.js
haxxmaxx: no hover is a property actually used by the component, so it should be passed

src/table/styles.js
haxxmaxx: no theme is the mui theme, no need to suppress it here

LiKang6688: right, I saw it is used by the component itself

PR #555, opened 2022-06-09
feat: shift + arrow up/down to select multiple values

LiKang6688:
- shift + arrow up/down key to select multiple values
Note:
	- press shift + arrow, then release the arrow up/down and the shift key to call the engine to do the selection 
	- when at the first row of the cell, shift + arrow up key, no value is selected
	- when at the head cell, shift + arrow down key, no value is selected
- screen reader correctly announce when selection multiple values
Screenshot 2022-06-16 at 03 02 16


LiKang6688:
If it is ok, I will add unit tests.

haxxmaxx:
As I think I mentioned in the old PR, I'd prefer if you put the previous and next `qElemNumber` on each cell, when looping through them in handleData. That way you get no extra arguments and don't need any logic in the body component

haxxmaxx:
Something like this 
```  const rows = qMatrix.map((r, rowIdx) => {
    const row = { key: `row-${rowIdx}` };
    columns.forEach((c, colIdx) => {
      row[c.id] = {
        ...r[colIdx],
        prevQElemNumber: qMatrix[rowIdx - 1][c.id].qElemNumber
        nextQElemNumber: qMatrix[rowIdx + 1][c.id].qElemNumber
```

LiKang6688:
> As I think I mentioned in the old PR, I'd prefer if you put the previous and next `qElemNumber` on each cell, when looping through them in handleData. That way you get no extra arguments and don't need any logic in the body component

The reason I put logic in the body component is that it only runs one array operation when you press the shift + arrow up/down key to select multiple values instead of two array operations with every interaction of the table.
But I do not know why `handleData` runs with every interaction of the table. We need to look into it.

haxxmaxx:
oh it absolutely shouldn't. It will run the first selection you make, but not after that

LiKang6688:
> oh it absolutely shouldn't. It will run the first selection you make, but not after that

Okay, I was wrong. Not every interaction, only when exiting selection mode even does not run with the first selection you make.

LiKang6688:
I still do not think it is a good idea to put two large array operations into handle data that runs when you change pages. As I think that going to the previous/next page is a frequent interaction and it is already slow.

src/table/utils/selections-utils.js
```diff
@@ -81,7 +81,12 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
     return { [qElemNumber]: rowIdx };
   }
 
-  if (selectedRows[qElemNumber] !== undefined) {
+  if (evt.shiftKey) {
+    selectedRows[qElemNumber] = rowIdx;
+    // move to a cell, and add it to the object
```
haxxmaxx, line 86:
```suggestion
    // add the next or previous cell to selectedRows, based on which arrow is pressed
```

haxxmaxx:
looks good, you probably need to look into the announcement for the new selected cell here

src/table/utils/selections-utils.js
```diff
@@ -81,7 +81,12 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
     return { [qElemNumber]: rowIdx };
   }
 
-  if (selectedRows[qElemNumber] !== undefined) {
+  if (evt.shiftKey) {
```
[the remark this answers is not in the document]
LiKang6688: Good point. Fixed

src/table/utils/selections-utils.js
```diff
@@ -81,7 +81,12 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
     return { [qElemNumber]: rowIdx };
   }
 
-  if (selectedRows[qElemNumber] !== undefined) {
+  if (evt.shiftKey && ((evt.key === 'ArrowUp' && cell.rowIdx !== 0) || evt.key === 'ArrowDown')) {
```
haxxmaxx, line 84: no need to check the row index twice. this should be enough
```suggestion
  if (evt.shiftKey && evt.key.includes('Arrow') {
```

src/table/utils/handle-key-press.js
```diff
@@ -142,6 +142,15 @@ export const bodyHandleKeyPress = ({
   switch (evt.key) {
     case 'ArrowUp':
     case 'ArrowDown':
+      // Shift + up/down arrow keys: select multiple values
+      // When at the first row of the cell, shift + arrow up key, no value is selected
+      evt.shiftKey &&
+        ((evt.key === 'ArrowUp' && cell.rowIdx !== 0) || evt.key === 'ArrowDown') &&
```
haxxmaxx, line 148: what about last row?

LiKang6688: When at the last row of the cell, there is no specific thing to consider to doing multiple value selections

LiKang6688: no, when at the last row of the cell, you press shift + arrow down, it just stops there.

LiKang6688: I agree. Though the engine does not complain, it is better that we avoid undefined: someRowIdx.

LiKang6688:
const rows = [];
  const data = dataPages[0].qMatrix;
  for (let rowIdx = 0; rowIdx < data.length; rowIdx++) {
    const row = { key: `row-${rowIdx}` };
    for (let colIdx = 0; colIdx < columns.length; colIdx++) {
      const r = data[rowIdx];
      const col = columns[colIdx];
      row[col.id] = {
        ...r[colIdx],
        rowIdx: rowIdx + top,
        colIdx: columnOrder[colIdx],
        isSelectable: col.isDim && !col.isLocked,
        rawRowIdx: rowIdx,
        rawColIdx: colIdx,
      };
    }
    rows.push(row);
  }
  
  I had a try to convert the `map` to `for`. But there is no time difference.

src/table/utils/handle-key-press.js
```diff
@@ -134,10 +134,17 @@ export const bodyHandleKeyPress = ({
   selectionsAPI = null,
 }) => {
   const isInSelectionMode = selectionsAPI?.isModal();
+  const isMultiValueSelections = () =>
```
haxxmaxx, line 137: Do you expect to use this anywhere else? If not, no need to defined a function

src/table/utils/handle-key-press.js
```diff
@@ -134,10 +134,17 @@ export const bodyHandleKeyPress = ({
...
+    (cell.prevQElemNumber || cell.nextQElemNumber) && evt.shiftKey && cell.isSelectable && isAnalysisMode;
```
haxxmaxx, line 138: Now this will make an engine call going up to the header as well... You could check them individually instead:
```suggestion
    evt.shiftKey && cell.isSelectable && isAnalysisMode && (cell.prevQElemNumber && evt.key === 'ArrowUp' || cell.nextQElemNumber && evt.key === 'ArrowDown') ;
```

src/table/utils/handle-key-press.js
haxxmaxx: or if you break it out so it is testable + reduces complecxity of `bodyHandleKeyPress`

LiKang6688: Updated

src/table/utils/handle-key-press.js
LiKang6688: Good point. Thanks. Fixed

haxxmaxx (approved):

src/table/utils/handle-key-press.js
```diff
@@ -121,6 +121,13 @@ export const headHandleKeyPress = ({
   }
 };
 
+const isMultiValueSelections = (evt, cell, isAnalysisMode) =>
+  evt.shiftKey &&
+  cell.isSelectable &&
+  isAnalysisMode &&
+  ((cell.prevQElemNumber !== undefined && evt.key === 'ArrowUp') ||
+    (cell.nextQElemNumber !== undefined && evt.key === 'ArrowDown'));
```
LiKang6688, line 129: QElemNumber might be 0

src/table/components/TableHeadWrapper.jsx
```diff
@@ -59,7 +59,7 @@ function TableHeadWrapper({
               <StyledSortLabel
                 headerStyle={headerStyle}
                 active={isCurrentColumnActive}
-                title={!constraints.passive && column.sortDirection} // passive: turn off tooltips.
+                title={!constraints.passive ? column.sortDirection : undefined} // passive: turn off tooltips.
```
LiKang6688, line 62: a minor fix

LiKang6688:
> looks good, you probably need to look into the announcement for the new selected cell here

screen reader does not work correctly, gonna look into it.

src/table/utils/selections-utils.js
```diff
@@ -110,7 +115,7 @@ const selectCell = (state, payload) => {
   handleAnnounceSelectionStatus({
     announce,
     rowsLength: selectedRowsLength,
-    isAddition: selectedRowsLength > rows.length,
+    isAddition: selectedRowsLength >= Object.keys(rows).length,
```
haxxmaxx, line 118: hmm is there a case where you shift+arrow to an an already selected value, so the number of selected values does not change? That would probably be a special case for the announcer.

LiKang6688: Correct. I added shouldAnnounce here
 https://github.com/qlik-oss/sn-table/pull/555/commits/c4b905776acba6a20166f958d138b140e5c83c38#diff-3f227d6a51355e588ecb2a10cd87fc83d3c270d50b08fe3cb560baa979976aeaR86

haxxmaxx:
I just realized me missed a big part of the feature. In the old table there is no backend call unitl you release shift... We need to find out if we want the same behaviour.

LiKang6688:
If it is okay, I will add unit tests.

src/table/utils/selections-utils.js
```diff
@@ -98,41 +103,54 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
...
   }
 
   api.cancel();
   return { ...state, rows: selectedRows, colIdx: -1 };
 };
 
+const selectMultiValues = (state, payload) => {
+  const { api, rows } = state;
+  const { colIdx } = payload.cell;
+  const rowsLength = Object.keys(rows).length;
```
haxxmaxx, line 137: I would say better to add a new prop indicating that you have values that need to be sent to backend. in your solution now you could be in selection mode, have some selections made, then just press and release shift and there will be a call to engine. We want to avoid that

haxxmaxx:
remove the zip file

src/table/utils/selections-utils.js
LiKang6688: Really good point. Fixed.

src/table/utils/__tests__/handle-key-press.spec.js
```diff
@@ -451,13 +451,57 @@ describe('handle-key-press', () => {
...
+        keyboard,
+      });
+      expect(evt.preventDefault).toHaveBeenCalledTimes(1);
+      expect(evt.stopPropagation).toHaveBeenCalledTimes(1);
+      expect(evt.target.setAttribute).toHaveBeenCalledTimes(1);
+      expect(setFocusedCellCoord).toHaveBeenCalledTimes(1);
+      expect(selectionDispatch).toHaveBeenCalledTimes(1);
+    });
+
+    it('when press shift + arrow up key on body cell, should prevent default behavior, remove current focus and set focus and attribute to the next cell, and select values for dimension', () => {
```
haxxmaxx, line 476: I think one of the arrows is enough to test this

src/table/utils/selections-utils.js
```diff
@@ -98,26 +103,25 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
...
   const selectedRowsLength = Object.keys(selectedRows).length;
   handleAnnounceSelectionStatus({
     announce,
     rowsLength: selectedRowsLength,
-    isAddition: selectedRowsLength > Object.keys(rows).length,
+    isAddition: selectedRowsLength >= Object.keys(rows).length,
   });
 
   if (selectedRowsLength) {
+    if (isSelectMultiValues) {
```
haxxmaxx, line 122: you can still figure this out with the event, no need to pass it in the payload

src/table/utils/selections-utils.js
```diff
@@ -129,10 +133,24 @@ const selectCell = (state, payload) => {
...
+const selectMultiValues = (state) => {
+  const { api, rows, colIdx, isSelectMultiValues } = state;
+
+  isSelectMultiValues &&
+    api.select({
+      method: 'selectHyperCubeCells',
+      params: ['/qHyperCubeDef', Object.values(rows), [colIdx]],
+    });
+
+  return { ...state, rows, colIdx };
```
haxxmaxx, line 145: isSelectMultiValues should be set to false here

haxxmaxx: and you are not changing anything else, so spread that:
```suggestion
  return { ...state, isSelectMultiValues: false };
```

haxxmaxx: you should also add a default value in TableContext

src/table/utils/selections-utils.js
```diff
@@ -98,26 +103,25 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
...
   handleAnnounceSelectionStatus({
     announce,
     rowsLength: selectedRowsLength,
-    isAddition: selectedRowsLength > Object.keys(rows).length,
+    isAddition: selectedRowsLength >= Object.keys(rows).length,
   });
 
   if (selectedRowsLength) {
+    if (isSelectMultiValues) {
+      return { ...state, rows: selectedRows, colIdx: cell.colIdx, isSelectMultiValues };
```
haxxmaxx, line 123: no need to have two identical return statements here, put the api.select call inside an else instead

src/table/utils/selections-utils.js
```diff
@@ -99,40 +104,52 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
...
 };
 
 export const reducer = (state, action) => {
   switch (action.type) {
     case 'select':
       return selectCell(state, action.payload);
+    case 'selectMultiValues':
+      return selectMultiValues(state);
     case 'reset':
       return state.api.isModal() ? state : { ...state, rows: {}, colIdx: -1 };
```
LiKang6688, line 154: why is state.api.isModal() needed?

src/table/utils/selections-utils.js
```diff
@@ -99,40 +104,52 @@ export const getSelectedRows = ({ selectedRows, cell, evt }) => {
...
     case 'clear':
```
LiKang6688, line 155: why is Object.keys(state.rows).length needed?

haxxmaxx: because we don't have to do anything if it is already cleared.

LiKang6688: if it is already cleared, can this 'clear' case still run?

src/table/utils/selections-utils.js
haxxmaxx: this is a good question. I've looked at this and I don't remember why. I haven't had time to look further at it, and don't want to change if I don't understand it 😅

src/table/utils/selections-utils.js
haxxmaxx: maybe not anymore. Before the reducer ran twice, bc of react.strict. Feel free to remove it if you test it and it works the same

src/table/utils/__tests__/handle-key-press.spec.js
```diff
@@ -473,6 +474,28 @@ describe('handle-key-press', () => {
       expect(selectionDispatch).toHaveBeenCalledTimes(1);
     });
 
+    it('when press shift + arrow down key on the last row cell, should prevent default behavior, remove current focus and set focus and attribute to the next cell, but not select values for dimension', () => {
```
haxxmaxx, line 477: nice

PR #557, opened 2022-06-10
fix: pass isInSelectionMode explicitly

haxxmaxx:
since we use memo and no props change when you start selection mode, the rows per page option was not hidden but it should. Passing `isInSelectionMode` explicitly makes sure we keep the memo and it updates when you go in and out of selection mode

### DIFF
--- a/src/table/components/PaginationContent.jsx
+++ b/src/table/components/PaginationContent.jsx
@@ -48,7 +48,7 @@ function PaginationContent({
   translator,
   constraints,
   footerContainer,
-  selectionsAPI,
+  isInSelectionMode,
   rect,
   handleChangePage,
   announce,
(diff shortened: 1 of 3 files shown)


LiKang6688:
we are not able to have rendering tests In selection mode for now. We need to figure it out in the future.

haxxmaxx:
> we are not able to have rendering tests In selection mode for now. We need to figure it out in the future.

agree

LiKang6688:
I cannot reproduce the case that when you start selection mode, the rows per page option was not hidden.

haxxmaxx:
> I cannot reproduce the case that when you start selection mode, the rows per page option was not hidden.

You're right. It was on my hookify branch. Since `announce` is not updated every render it does not trigger a re-render. I will add this to that PR instead

PR #559, opened 2022-06-13
fix: table moves up when doing selection

LiKang6688:
Head:
when a focused cell is updated, the scroll position is updated accordingly.

Current:
when a focused cell is updated only when using the keyboard to move up, the scroll position is updated accordingly


src/table/components/TableBodyWrapper.jsx
```diff
@@ -18,9 +18,11 @@ function TableBodyWrapper({
   setShouldRefocus,
   keyboard,
   tableWrapperRef,
+  tableContainerRef,
   announce,
 }) {
   const { rows, columns, paginationNeeded } = tableData;
+  const focusedCellCoord = useContextSelector(TableContext, (value) => value.focusedCellCoord);
```
haxxmaxx, line 25: Unfortunately this will break the performance improvement I did. `focusCellCoord` updates everytime you click to select, thus re-rendering the body component and all the cells.

LiKang6688: updated to another way of putting focusedCellCoord and logic inside tablewrapper.

src/table/components/TableBodyWrapper.jsx
```diff
@@ -8,90 +8,97 @@ import { getBodyCellStyle } from '../utils/styling-utils';
...
-  // select: turn off selections.
-  const selectionsEnabled = !!selectionsAPI && !constraints.active && !constraints.select;
-  const columnRenderers = useMemo(
-    () => columns.map((column) => getCellRenderer(!!column.stylingInfo.length, selectionsEnabled)),
-    [columns.length, selectionsEnabled]
-  );
-  const bodyCellStyle = useMemo(() => getBodyCellStyle(layout, theme), [layout, theme.name()]);
-  const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
-  const cellStyle = { color: bodyCellStyle.color, backgroundColor: theme.table.backgroundColor };
+const TableBodyWrapper = forwardRef(
```
LiKang6688, line 11: To fix a warning, now it looks weird
Reference:
https://stackoverflow.com/questions/59716140/using-forwardref-with-proptypes-and-eslint
https://github.com/facebook/react/issues/16653

src/table/components/TableWrapper.jsx
```diff
@@ -73,7 +74,7 @@ export default function TableWrapper(props) {
     return () => {
       memoedWrapper.removeEventListener('focusout', focusOutCallback);
     };
-  }, []);
+  }, [keyboard]);
```
LiKang6688, line 77: Does `keyboard` affect the hebaviour of focusout?

LiKang6688: Good

src/table/utils/handle-scroll.js
```diff
@@ -26,27 +26,18 @@ export const handleHorizontalScroll = (evt, isRTL, memoedContainer) => {
...
+  const cell = rowElements[x]?.getElementsByClassName('sn-table-cell')[y];
 
-  if (focusedCellCoord[0] < MIN_ROW_COUNT) {
+  if (cell.offsetTop - tableHead.offsetHeight - cell.offsetHeight < tableContainerRef.current.scrollTop) {
+    const targetOffsetTop = tableContainerRef.current.scrollTop - cell.offsetHeight - tableHead.offsetHeight;
     tableContainerRef.current.scrollTo({
-      top: 0,
-      behavior: 'smooth',
+      top: Math.max(0, targetOffsetTop),
+      behavior: 'instant',
```
LiKang6688, line 40: feel instant is more like native one

haxxmaxx:
I'm looking at the `handleNavigateTop` code and I don't think we need the `tableContainerRef`. We already have the `rootElement` in there and could get the container element from there. Seems like that is enought, don't need the react ref. In that case the listener could be on the body. In fact we already have a onKeyDown there sou should be able to call `handleNavigateTop` in there

LiKang6688:
> I'm looking at the `handleNavigateTop` code and I don't think we need the `tableContainerRef`. We already have the `rootElement` in there and could get the container element from there. Seems like that is enought, don't need the react ref. In that case the listener could be on the body. In fact we already have a onKeyDown there sou should be able to call `handleNavigateTop` in there

Sounds good. I am looking into it.

LiKang6688:
switch to https://github.com/qlik-oss/sn-table/pull/627

PR #560, opened 2022-06-13
refactor: use react root hook

LiKang6688:


src/table/context/useReactRoot.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
haxxmaxx, line 1: This file should be in the hooks folder

haxxmaxx: ahh wait it doesn't exist yet it is on my PR. leave it here and I can move it

haxxmaxx (approved):
Nice. Don't think we have to wait for this to be tested on mobile. I think it is broken anyway and this should not make it worse or better

PR #561, opened 2022-06-13
refactor: migrate typescript

LiKang6688:
refactor: migrate typescript


haxxmaxx:
I think we said we would do a bottom up approach, this is top down since you are doing index file... Is there a problem of doing the utils function first

PR #562, opened 2022-06-13
fix: clicking the header cell does not always sets the focus

LiKang6688:
- When clicking TableSortLabel, it gets focused instead of the header cell.
So adding an explicit click method to TableSortLabel adds focus to the header cell.

- Add a cursor to the head cell

### DIFF
--- a/src/table/components/TableHeadWrapper.jsx
+++ b/src/table/components/TableHeadWrapper.jsx
@@ -6,7 +6,7 @@ import { useContextSelector, TableContext } from '../context';
 import { VisuallyHidden, StyledHeadRow, StyledSortLabel } from '../styles';
 import { getHeaderStyle } from '../utils/styling-utils';
 import { headHandleKeyPress } from '../utils/handle-key-press';
-import { getCellElement, updateFocus, handleClickToFocusHead } from '../utils/handle-accessibility';
+import { handleMouseDownLabelToFocusHeadCell, handleClickToFocusHead } from '../utils/handle-accessibility';
 
 function TableHeadWrapper({
   rootElement,
@@ -42,11 +42,6 @@ function TableHeadWrapper({
               setFocusedCellCoord,
             });
           };
-          const handleMouseDown = (evt) => {
-            evt.preventDefault();
-            const cell = getCellElement(rootElement, [columnIndex, 0]);
-            updateFocus({ focusType: 'focus', cell });
-          };
 
           return (
             <TableCell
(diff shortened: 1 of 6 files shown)


src/table/components/TableHeadWrapper.jsx
```diff
@@ -42,6 +42,11 @@ function TableHeadWrapper({
               setFocusedCellCoord,
             });
           };
+          const handleMouseDown = (evt) => {
```
haxxmaxx, line 45: their are multiple mouse downs in this file, this name is too generic.

src/table/components/TableHeadWrapper.jsx
```diff
@@ -42,6 +42,11 @@ function TableHeadWrapper({
...
+            evt.preventDefault();
+            const cell = getCellElement(rootElement, [columnIndex, 0]);
+            updateFocus({ focusType: 'focus', cell });
```
haxxmaxx, line 48: what happens if all this function does is preventDefault? Does the event bubble to onMouseDown on the cell?

LiKang6688: Yes, the event bubbles further up the chain as evt.stoppropagation() is intentionally not used.

https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#fixing_the_problem_with_stoppropagation

haxxmaxx: so then do you need to to anything else here than preventDefault?

src/table/utils/handle-accessibility.js
```diff
@@ -1,5 +1,5 @@
 export const getCellElement = (rootElement, cellCoord) =>
-  rootElement.getElementsByClassName('sn-table-row')[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[
+  rootElement?.getElementsByClassName('sn-table-row')[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[
```
haxxmaxx, line 2: how can rootElement be undefined?

src/table/components/TableHeadWrapper.jsx
LiKang6688: Fixed

src/table/utils/handle-accessibility.js
LiKang6688: Fixed

src/table/components/TableHeadWrapper.jsx
LiKang6688: no, I do not think so.

haxxmaxx: I tried and it doesn't for some weird reason, might be some other listener by deafult on the mui component. anyway, just use `handleClickToFocusHead` here too. it should do the same thing

LiKang6688: [columnIndex, 0] the order is wrong, should be [0, columnIndex].  handleClickToFocusHead also calls when clicking label.

LiKang6688: just use handleClickToFocusHead here does not work.

PR #564, opened 2022-06-15
refactor: is selection mode

LiKang6688:
- Rename selectionsAPI.isModal() to isInSelectionMode
neither isSelectionActive or hasSelections is not that accurate.
- Add some comments

### DIFF
--- a/src/table/components/TableWrapper.jsx
+++ b/src/table/components/TableWrapper.jsx
@@ -80,6 +80,8 @@ export default function TableWrapper(props) {
   useDidUpdateEffect(() => {
     // When nebula handles keyboard navigation and keyboard.active changes,
     // make sure to blur or focus the cell corresponding to focusedCellCoord
+    // when keyboard.focus() runs, keyboard.active is true
+    // when keyboard.blur() runs, keyboard.active is false
     updateFocus({ focusType: keyboard.active ? 'focus' : 'blur', cell: getCellElement(rootElement, focusedCellCoord) });
   }, [keyboard.active]);
 
--- a/src/table/utils/handle-accessibility.js
+++ b/src/table/utils/handle-accessibility.js
@@ -84,6 +84,8 @@ export const handleFocusoutEvent = (evt, shouldRefocus, keyboard) => {
   if (keyboard.enabled && !evt.currentTarget.contains(evt.relatedTarget) && !shouldRefocus.current) {
     evt.currentTarget.querySelector('#sn-table-announcer--01').innerHTML = '';
     evt.currentTarget.querySelector('#sn-table-announcer--02').innerHTML = '';
+    // Blur the table but not focus its parent element
+    // when keyboard.active is false, this has no effect
     keyboard.blur(false);
   }
 };
--- a/src/table/utils/handle-key-press.js
+++ b/src/table/utils/handle-key-press.js
@@ -29,7 +29,9 @@ export const handleTableWrapperKeyDown = ({
       handleChangePage(page - 1);
     }
   } else if (evt.key === 'Escape' && keyboard.enabled && !isSelectionMode) {
-    // escape key: tell Nebula to relinquish focus when not in selection mode
+    // escape key: tell Nebula to relinquish the table's focus to
+    // its parent element when nebula handles keyboard navigation
+    // and not in selection mode
     preventDefaultBehavior(evt);
     keyboard.blur(true);
   }
--- a/src/table/utils/selections-utils.js
+++ b/src/table/utils/selections-utils.js
@@ -13,11 +13,15 @@ export function addSelectionListeners({ api, selectionDispatch, setShouldRefocus
     selectionDispatch({ type: 'clear' });
   };
   const resetSelectionsAndSetupRefocus = () => {
-    // if there is focus in the chart, set shouldRefocus so that you should either focus or just set the tabstop, after data has reloaded.
-    // if there is no focus on the chart, make sure you blur and focus the entire chart
     if (tableWrapperRef.current?.contains(document.activeElement)) {
+      // if there is a focus in the chart,
+      // set shouldRefocus so that you should either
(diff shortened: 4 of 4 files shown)


haxxmaxx:
If you are going to change this, `isInSelectionMode` is the most accurate IMO. The only case where we care about the toolbar is the tabbing. Otherwise we want to now if there are unconfirmed selections, that affects

src/table/utils/handle-key-press.js
```diff
@@ -28,7 +28,8 @@ export const handleTableWrapperKeyDown = ({
       setShouldRefocus();
       handleChangePage(page - 1);
     }
-  } else if (evt.key === 'Escape' && keyboard.enabled && !isSelectionActive) {
+  } else if (evt.key === 'Escape' && keyboard.enabled && !isSelectionToolbarOn) {
+    // escape key: tell Nebula to relinquish focus
```
haxxmaxx, line 32:
```suggestion
    // escape key: tell Nebula to relinquish focus when not in selection mode
```

src/table/components/TableWrapper.jsx
```diff
@@ -95,6 +96,9 @@ export default function TableWrapper(props) {
   useDidUpdateEffect(() => {
     if (!keyboard.enabled) return;
 
+    // When Nebula handles keyboard navigation,
+    // a user tabs to the chart and presses Enter or Space,
+    // the focus should be set to the previous focused cell or the first cell in the head
```
haxxmaxx, line 101: Not really, it could be set to false as well (keyboard.blur). Something like this is accurete but not sure if it makes it easier to understand
```suggestion
    // When Nebula handles keyboard navigation,
    // and the the table is focused/blurred, the last focused cell should be focused/blurred
```

haxxmaxx:
> > isInSelectionMode
> 
> sure, I will change. inSelectionMode = selection toolbar on = unconfirmed selections mode. Do i undertstand correctly?

yes they are all the same

src/table/components/TableWrapper.jsx
haxxmaxx: they both blur the table, the argument determines if it should focus the parent (in the client that is the grid cell), at least the way we defined it in client and nebula. `blurCallback` gets it as the `resetFocus` argument in the client. https://github.com/qlik-trial/sense-client/blob/master/web/assets/objects/extension/supernova-wrapper/sn-ext.js#L362

src/table/components/TableHeadWrapper.jsx
```diff
@@ -28,6 +28,8 @@ function TableHeadWrapper({
     <TableHead>
       <StyledHeadRow paginationNeeded={paginationNeeded} className="sn-table-row">
         {columns.map((column, columnIndex) => {
+          // The first cell in the head be focusable in sequential keyboard navigation,
+          // when nebula does not handle keyboard navigation
```
haxxmaxx, line 32:
```suggestion
          // The first cell in the head is focusable in sequential keyboard navigation,
          // when nebula does not handle keyboard navigation
```

src/table/components/TableWrapper.jsx
```diff
@@ -93,8 +94,11 @@ export default function TableWrapper(props) {
   );
 
   useDidUpdateEffect(() => {
-    if (!keyboard.enabled) return;
-
+    // When nebula handles keyboard navigation,
+    // - when a user tabs to the chart and presses Enter or Space (the table is focused),
+    // the focus should be set to the last focused cell or the first cell in the head
+    // - when a user move the focus outside the table (the table is blurred),
+    // the last focused cell should be blurred (loses the focus)
```
haxxmaxx, line 101: This is still inaccurate. All this does is to make sure to focus/blur when keyboard.active changes. This includes confirming selections from the selections toolbar, clicking to focus the table, cancelling selections, pressing escape in the body/header. I barely remember everything here, so we defensively need some kind of comments, but It is rather where it is called from that is complex, not what happens in here.

src/table/utils/handle-accessibility.js
```diff
@@ -84,6 +84,7 @@ export const handleFocusoutEvent = (evt, shouldRefocus, keyboard) => {
   if (keyboard.enabled && !evt.currentTarget.contains(evt.relatedTarget) && !shouldRefocus.current) {
     evt.currentTarget.querySelector('#sn-table-announcer--01').innerHTML = '';
     evt.currentTarget.querySelector('#sn-table-announcer--02').innerHTML = '';
+    // Blur the table but not focus its parent element
     keyboard.blur(false);
```
LiKang6688, line 88: is it conflict with https://github.com/qlik-oss/sn-table/pull/564/files#diff-3f227d6a51355e588ecb2a10cd87fc83d3c270d50b08fe3cb560baa979976aeaR36?

haxxmaxx: what do you mean? They should do different things in these two cases

src/table/components/TableWrapper.jsx
```diff
@@ -77,6 +78,10 @@ export default function TableWrapper(props) {
   );
 
   useDidUpdateEffect(() => {
+    // When nebula handles keyboard navigation and keyboard.active changes,
+    // make sure to blur or focus the cell corresponding to focusedCellCoord
+    // when keyboard.focus() runs, keyboard.active is true
+    // when keyboard.blur() runs, keyboard.active is false
```
haxxmaxx, line 84:
```suggestion
    // when keyboard.focus() runs, keyboard.active is set to true
    // when keyboard.blur() runs, keyboard.active is set to false
```

haxxmaxx (approved):

src/table/utils/handle-accessibility.js
LiKang6688: when nebula handles keyboard navigation and is not in selection mode, pressing the escape key to tell Nebula to relinquish the table's focus to its parent element. Both keyboard.blur(true) and keyboard.blur(false) run.

PR #567, opened 2022-06-16
fix: hide rpp select when entering selection mode

haxxmaxx:
Adding a new prop since all the other props stayed the same when entering selection mode


LiKang6688:
the title should be 'fix: pass isInSelectionMode explicitly'?

haxxmaxx:
> the title should be 'refactor: pass isInSelectionMode explicitly'?

Nope, this is a verifail. without this the pagination content does not update when you start selection mode

LiKang6688:
> > the title should be 'refactor: pass isInSelectionMode explicitly'?
> 
> Nope, this is a verifail. without this the pagination content does not update when you start selection mode

aha, because of memo, selectionsAPI does not change?

LiKang6688 (approved):

haxxmaxx:
> > > the title should be 'refactor: pass isInSelectionMode explicitly'?
> > 
> > 
> > Nope, this is a verifail. without this the pagination content does not update when you start selection mode
> 
> aha, because of memo, selectionsAPI does not change?

It doesn't, but it never has. Acutally it was the announcer changing everytime before, whenever the useEffect in index ran. Now that is memoed so we need to look at isModal() explicitely

PR #568, opened 2022-06-17
fix: move focus to selection toolbar + block focus from totals when selecting

haxxmaxx:
- for some reason, the `keyboard` argument was not passed to `handleLastTab`. put that back
- when `paginatioNeeded` is false, tab to the selection toolbar
- refactored the tests a bit and clarified some descriptions
- make sure you can't navigate to totals row when in selections mode
- had to refactor the arrow keys on table body flow a bit to reduce complexity and argument count


src/table/utils/handle-key-press.js
```diff
@@ -148,27 +149,37 @@ export const bodyHandleKeyPress = ({
...
+        isSelectionsEnabled &&
+        selectionDispatch({ type: 'select', payload: { cell, evt, announce } });
       break;
     // Enter: Confirms selections.
     case 'Enter':
       preventDefaultBehavior(evt);
-      if (!isSelectionMode) break;
-      selectionsAPI.confirm();
-      announce({ keys: ['SNTable.SelectionLabel.SelectionsConfirmed'] });
+      if (isSelectionMode) {
```
haxxmaxx, line 159: I found the early returns here hard to read, I can change back if you don't agree

src/table/utils/handle-key-press.js
```diff
@@ -178,7 +189,7 @@ export const bodyHandleKeyPress = ({
 
 export const handleLastTab = (evt, isSelectionMode, keyboard) => {
   if (isSelectionMode && evt.key === 'Tab' && !evt.shiftKey) {
-    // tab key: focus on the confirm button in the selection toolbar
+    // tab key: focus on the selection toolbar
```
haxxmaxx, line 192: The last argument for `focusSelectionToolbar` determines if the last or first option in the toolbar will be focused. This way the focus order almost makes sense. We maybe can do something similar for the sense case, but for now we always focus the confirm button

src/table/utils/handle-key-press.js
LiKang6688: your change here is good

src/table/utils/handle-key-press.js
LiKang6688: Do you consider the mashup? I had a test and it does not work as expected.

LiKang6688: In the client, when there are no pagination controls, either pressing tab or shift+table, focus on the confirm selection button. It is also not what we expected, right?

LiKang6688: Anyway, it fixes the issue you described. We take it step by step.

haxxmaxx: in the client, that is what is expected. At least that is what we settled for when we did view data

haxxmaxx: I think I tried mashup, with what keyboard setting?

LiKang6688:
```
context: {
	keyboardNavigation: true,
	},
}
```

https://user-images.githubusercontent.com/4471489/176118397-b2551381-fb23-483d-9c31-3c03eb437f01.mov

haxxmaxx: it is because you are tabbing from the totals, this was not merged when i was testing this I think. I'll have a look

src/table/utils/__tests__/handle-key-press.spec.js
```diff
@@ -478,218 +486,105 @@ describe('handle-key-press', () => {
...
-        keyboard,
-      });
+      runBodyHandleKeyPress();
       expect(evt.preventDefault).toHaveBeenCalledTimes(1);
       expect(evt.stopPropagation).toHaveBeenCalledTimes(1);
       expect(selectionDispatch).toHaveBeenCalledTimes(1);
       expect(setFocusedCellCoord).not.toHaveBeenCalled();
     });
 
-    it('when moving to a cell on a dimension column which is selected, also have some selections already, should announce the value is selected', () => {
```
haxxmaxx, line 615: moved to the handle-accessibility test file

src/table/components/TableBodyWrapper.jsx
```diff
@@ -24,14 +24,15 @@ function TableBodyWrapper({
   const columnsStylingInfoJSON = JSON.stringify(columns.map((column) => column.stylingInfo));
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const selectionDispatch = useContextSelector(TableContext, (value) => value.selectionDispatch);
-  // constraints.active: true - turn off interactions that affect the state of the visual
-  // representation including selection, zoom, scroll, etc.
-  // constraints.select: true - turn off selections.
-  const selectionsEnabled = !!selectionsAPI && !constraints.active && !constraints.select;
+  // active: turn off interactions that affect the state of the visual representation including selection, zoom, scroll, etc.
```
LiKang6688, line 27: Did you change those comments on purpose?

LiKang6688:
please add 'cannot navigate to totals in selection mode' something like that in the title

LiKang6688:
the focus is not correct when clicking a cell for the first time, which does not happen in the main branch

https://user-images.githubusercontent.com/4471489/176889666-ca2493a7-1754-4ef6-a3c2-b542b0d2348c.mov

src/table/components/TableBodyWrapper.jsx
haxxmaxx: no must be a merge error, good find

haxxmaxx:
> the focus is not correct when clicking a cell for the first time, which does not happen in the main branch
> table.mov

It must be because the clickToFocus function gets the row index wrong due to the totals. The weird thing is that is works on main. Seems like the keyboard.active useEffect in TableWrapper doesn't run. Investigating...

haxxmaxx:
I get the same problem om main. `keyboardEnabled` needs to be true. There is a problem with it set to false also. The focusedCellCoord is wrong. I can easily fix it in a different PR

LiKang6688 (approved):

PR #573, opened 2022-06-20
fix: tweaked example app

vcellu:
This Pr adds an example project for react-native.  Please the README.md in the examples/ReactNativeTable for instructions on setting up and running.


haxxmaxx:
Without yarn.lock this is still 5000 rows... How come? Isn't there things that are packaged that could be imported? If not, is it possible to have as a separate repo?

btw it is easier to just create a branch and not a fork. I thought I added you as a collaborator, but apparently not. You should be added now 👍

haxxmaxx:
closing this since we decided to do the app creation with a script instead

PR #574, opened 2022-06-21
refactor: remove optional chaining for react root

LiKang6688:


src/index.js
```diff
@@ -40,12 +40,17 @@ const renderWithCarbon = ({ env, rootElement, model, theme, selectionsAPI, app,
...
-      properties: {
-        initial: properties,
-      },
-      data: data(env),
+      properties: { initial: properties },
+      data: data(),
     },
+    // Define the extension definition.
+    ext: ext(env),
+    // Render the visual part.
```
haxxmaxx, line 53: This is explaining how nebula works, not the table specifically. I'm all for documenting this, but could it fit in nebula somehow?

LiKang6688: nebula has those docs somewhere, I will delete here

haxxmaxx (approved):

PR #585, opened 2022-06-28
fix: TotalTable overlapping with TotalHeadCells 

Donya-qlik:


src/table/styles.js
```diff
@@ -126,15 +126,19 @@ export const VisuallyHidden = styled('span')({
...
+  shouldForwardProp: (prop) => prop !== 'headerStyle' && prop !== 'isTop' && prop !== 'bodyCellStyle',
+})(({ headerStyle, bodyCellStyle, isTop }) => {
+  return {
+    ...bodyCellStyle,
+    position: 'sticky',
+    backgroundColor: headerStyle.backgroundColor,
+    fontWeight: 'bold',
+    borderWidth: isTop ? '0px 1px 2px 0px' : '2px 1px 1px 0px',
+    boxShadow: isTop ? '0px 3px 5px -5px #a6a6a6' : '3px 0px 5px -5px #a6a6a6',
+    top: isTop && 35,
```
Donya-qlik, line 138: The idea is to replace 35 with `headerStyle.height`  but up to here headerStyle does not include height( it is auto by default) so injecting the height value to headerStyle when it has already been rendered does not seem to work logically ...

src/table/components/TableHeadWrapper.jsx
```diff
@@ -18,13 +18,19 @@ function TableHeadWrapper({
...
-  children,
+  updateCellHeight,
 }) {
   const { columns, paginationNeeded } = tableData;
   const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme]);
 
+  useEffect(() => {
+    const height = rootElement.getElementsByClassName('sn-table-head-cell')[0].offsetHeight - 0.5;
```
Donya-qlik, line 29: I know subtracting 0.5px might not make sense, it's just to temporarily fix the tiny gap between head cells and total cells. I should probably find a better solution since it might be different regarding the screen size and so on.

src/table/components/TableHeadWrapper.jsx
```diff
@@ -18,13 +18,18 @@ function TableHeadWrapper({
   translator,
   selectionsAPI,
   keyboard,
-  children,
+  updateCellHeight,
```
haxxmaxx, line 21: Don't pass anything here

src/table/components/TableHeadWrapper.jsx
haxxmaxx: That could be fine for now I guess. use the setter from the context here instead as Li did: https://github.com/qlik-oss/sn-table/pull/591/files

src/table/components/TableWrapper.jsx
```diff
@@ -106,6 +106,12 @@ export default function TableWrapper(props) {
     columns.length,
   ])} ${translator.get('SNTable.Accessibility.NavigationInstructions')}`;
 
+  const [height, setHeight] = useState(0);
+
+  const updateCellHeight = (newHeight) => {
+    setHeight(newHeight);
+  };
```
haxxmaxx, line 113: remove

src/table/components/TableWrapper.jsx
```diff
@@ -124,10 +130,10 @@ export default function TableWrapper(props) {
         data-testid="table-container"
       >
         <Table stickyHeader aria-label={tableAriaLabel}>
-          <TableHeadWrapper {...props}>
-            {totalsPosition === 'top' ? <TableTotals {...props} /> : undefined}
-          </TableHeadWrapper>
-          <TableBodyWrapper {...props} setShouldRefocus={setShouldRefocus} tableWrapperRef={tableWrapperRef} />
+          <TableHeadWrapper {...props} updateCellHeight={updateCellHeight} />
+          <TableBodyWrapper {...props} setShouldRefocus={setShouldRefocus} tableWrapperRef={tableWrapperRef}>
+            {totalsPosition === 'top' ? <TableTotals {...props} height={height} /> : undefined}
```
haxxmaxx, line 135: This is where we want to pass it now. Just move the position logic inside the table body

src/table/styles.js
```diff
@@ -126,13 +126,18 @@ export const VisuallyHidden = styled('span')({
...
+})(({ headerStyle, bodyCellStyle, isTop, headRowHeight }) => ({
+  ...bodyCellStyle,
   fontWeight: 'bold',
   position: 'sticky',
+  backgroundColor: headerStyle.backgroundColor,
   borderWidth: isTop ? '0px 1px 2px 0px' : '2px 1px 1px 0px',
-  top: isTop && 35,
+  // Since the shadow is applied cell by cell and shadowbox blur at cell corners spreads little bit less, the end result is
+  // a wavy shadow at bottom of the total row.
+  // boxShadow: isTop ? '0px 3px 5px -5px #a6a6a6' : '3px 0px 5px -5px #a6a6a6',
```
Donya-qlik, line 139: Do you think we want to have a bit shadow under the Total row as the old table has ?

haxxmaxx: Don't know but we should not spend too much time on it, since we will most likely have to change it when we get a design from UX. So no need to change anything

src/table/components/TableBodyWrapper.jsx
```diff
@@ -105,6 +112,7 @@ TableBodyWrapper.propTypes = {
   keyboard: PropTypes.object.isRequired,
   tableWrapperRef: PropTypes.object.isRequired,
   announce: PropTypes.func.isRequired,
+  children: PropTypes.object,
```
haxxmaxx, line 115: I think this will always be defined, so try making it required

src/table/components/TableTotals.jsx
```diff
@@ -1,15 +1,17 @@
...
 import { totalHandleKeyPress } from '../utils/handle-key-press';
 import { removeAndFocus } from '../utils/handle-accessibility';
 import { StyledHeadRow, TotalsCell } from '../styles';
 
 function TableTotals({ rootElement, tableData, theme, layout, keyboard }) {
   const { columns, paginationNeeded, totalsPosition, rows } = tableData;
+  const headRowHeight = useContextSelector(TableContext, (value) => value.headRowHeight);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme.name()]);
+  const bodyCellStyle = useMemo(() => getBodyCellStyle(layout, theme), [layout, theme.name()]);
```
haxxmaxx, line 14: not needed here, but would be nice to refactor this to a getTotalsStyle that combine these.

src/table/components/TableWrapper.jsx
```diff
@@ -124,15 +129,10 @@ export default function TableWrapper(props) {
...
-          <TableHeadWrapper {...props}>
-            {totalsPosition === 'top' ? <TableTotals {...props} /> : undefined}
-          </TableHeadWrapper>
-          <TableBodyWrapper {...props} setShouldRefocus={setShouldRefocus} tableWrapperRef={tableWrapperRef} />
-          {totalsPosition === 'bottom' && (
-            <TableFooter>
-              <TableTotals {...props} />
-            </TableFooter>
-          )}
+          <TableHeadWrapper {...props} updateCellHeight={updateCellHeight} />
```
haxxmaxx, line 132: remove updateCellHeight prop

src/table/components/TableTotals.jsx
haxxmaxx: yes, but not necessary

src/table/components/TableHeadWrapper.jsx
```diff
@@ -18,16 +19,21 @@ function TableHeadWrapper({
...
   const { columns, paginationNeeded } = tableData;
+  const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme]);
+  const headRowRef = useRef();
+
+  useEffect(() => {
+    setHeadRowHeight(headRowRef.current.clientHeight);
+  }, [headerStyle.fontSize]);
```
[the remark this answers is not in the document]
Donya-qlik: I checked it in my branch with/without dynamic total height and it looks working, also tried in your branch didn't face similar issue! on which commit did you build it ? Is there any specific settings? (e.x fontszie).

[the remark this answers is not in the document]
haxxmaxx: In any case it is good practice to include the variables that are used inside the useEffect. (We don't do it in some places in the repo, since we don't want to trigger updates for those "missing" variables)

src/table/components/TableHeadWrapper.jsx
```diff
@@ -18,16 +19,21 @@ function TableHeadWrapper({
...
   const { columns, paginationNeeded } = tableData;
+  const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme]);
+  const headRowRef = useRef();
+
+  useEffect(() => {
+    setHeadRowHeight(headRowRef.current.clientHeight);
+  }, [headRowRef.current, headerStyle.fontSize]);
```
LiKang6688, line 32:
```suggestion
  useEffect(() => {
    setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
  }, [headRowRef.current, headerStyle.fontSize]);
```

LiKang6688: reference: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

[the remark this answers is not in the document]
LiKang6688: check this:

https://user-images.githubusercontent.com/4471489/177321037-a95a2949-3638-4ea7-adaa-f3f1e1c56d4a.mov

Donya-qlik: Yes! good idea! It's more accurate ! I'll replace it with clientHeight! However the issue with the tiny gap between rows comes from somewhere else I think. There should be something related to the first and last child styling because this misbehavior can only be seen on the last header cell. I actually saw it on the first cell as well.

src/table/components/TableTotals.jsx
```diff
@@ -18,7 +19,8 @@ function TableTotals({ rootElement, tableData, theme, layout, keyboard }) {
         const cellCoord = [isTop ? 1 : rows.length + 1, columnIndex];
         return (
           <TotalsCell
```
LiKang6688, line 21: I prefer `StyledTotalsCell` to keep consistent as others to indicate that it is styled.

src/table/components/TableHeadWrapper.jsx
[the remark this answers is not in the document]
Donya-qlik: I wonder how you reproduce it because I can't. Have you set a specific fontSize? You saw the table in my screen, right? I changed the screen size and fontSize but never get the same locally

src/table/components/TableTotals.jsx
Donya-qlik: It's the child of `StyledHeadRow`. We have the similar naming `TableCell` in TableHeadWrapper as the child of 'StyledHeadRow'. Don't you think it's more fitted or you still prefer it to be changed?

src/table/components/TableHeadWrapper.jsx
```diff
@@ -18,16 +19,21 @@ function TableHeadWrapper({
...
 }) {
   const { columns, paginationNeeded } = tableData;
+  const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme]);
+  const headRowRef = useRef();
+
+  useEffect(() => {
+    setHeadRowHeight(rootElement.getElementsByClassName('sn-table-head-cell')[0].getBoundingClientRect().height);
```
haxxmaxx, line 31: How come you are using `getElementsByClassName`?

src/table/components/TableTotals.jsx
LiKang6688: `TableCell` in 'StyledHeadRow' is not styled, so `TableCell` is correct for me. I see that when a component is styled, we add the 'styled' prefix.

src/table/styles.js
```diff
@@ -126,14 +126,15 @@ export const VisuallyHidden = styled('span')({
...
 export const TotalsCell = styled(TableCell, {
-  shouldForwardProp: (prop) => prop !== 'headerStyle' && prop !== 'isTop',
-})(({ headerStyle, isTop }) => ({
-  ...headerStyle,
+  shouldForwardProp: (prop) => prop !== 'isTop' && prop !== 'headRowHeight' && prop !== 'totalsStyle',
+})(({ totalsStyle, isTop, headRowHeight }) => ({
+  ...totalsStyle,
   fontWeight: 'bold',
   position: 'sticky',
   borderWidth: isTop ? '0px 1px 2px 0px' : '2px 1px 1px 0px',
```
LiKang6688, line 134:
```suggestion
  borderWidth: isTop ? '0px 1px 1px 0px' : '1px 1px 1px 0px',
```

Donya-qlik: The 2px is the border width that get thicker regarding where the total row shows up. It's for distinguishing between total row and body row.

haxxmaxx: This is fine for now. Let's start making changes when UX has input

LiKang6688 (approved):
good work

PR #586, opened 2022-06-28
refactor: migrate handle-data and local/src/index to typescript

LiKang6688:


src/handle-data.ts
```diff
@@ -1,6 +1,10 @@
+import { stardust } from '@nebula.js/stardust';
+import { TableLayout, HyperCube, ExtendedNxDimensionInfo, ExtendedNxMeasureInfo, Column, PageInfo } from './types';
+
 const directionMap = {
   A: 'asc',
   D: 'desc',
+  N: 'asc',
```
haxxmaxx, line 7: why is this needed?

LiKang6688: good question. Becasue of the value of`info.qSortIndicator` in `directionMap[info.qSortIndicator]` is A or D or N. https://github.com/qlik-oss/sn-table/pull/586/files#diff-26fa10f48243e01a30f0b2eee16fa394336656c6e83d68066718bad1d1ccb1cdR55

src/handle-data.ts
```diff
@@ -13,29 +17,31 @@ const MAX_CELLS = 10000;
...
   const { qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIndex < numDims;
-  const info = isDim ? qDimensionInfo[colIndex] : qMeasureInfo[colIndex - numDims];
+  const info = (isDim ? qDimensionInfo[colIndex] : qMeasureInfo[colIndex - numDims]) as
+    | ExtendedNxMeasureInfo
+    | ExtendedNxDimensionInfo;
   const isHidden = info.qError?.qErrorCode === 7005;
-  const isLocked = info.qLocked;
+  const isLocked = isDim && qDimensionInfo[colIndex].qLocked;
```
haxxmaxx, line 44: why can't you use info here

src/handle-data.ts
```diff
@@ -75,7 +81,16 @@ export function getTotalPosition(layout) {
...
-export default async function manageData(model, layout, pageInfo, setPageInfo) {
+export default async function manageData(
+  model: EngineAPI.IGenericObject,
+  layout: TableLayout,
+  pageInfo: PageInfo,
+  setPageInfo: stardust.SetStateFn<{
+    page: number;
+    rowsPerPage: number;
+    rowsPerPageOptions: number[];
+  }>
```
haxxmaxx, line 92: doesn't the function take a `PageInfo` object as argument?

src/handle-data.ts
LiKang6688: as qLocked does not exist in qMeasureInfo

src/handle-data.ts
```diff
@@ -99,13 +114,14 @@ export default async function manageData(model, layout, pageInfo, setPageInfo) {
...
+  const columns = columnOrder.map((colIndex: number) => getColumnInfo(layout, colIndex)).filter(Boolean) as Column[];
   const dataPages = await model.getHyperCubeData('/qHyperCubeDef', [
     { qTop: top, qLeft: 0, qHeight: height, qWidth: totalColumnCount },
   ]);
 
-  const rows = dataPages[0].qMatrix.map((r, rowIdx) => {
-    const row = { key: `row-${rowIdx}` };
+  const rows = dataPages[0].qMatrix.map((r, rowIdx: number) => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const row = { key: `row-${rowIdx}` } as any;
```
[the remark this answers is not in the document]
LiKang6688: I cannot come up with one. any suggestions?

src/handle-data.ts
haxxmaxx: but it will never be qMeasureInfo if isDim is true. Does
```suggestion
  const isLocked = isDim && info.qLocked;
```
work or does typescript complain?

LiKang6688: yes, typescript does not know that, so it complains

haxxmaxx:
```suggestion
  const isLocked = isDim && (info as ExtendedNxDimensionInfo).qLocked;
```

src/handle-data.ts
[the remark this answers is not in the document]
haxxmaxx: I have a commit that I think would work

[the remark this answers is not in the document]
LiKang6688: which commit?

src/handle-data.ts
LiKang6688: Fixed

src/handle-data.ts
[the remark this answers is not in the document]
haxxmaxx: I have it locally. Something like this, types:
```
export interface TableCell {
  qText: string | undefined;
  qAttrExps: EngineAPI.INxAttributeExpressionValues;
  rowIdx: number;
  colIdx: number;
  isSelectable: boolean;
  rawRowIdx: number;
  rawColIdx: number;
  prevQElemNumber: number;
  nextQElemNumber: number;
}

export interface TableRow {
  // for the row key, string is needed
  [key: string]: TableCell | string;
}
```
and then in handle data:
```
    const row: TableRow = { key: `row-${rowIdx}` };
    columns.forEach((c, colIdx) => {
      const { qText, qAttrExps } = r[colIdx];
      row[c.id] = {
        qText,
        qAttrExps,
        rowIdx: rowIdx + top,
        ...
```

src/handle-data.ts
```diff
@@ -13,29 +17,31 @@ const MAX_CELLS = 10000;
  * @param {Number} numDims
  * @returns dimensions and measures total cell values as strings
  */
-export function getTotalInfo(isDim, layout, colIndex, numDims) {
+export function getTotalInfo(isDim: boolean, layout: TableLayout, colIndex: number, numDims: number) {
```
haxxmaxx, line 20: have a look if the type for layout exists on stardust

src/types.ts
```diff
@@ -1,4 +1,73 @@
 export interface Translator {
   get: (str: string, args?: string[]) => string;
+  add: (item: { id: string; locale: object }) => void;
```
haxxmaxx, line 3: translator exists on stardust

haxxmaxx:
change the tests as well

PR #587, opened 2022-06-29
refactor: migrate ext and verify-translations to typescript

LiKang6688:


src/ext.ts
```diff
@@ -1,5 +1,9 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
haxxmaxx, line 2: what does this mean?

src/ext.ts
```diff
@@ -19,7 +23,7 @@ const columnExpressionItems = {
     translation: 'Object.Table.Columns.VisibilityCondition',
     defaultValue: { qv: '' },
     tid: 'visibilityCondition',
-    isExpression: (val) => typeof val === 'string' && val.trim().length > 0,
+    isExpression: (val: string) => typeof val === 'string' && val.trim().length > 0,
```
haxxmaxx, line 26: I'm pretty sure val can be something else since the function checks if it is a string

src/ext.ts
LiKang6688: as qlik-modifiers does not contain types, it is a workaround to let typescript not complain.
I saw that people did the same for sn-table.
Screenshot 2022-06-29 at 22 24 57

src/ext.ts
LiKang6688: right, then any idea what other types could be?

src/ext.ts
haxxmaxx: ok. If we notice that we need  // @ts-ignore in other places we can disable the `@typescript-eslint/ban-ts-comment` rule

src/ext.ts
haxxmaxx: I guess yeah

haxxmaxx:
So for a lot of the typings you are manually defining stuff that are in the api specs. I saw that you can generate types from the api specs so you don't have to do it twice, which might lead to mismatches. I think Thomas did it [here](https://github.com/qlik-oss/nebula.js/pull/752)

LiKang6688:
> So for a lot of the typings you are manually defining stuff that are in the api specs. I saw that you can generate types from the api specs so you don't have to do it twice, which might lead to mismatches. I think Thomas did it [here](https://github.com/qlik-oss/nebula.js/pull/752)

Good point. I will look into the way to generate types from the api specs

PR #588, opened 2022-06-29
docs: update api specs for totals

haxxmaxx:
Struggling to decide on what the defaults should be and what to classify them as. The defaults in object-properties are used, unless they are overridden. So we will always have a `totals` object, which we assume in the code. Now the tricky part is that there are default values in the `ext` file as well, defining the property panel. There are default values in there too. One of them being the totals label, which is dependent on locale... I think we want to avoid having double default values? Or is that fine?

If the defaults are removed from ext file, then do we need to have a translator in object-properties? I don't think that is posible


LiKang6688 (approved):
having same double default values looks correct to me

haxxmaxx:
> having same double default values looks correct to me

yeah maybe we can keep it it like this, there should be no major changes to the api specs if we decide to do something different. We need a new release soon so this will have to do

veinfors (approved):

PR #594, opened 2022-06-30
refactor: convert context folder to typescript

haxxmaxx:
the `createSelectorProvider` and `useContextSelector` where originally in typescript so just reused that code. Had to do some slight changes since we are using typescript in strict mode. Also had to add support for tsx files


src/table/context/TableContext.tsx
```diff
@@ -1,16 +1,29 @@
...
+export const TableContext = createContext({});
 
 const ProviderWithSelector = createSelectorProvider(TableContext);
 
-export const TableContextProvider = ({ children, selectionsAPI, cellCoordMock, selectionDispatchMock }) => {
+interface ContextProviderProps {
+  children: JSX.Element;
+  selectionsAPI: stardust.ObjectSelections;
+  cellCoordMock: [number, number];
+  selectionDispatchMock: jest.Mock<any, any>;
```
LiKang6688, line 15: any is used often. Shall we disable `Unexpected any. Specify a different type.eslint[@typescript-eslint/no-explicit-any](https://typescript-eslint.io/rules/no-explicit-any)`?

LiKang6688 (approved):

src/table/context/TableContext.tsx
haxxmaxx: Hmm not sure... This is the type that comes from jest so we can't type more than this. But in other cases it might just be laziness and then there should be a warning. If anything it can be disabled for this specific line maybe. Let's keep it in mind and see what we want to do when we have converted some more

PR #598, opened 2022-07-04
refactor: convert selection-utils to typescript

haxxmaxx:
- convert selection-utils and its test file
- remove rule for neste terneries. only used in one place
- changed to use the `stardust.Translator` type
- changed to three arguments instead of passing an object in two functions in selection-utils
- add enum for action types


src/table/utils/__tests__/selections-util.spec.ts
```diff
@@ -6,50 +7,59 @@ import {
...
+    let tableWrapperRef: React.MutableRefObject<HTMLElement>;
 
     beforeEach(() => {
       selectionDispatch = jest.fn();
       setShouldRefocus = jest.fn();
       api = {
         on: jest.fn(),
         removeListener: jest.fn(),
-      };
+      } as unknown as ExtendedSelectionAPI;
```
haxxmaxx, line 36: Doing a bunch of these. Not sure if it is good practice. But for the tests it feels overkill do define everything, especially for events. Could also allow a simpler type in the arguments for easier mocking, but that just feels like it would be less strictly types -> not as effective catching wrong types

src/table/utils/__tests__/selections-util.spec.ts
```diff
@@ -134,17 +148,23 @@ describe('selections-utils', () => {
       expect(state.api.begin).toHaveBeenCalledTimes(1);
       expect(state.api.select).toHaveBeenCalledWith({ method: 'selectHyperCubeCells', params });
       expect(state.api.cancel).not.toHaveBeenCalled();
-      expect(action.payload.announce).toHaveBeenCalledTimes(1);
+      expect(action?.payload?.announce).toHaveBeenCalledTimes(1);
```
haxxmaxx, line 151: payload is optional since some actions don't have a payload, need optional chaining here

src/table/utils/__tests__/selections-util.spec.ts
```diff
@@ -217,21 +237,12 @@ describe('selections-utils', () => {
       const newState = reducer(state, action);
       expect(newState).toEqual(state);
     });
-
-    it('should return error when incorrect action type', () => {
```
haxxmaxx, line 221: Not sure if this is needed anymore. There can't be an incorrect action since it would mean the wrong type and not even compile

src/table/utils/selections-utils.ts
```diff
@@ -1,11 +1,38 @@
...
+  Action,
+  ExtendedSelectionAPI,
+  ActionPayload,
+  AnnounceFn,
+  SelectedRows,
+  ContextValue,
+} from '../../types';
+
+// eslint-disable-next-line no-shadow
+export enum SelectionStates {
```
haxxmaxx, line 14: I have no idea why lint thinks this is shadowing. The error says that it has already been defined on line 14.... redefined on the line it is being defined? 👍

LiKang6688:
a new title?

haxxmaxx:
> a new title?

Of course, done

src/table/utils/__tests__/selections-util.spec.ts
MTjody: Correct!

src/table/utils/__tests__/selections-util.spec.ts
MTjody: Yeah I agree. Especially also since you're mocking everything.

haxxmaxx: So this is fine?

src/table/utils/selections-utils.ts
haxxmaxx: let me check

src/types.ts
```diff
@@ -1,4 +1,64 @@
...
+}
+
+export interface ExtendedSelectionAPI extends stardust.ObjectSelections {
+  on(eventType: string, callback: () => void): void;
+  removeListener(eventType: string, callback: () => void): void;
+}
+
+export interface SelectedRows {
+  [key: string]: number;
+}
```
MTjody, line 24: This one is a bit replaceable. You can use the built-in type `Record<string, number>` instead.

src/types.ts
```diff
@@ -1,4 +1,64 @@
...
+export interface Action {
+  type: 'select' | 'selectMultiValues' | 'reset' | 'clear';
+  payload?: ActionPayload;
+}
+
+// export type SelectionAction =
+//   | { type: 'select'; payload: SelectPayload }
+//   | { type: 'selectMultiValues'; payload?: SelectPayload }
+//   | { type: 'reset'; payload?: SelectPayload }
+//   | { type: 'clear'; payload?: SelectPayload };
```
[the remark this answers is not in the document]
MTjody: Going to need some more info in order to help :) Which tests? What's "the bottom one"?

src/types.ts
haxxmaxx: I knew I had seen something like that, will update

src/types.ts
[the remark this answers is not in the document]
haxxmaxx: I'm referring to the commented `type SelectionAction` and I had it like this before
```
export type SelectionAction =
  | { type: 'select'; payload: SelectPayload }
  | { type: 'selectMultiValues' }
  | { type: 'reset' }
  | { type: 'clear };
```
I couldn't get the tests for the reducer to work, starts on line 109 in the selection-utils.spec file

src/table/utils/selections-utils.ts
haxxmaxx: Found it: https://stackoverflow.com/questions/63961803/eslint-says-all-enums-in-typescript-app-are-already-declared-in-the-upper-scope

The normal eslint rule had to be disabled and enabled as a typescript specific rule. Should be fine to do already since we shouldn't be adding any js code anymore

src/table/utils/__tests__/selections-util.spec.ts
MTjody: Depends on your level of ambition. Since it's a test file I would say it's fine.

haxxmaxx: Yeah I would never do this in the actual code, kind of defeats the point

src/table/utils/selections-utils.ts
```diff
@@ -16,6 +10,26 @@ export enum SelectionStates {
...
+export enum SelectionActions {
+  SELECT = 'select',
+  RESET = 'reset',
+  CLEAR = 'clear',
+  SELECT_MULTI_VALUES = 'selectMultiValues',
+}
+
+export interface Action<T = any> {
+  type: T;
+}
```
MTjody, line 22: You prob want this one somewhere else so it's re-usable

src/table/components/withSelections.jsx
```diff
@@ -11,7 +11,9 @@ export default function withSelections(CellComponent) {
     const selectionDispatch = useContextSelector(TableContext, (value) => value.selectionDispatch);
 
     const handleMouseUp = (evt) =>
-      cell.isSelectable && evt.button === 0 && selectionDispatch({ type: 'select', payload: { cell, evt, announce } });
+      cell.isSelectable &&
+      evt.button === 0 &&
+      selectionDispatch({ type: SelectionActions.SELECT, payload: { cell, evt, announce } });
```
MTjody, line 16: Have you heard of the 'action-creator' concept? it abstracts creating actions with the right type and params, it cn also help with default values and such. Define them in the same file as your actions. You can see some examples [here](https://github.com/qlik-trial/direct-query-modelling-ui/blob/main/src/state/shared.ts#L65-L87). Something like:

```typescript
function select(cell, evt, announce) {
  return 
    { type: SelectionActions.SELECT, payload: { cell, evt, announce } }
}

MTjody (approved):
LGTM. Action creators are pretty handy and is a good pattern to use, consider that before merge 🥇

PR #606, opened 2022-07-06
fix: adjust cell coord to totals position when clikcing to focus

haxxmaxx:
SImply need to take the totalsPosition into accounce, since it means that the row index should be one more + 1 if it is on top

### DIFF
--- a/src/table/utils/handle-accessibility.js
+++ b/src/table/utils/handle-accessibility.js
@@ -34,9 +34,10 @@ export const removeAndFocus = (newCoord, rootElement, setFocusedCellCoord, keybo
   keyboard.enabled && keyboard.focus();
 };
 
-export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, keyboard) => {
+export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, keyboard, totalsPosition) => {
   const { rawRowIdx, rawColIdx } = cell;
-  removeAndFocus([rawRowIdx + 1, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
+  const firstBodyRowIdx = totalsPosition === 'top' ? 2 : 1;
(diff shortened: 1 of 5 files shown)


src/table/utils/handle-accessibility.js
```diff
@@ -34,9 +34,10 @@ export const removeAndFocus = (newCoord, rootElement, setFocusedCellCoord, keybo
   keyboard.enabled && keyboard.focus();
 };
 
-export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, keyboard) => {
+export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, keyboard, totalsPosition) => {
   const { rawRowIdx, rawColIdx } = cell;
-  removeAndFocus([rawRowIdx + 1, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
+  const firstBodyRowIdx = totalsPosition === 'top' ? 2 : 1;
```
LiKang6688, line 39: the variable name `firstBodyRowIdx` does looks confused to me.

haxxmaxx: it is the index that the first row has, when you take header and totals into account. the `rawRowIdx` is the index in the body, then we need to add 1 or 2 to compensate. Do you have any other suggestions?

LiKang6688: how about
```
const cellRowIdx = totalsPosition === 'top' ? rawRowIdx + 2 : rawRowIdx + 1;
  removeAndFocus([cellRowIdx, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
```

LiKang6688 (approved):

src/table/utils/handle-accessibility.js
haxxmaxx: sure, I might call it `adjustedRowIdx` though

PR #607, opened 2022-07-06
fix: sometimes the border color of cell in header disappears

LiKang6688:
Before:
Screenshot 2022-07-06 at 10 22 30
sometimes when you add or remove measure/dimensions, the border color of a cell in header disappears as shown in the iamge above.

Now:
the border color of a cell should always be present.


haxxmaxx:
is that an image of before or after the fix? Can you add a bit more description?

LiKang6688:
> is that an image of before or after the fix? Can you add a bit more description?

before, sure

PR #608, opened 2022-07-06
refactor: convert color utils

haxxmaxx:
Needed to add some sanity checks
Tested with a couple of themes, looks the same as before


src/table/utils/color-utils.ts
```diff
@@ -43,29 +46,37 @@ export function resolveToRGBAorRGB(input) {
...
+  const a = matches?.[4];
+  return a !== undefined ? +a === 0 : false;
 }
 
-export function removeOpacity(color) {
-  const rgba = resolveToRGBAorRGB(color);
-  const matches = /^rgba\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d(\.\d+)?)\s*\)$/i.exec(rgba);
-  if (matches) return `rgb(${matches[1]},${matches[2]},${matches[3]})`;
+export function removeOpacity(color: string | undefined): string | undefined {
+  if (color !== undefined) {
```
haxxmaxx, line 76: I decided to take care of the undefined on this level so that `resolveToRGBAorRGB` always receives a string

src/table/utils/color-utils.ts
```diff
@@ -24,7 +27,7 @@ export function resolveToRGBAorRGB(input) {
   // argb
   matches = /^argb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/i.exec(input);
   if (matches) {
-    const a = Math.round(matches[1] / 2.55) / 100;
+    const a = Math.round(Number(matches[1]) / 2.55) / 100;
```
LiKang6688, line 30: rarely use Number(). How about '+' instead.

haxxmaxx: ah yes that is better

src/table/utils/__tests__/color-utils.spec.ts
```diff
@@ -197,8 +196,8 @@ describe('color-utils', () => {
       const result = removeOpacity(color);
       expect(result).toBe('rgb(0,0,0)');
     });
-    it('should return undefined when nothing is provided', () => {
-      const result = removeOpacity();
+    it('should return undefined when color is undefined', () => {
```
LiKang6688, line 199: 'should return undefined when nothing is provided' is also a valid test case I think

src/table/utils/__tests__/color-utils.spec.ts
```diff
@@ -46,8 +46,8 @@ describe('color-utils', () => {
       const result = resolveToRGBAorRGB(color);
       expect(result).toBe('rgba(255,255,255,0)');
     });
-    it('should return when nothing is provided', () => {
-      const result = resolveToRGBAorRGB();
+    it('should return none when invalid color', () => {
```
LiKang6688, line 49: 'should return when nothing is provided' is also a valid case i think

src/table/utils/__tests__/color-utils.spec.ts
```diff
@@ -129,49 +129,48 @@ describe('color-utils', () => {
...
-      color = '#000000';
+      const color = '#000000';
 
       const result = isTransparentColor(color);
       expect(result).toBe(false);
     });
 
-    it('should be false when nothing is provided', () => {
-      const result = isTransparentColor();
+    it('should be false when color is undefined', () => {
```
LiKang6688, line 156: 'should be false when nothing is provided' is also a valid test case i think

src/table/utils/__tests__/color-utils.spec.ts
haxxmaxx: No, when this is typed it always wants an argument. so `removeOpacity()` is invalid. You can still do `removeOpacity(someVariable)` where the variable is undefined

LiKang6688: aha, color: string | undefined. I see.

LiKang6688 (approved):

PR #610, opened 2022-07-07
fix: sometimes the border color of a cell in header disappears

LiKang6688:
177504868-e79bddb0-a122-44d1-b9a0-0ff0bd5f64e5

image (3)

- the border sometimes does not work as expected in the `position: sticky` Dom node in the client
therefore, switch from `border` to `box-shadow`

- in the client,
Screenshot 2022-07-10 at 18 16 48


haxxmaxx:
let's discuss this on the meeting today

src/table/styles.js
```diff
@@ -77,29 +67,25 @@ export const StyledTableBody = styled(TableBody, {
...
-);
+})(({ hover, bodyCellStyle }) => ({
+  '&&:hover': {
+    '& td:not(.selected), th:not(.selected)': {
+      backgroundColor: hover && bodyCellStyle.hoverBackgroundColor,
+      color: hover && bodyCellStyle.hoverFontColor,
+    },
+  },
+  '& :last-child': {
+    boxShadow: `inset 1px -1px 0 ${bodyCellStyle.borderColor}, inset -1px 0 0 ${bodyCellStyle.borderColor}`,
```
haxxmaxx, line 78: So this is so that the last cell looks correct? I think a comment here would clarify

haxxmaxx: same for the other ones

LiKang6688: sure, gonna add a comment.

LiKang6688: Done

haxxmaxx:
Closing this for now. Since this is a really small issue only visible sometimes when toggling between edit and analysis mode, which is not even a use-case in the client yet.

PR #611, opened 2022-07-07
refactor: convert nebula hooks

haxxmaxx:
- convert nebula hooks
- exclude ts spec files from codeclimate


src/nebula-hooks/__tests__/use-sorting.spec.ts
```diff
@@ -44,33 +45,39 @@ describe('use-sorting', () => {
     it('should call apply patches with second dimension first in sort order', async () => {
       expectedPatches[0].qValue = '[1,0,2,3]';
 
-      await changeSortOrder(layout, column);
+      if (changeSortOrder) {
```
haxxmaxx, line 48: possibly undefined so easiest to check it like this

LiKang6688:
Screenshot 2022-08-01 at 23 47 08

src/nebula-hooks/__tests__/use-sorting.spec.ts
```diff
@@ -11,31 +12,31 @@ describe('use-sorting', () => {
...
+    let originalOrder: number[];
+    let column: Column;
+    let layout: TableLayout;
+    let model: EngineAPI.IGenericObject;
+    let changeSortOrder: ((layout: TableLayout, column: Column) => Promise<void>) | undefined;
+    let expectedPatches: EngineAPI.INxPatch[];
 
     beforeEach(() => {
       originalOrder = [0, 1, 2, 3];
       layout = generateLayout(2, 2, 2);
```
LiKang6688, line 24: here the type does not match. I do not know how to fix it yet.

src/table/Root.jsx
```diff
@@ -24,7 +24,3 @@ export function render(reactRoot, props) {
 export function teardown(reactRoot) {
   reactRoot.unmount();
 }
-
-export function mount() {
```
LiKang6688, line 28: Hi @vcellu, I see that you added the mount function. Is it okay to remove it? I do not see that it is needed,

src/nebula-hooks/__tests__/use-sorting.spec.ts
LiKang6688: Screenshot 2022-08-02 at 00 09 42

haxxmaxx: I will have alook

src/table/Root.jsx
haxxmaxx: it is defined in the root.native file, it needs to run in the native case. In the web case it just does nothing and thus no condition needs to check the environment in the useReactRoot hook. You should probably revert this

src/__test__/generate-test-data.ts
```diff
@@ -38,12 +46,12 @@ export function generateLayout(nDims, nMeas, nRows, qColumnOrder = [], qGrandTot
 
   return {
     qHyperCube: {
-      qDimensionInfo,
-      qMeasureInfo,
+      qDimensionInfo: qDimensionInfo as unknown as EngineAPI.INxDimensionInfo[],
```
LiKang6688, line 49: 'as unknown as xx' what does it mean?

src/__test__/generate-test-data.ts
```diff
@@ -38,12 +46,12 @@ export function generateLayout(nDims, nMeas, nRows, qColumnOrder = [], qGrandTot
...
+      qMeasureInfo: qMeasureInfo as unknown as EngineAPI.INxMeasureInfo[],
       qGrandTotalRow,
       qColumnOrder,
       qSize: { qcx: nDims + nMeas, qcy: nRows },
-    },
+    } as unknown as HyperCube,
```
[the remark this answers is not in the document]
haxxmaxx: Yeah the thing is that we only want to mock a subset of everything that is on some of these types. And I get an error if I just do ` as HyperCube`. Already doing this in some tests. The option would be, as you say, to create a new type, but that would make the typing less strict, since we would have `theRealType | theMockedType`. Or we mock it properly, but it does not add any real value

LiKang6688 (approved):
Nice one

PR #612, opened 2022-07-07
fix: total label should appear only on dimension with zero column order

Donya-qlik:
there is no label when there is no dimension in the first column

Screenshot 2022-07-07 at 11 14 57

### DIFF
--- a/src/handle-data.js
+++ b/src/handle-data.js
@@ -23,13 +23,13 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
  * @param {Number} numDims
  * @returns dimensions and measures total cell values as strings
  */
-export function getTotalInfo(isDim, layout, colIndex, numDims, colOrder) {
+export function getTotalInfo(isDim, layout, colIndex, numDims, columnOrder) {
   if (!isDim) return layout.qHyperCube.qGrandTotalRow[colIndex - numDims]?.qText;
-  if (colIndex === 0 && colOrder[0] === 0) return layout.totals.label;
+  if (colIndex === 0 && columnOrder[0] === 0) return layout.totals.label;
   return '';
 }
 
-export function getColumnInfo(layout, colIndex, colOrder) {
+export function getColumnInfo(layout, colIndex, columnOrder) {
   const { qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIndex < numDims;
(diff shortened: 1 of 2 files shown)


haxxmaxx:
add a proper commit type to the title, I'm assuming it is `fix:`

src/handle-data.js
```diff
@@ -5,6 +5,16 @@ const directionMap = {
...
+export function getHighestPossibleRpp(width, rowsPerPageOptions) {
+  const highestPossibleOption = [...rowsPerPageOptions].reverse().find((opt) => opt * width <= MAX_CELLS);
+  return highestPossibleOption || Math.floor(MAX_CELLS / width); // covering corner case of lowest option being too high
+}
+
+export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
+  const columnsLength = qDimensionInfo.length + qMeasureInfo.length;
+  return qColumnOrder?.length === columnsLength ? qColumnOrder : [...Array(columnsLength).keys()];
+}
+
```
haxxmaxx, line 17: How come you moved these?

haxxmaxx: ah because you are using it in `getTotalInfo`

src/handle-data.js
```diff
@@ -15,18 +25,8 @@ const MAX_CELLS = 10000;
...
-
-export function getHighestPossibleRpp(width, rowsPerPageOptions) {
-  const highestPossibleOption = [...rowsPerPageOptions].reverse().find((opt) => opt * width <= MAX_CELLS);
-  return highestPossibleOption || Math.floor(MAX_CELLS / width); // covering corner case of lowest option being too high
-}
-
-export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
-  const columnsLength = qDimensionInfo.length + qMeasureInfo.length;
-  return qColumnOrder?.length === columnsLength ? qColumnOrder : [...Array(columnsLength).keys()];
+  if (colIndex === 0 && getColumnOrder(layout.qHyperCube)[0] === 0) return layout.totals.label;
```
haxxmaxx, line 28: Running `getColumnOrder` here isn't the best idea. `getColumnInfo` runs for every column, which means you could end up running `getColumnOrder` 2000 times at most. Better to pass it through in this case, since it is already calculated before `getColumnInfo` runs

src/handle-data.js
```diff
@@ -99,7 +99,7 @@ export default async function manageData(model, layout, pageInfo, setPageInfo) {
 
   const columnOrder = getColumnOrder(qHyperCube);
   // using filter to remove hidden columns (represented with false)
-  const columns = columnOrder.map((colIndex) => getColumnInfo(layout, colIndex)).filter(Boolean);
+  const columns = columnOrder.map((colIndex) => getColumnInfo(layout, colIndex, columnOrder)).filter(Boolean);
```
LiKang6688, line 102: i see that you are using the name `columnOrder` here and i think it is good to use the name every place.

Donya-qlik: Done

PR #615, opened 2022-07-15
refactor: convert handle-scrollto typescript

Donya-qlik:


src/table/utils/__tests__/handle-scroll.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      focusedCellCoord = [8, 0];
+      tableContainerRef = {
+        current: {
+          scrollTo: jest.fn(),
+          scrollTop: SCROLL_TOP_IDX * rowHeight,
+        },
+      };
+
+      rootElement = {
+        getElementsByClassName: (query: string) => {
```
Donya-qlik, line 169: There is an issue here: 
It complains about `getElementByClassName` which should be `(className: string) => <HTMLCollectionOfElement>`  but due to using a function(to set a value for height and top), the type has been changed. I tried to  kind of mock it or use casting but it didn't work!

haxxmaxx:
Li finished this since we he did some other refactoring in this file. Closing this

Donya-qlik:
The related files have been modified in another PR: https://github.com/qlik-oss/sn-table/pull/629

PR #619, opened 2022-07-28
fix: errors when doing typescript compiling

LiKang6688:
Before:
Screenshot 2022-07-28 at 11 05 09

After:
Screenshot 2022-07-28 at 10 50 15


src/table/utils/__tests__/get-cell-renderer.spec.js
```diff
@@ -5,8 +5,8 @@ import * as withStyling from '../../components/withStyling';
 
 describe('render', () => {
   describe('getCellRenderer', () => {
-    let selectionsEnabled: boolean;
-    let hasColumnStyling: boolean;
+    let selectionsEnabled;
+    let hasColumnStyling;
```
haxxmaxx, line 9: why revert this file?

src/nebula-hooks/use-announce-and-translations.ts
```diff
@@ -8,6 +8,10 @@ enum AnnouncerElements {
   SECOND = 'second-announcer-element',
 }
 
+interface Translator extends stardust.Translator {
+  language: () => string;
+}
```
haxxmaxx, line 13: I've probably mentioned this at some point. Nebulas API for a lot of the stardust hooks is wrong. Optimally we should be able to use the type from stardust. Or am I missing a reason that it i wont be possible?

LiKang6688: you are right. It is wrong in nebula types so we need to extend it as you did in https://github.com/qlik-oss/sn-table/pull/611/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR22.

src/table/utils/__tests__/get-cell-renderer.spec.js
LiKang6688: Because it causes errors when doing yarn types:check

src/nebula-hooks/use-announce-and-translations.ts
haxxmaxx: Can you just use that code in this PR?

src/table/utils/__tests__/get-cell-renderer.spec.js
haxxmaxx: yes, but could you actually fix the errors instead? Since we want to merge this asap, I guess we could come back to it. just make sure we don't lose track of it then

src/nebula-hooks/use-announce-and-translations.ts
LiKang6688: you mean use 
```
interface ExtendedTranslator extends stardust.Translator {
  language: () => string;
}
```
right?
language: () => string; and language(): string; is same

src/table/utils/__tests__/get-cell-renderer.spec.js
haxxmaxx: hmm Is it the mockImplementation that is the problem?

LiKang6688: Screenshot 2022-08-09 at 14 03 26
Do you have a quick fix for the issue above?
I think we need to convert withSelections, withColumnStyling, and withStyling to typescript to fix it.
If no quick fix, I suggest that we look at it after merging this PR.

src/nebula-hooks/use-announce-and-translations.ts
haxxmaxx: Ah I missed that you are extending it already, then this is fine. not sure where it should live. But that is a minor problem. I could be merged like this for now

src/table/utils/__tests__/get-cell-renderer.spec.js
haxxmaxx: yeah, that was my thought too. It is fine like this for now

haxxmaxx (approved):

src/nebula-hooks/use-announce-and-translations.ts
LiKang6688: okay

src/table/utils/__tests__/get-cell-renderer.spec.js
LiKang6688: okay

PR #622, opened 2022-08-05
refactor: convert handle-accessibility to typescript

Donya-qlik:


src/table/utils/handle-accessibility.ts
```diff
@@ -81,28 +121,43 @@ export const handleResetFocus = ({
...
+  keyboard: stardust.Keyboard
+) => {
+  const targetElement = evt.currentTarget as HTMLDivElement;
+  if (keyboard?.enabled && !targetElement.contains(evt.relatedTarget as Node) && !shouldRefocus.current) {
+    targetElement.querySelector('#sn-table-announcer--01')!.innerHTML = '';
+    targetElement.querySelector('#sn-table-announcer--02')!.innerHTML = '';
     // Blur the table but not focus its parent element
     // when keyboard.active is false, this has no effect
-    keyboard.blur(false);
+    keyboard.blur?.(false);
```
Donya-qlik, line 137: blur and focusSelection need to be modified on `index.d.ts`.

src/table/hooks/use-focus-listener.ts
```diff
@@ -1,10 +1,12 @@
...
-import { useEffect } from 'react';
+import React, { useEffect } from 'react';
 import { handleFocusoutEvent } from '../utils/handle-accessibility';
 
 const useFocusListener = (
   tableWrapperRef: React.MutableRefObject<HTMLDivElement>,
-  shouldRefocus: boolean,
+  shouldRefocus: {
+    current: boolean;
+  },
```
haxxmaxx, line 9:
```suggestion
  shouldRefocus: React.MutableRefObject<boolean>,
```

src/table/utils/handle-accessibility.ts
```diff
@@ -1,11 +1,31 @@
...
+interface ICellFocusProps {
+  focusType: string;
+  cell: HTMLTableCellElement | undefined;
+}
+interface IHandleResetFocusProps {
+  focusedCellCoord: [number, number];
+  rootElement: HTMLDivElement;
+  shouldRefocus: {
+    current: boolean;
+  };
```
haxxmaxx, line 13:
```suggestion
  shouldRefocus: React.MutableRefObject<boolean>;
```

src/table/utils/handle-accessibility.ts
```diff
@@ -1,11 +1,31 @@
...
+  isSelectionMode: boolean;
+  setFocusedCellCoord: (coord: [number, number]) => void;
+  keyboard: stardust.Keyboard;
+  announce: (arg: { keys: string[] }) => void;
```
haxxmaxx, line 17: There is a type for this in types already

src/table/utils/handle-accessibility.ts
```diff
@@ -1,11 +1,31 @@
...
+  cell: HTMLTableCellElement | undefined;
+}
+interface IHandleResetFocusProps {
+  focusedCellCoord: [number, number];
+  rootElement: HTMLDivElement;
+  shouldRefocus: {
+    current: boolean;
+  };
+  isSelectionMode: boolean;
+  setFocusedCellCoord: (coord: [number, number]) => void;
```
haxxmaxx, line 15:
```suggestion
  setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>
```

src/table/utils/handle-accessibility.ts
```diff
@@ -1,11 +1,31 @@
...
+}
+
+export const getCellElement = (rootElement: HTMLDivElement, cellCoord: [number, number]) =>
   rootElement.getElementsByClassName('sn-table-row')[cellCoord[0]]?.getElementsByClassName('sn-table-cell')[
     cellCoord[1]
-  ];
+  ] as HTMLTableCellElement;
 
-export const findCellWithTabStop = (rootElement) => rootElement.querySelector("td[tabindex='0'], th[tabindex='0']");
+export const findCellWithTabStop = (rootElement: Element) =>
```
haxxmaxx, line 25: add return type

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
   }
 };
 
-export const removeAndFocus = (newCoord, rootElement, setFocusedCellCoord, keyboard) => {
+export const removeAndFocus = (
+  newCoord: [number, number],
+  rootElement: Element,
```
haxxmaxx, line 53: This should be consistent with other usage of rootElement

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+) => {
   updateFocus({ focusType: 'removeTab', cell: findCellWithTabStop(rootElement) });
   setFocusedCellCoord(newCoord);
-  keyboard.enabled && keyboard.focus();
+  keyboard.enabled && keyboard.focus?.();
 };
 
-export const handleClickToFocusBody = (cell, rootElement, setFocusedCellCoord, keyboard, totalsPosition) => {
+export const handleClickToFocusBody = (
+  cell: { rawRowIdx: number; rawColIdx: number },
```
haxxmaxx, line 63: There is a type for this in types already

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+  rootElement: Element,
```
haxxmaxx, line 64: needs to be consistent with other usage of rootElement

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+  setFocusedCellCoord: (coord: [number, number]) => void,
```
haxxmaxx, line 65:
```suggestion
  setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>
```

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+  keyboard: stardust.Keyboard,
+  totalsPosition: string
```
haxxmaxx, line 67: There is a type for this in types. It limits it to three different values, not just any string

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+) => {
   const { rawRowIdx, rawColIdx } = cell;
   const adjustedRowIdx = totalsPosition === 'top' ? rawRowIdx + 2 : rawRowIdx + 1;
   removeAndFocus([adjustedRowIdx, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
 };
 
-export const handleClickToFocusHead = (columnIndex, rootElement, setFocusedCellCoord, keyboard) => {
+export const handleClickToFocusHead = (
+  columnIndex: number,
+  rootElement: Element,
```
haxxmaxx, line 76: same here, should be consistent

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+  setFocusedCellCoord: (coord: [number, number]) => void,
```
haxxmaxx, line 77: see above

src/table/utils/handle-accessibility.ts
```diff
@@ -28,23 +48,43 @@ export const updateFocus = ({ focusType, cell }) => {
...
+  keyboard: stardust.Keyboard
+) => {
   removeAndFocus([0, columnIndex], rootElement, setFocusedCellCoord, keyboard);
 };
 
-export const handleMouseDownLabelToFocusHeadCell = (evt, rootElement, columnIndex) => {
+export const handleMouseDownLabelToFocusHeadCell = (
+  evt: BaseSyntheticEvent,
```
haxxmaxx, line 84: This is an event from onMouseDown so I think it is a React.MouseEvent

src/table/utils/handle-accessibility.ts
```diff
@@ -57,10 +97,10 @@ export const handleResetFocus = ({
   setFocusedCellCoord,
   keyboard,
   announce,
-}) => {
+}: IHandleResetFocusProps) => {
   updateFocus({ focusType: 'removeTab', cell: findCellWithTabStop(rootElement) });
   // If you have selections ongoing, you want to stay on the same column
-  const cellCoord = isSelectionMode ? [1, focusedCellCoord[1]] : [0, 0];
+  const cellCoord = (isSelectionMode ? [1, focusedCellCoord[1]] : [0, 0]) as [number, number];
```
haxxmaxx, line 103: try this
```suggestion
  const cellCoord: [number, number] = isSelectionMode ? [1, focusedCellCoord[1]] : [0, 0];
```

src/table/utils/handle-accessibility.ts
```diff
@@ -81,28 +121,43 @@ export const handleResetFocus = ({
...
 
-export const handleFocusoutEvent = (evt, shouldRefocus, keyboard) => {
-  if (keyboard.enabled && !evt.currentTarget.contains(evt.relatedTarget) && !shouldRefocus.current) {
-    evt.currentTarget.querySelector('#sn-table-announcer--01').innerHTML = '';
-    evt.currentTarget.querySelector('#sn-table-announcer--02').innerHTML = '';
+export const handleFocusoutEvent = (
+  evt: FocusEvent,
+  shouldRefocus: {
+    current: boolean;
+  },
```
haxxmaxx, line 128: same as above comment

src/table/utils/handle-accessibility.ts
haxxmaxx: It needs to be changed in nebula. you can add a ts-ignore here. see other usages of keyboard.blur, you can copy that comment

src/table/utils/handle-accessibility.ts
```diff
@@ -81,28 +121,43 @@ export const handleResetFocus = ({
...
+    targetElement.querySelector('#sn-table-announcer--02')!.innerHTML = '';
     // Blur the table but not focus its parent element
     // when keyboard.active is false, this has no effect
-    keyboard.blur(false);
+    keyboard.blur?.(false);
   }
 };
 
-export const focusSelectionToolbar = (element, keyboard, last) => {
+export const focusSelectionToolbar = (element: HTMLTableCellElement, keyboard: stardust.Keyboard, last: boolean) => {
```
haxxmaxx, line 141: element is not necessarily a HTMLTableCellElement. It can come from the last button as well

src/table/utils/handle-accessibility.ts
```diff
@@ -81,28 +121,43 @@ export const handleResetFocus = ({
...
   const clientConfirmButton = element
     .closest('.qv-object-wrapper')
     ?.querySelector('.sel-toolbar-confirm')?.parentElement;
   if (clientConfirmButton) {
     clientConfirmButton.focus();
     return;
   }
-  keyboard.focusSelection(last);
+  keyboard.focusSelection?.(last);
```
haxxmaxx, line 149: add a ts.ignore if this is not on the stardust API

src/table/utils/handle-accessibility.ts
```diff
@@ -81,28 +121,43 @@ export const handleResetFocus = ({
...
 };
 
-export const announceSelectionState = (announce, nextCell, isSelectionMode) => {
+export const announceSelectionState = (
+  announce: (arg: { keys: string[] }) => void,
```
haxxmaxx, line 153: use type from the types file

src/table/utils/handle-accessibility.ts
```diff
@@ -81,28 +121,43 @@ export const handleResetFocus = ({
...
+  nextCell: {
+    classList: {
+      contains: (token: string) => boolean;
+    };
+  },
```
haxxmaxx, line 158: Pretty sure this is some kind of element

src/table/utils/handle-accessibility.ts
```diff
@@ -1,11 +1,31 @@
...
+import { stardust } from '@nebula.js/stardust';
+import { BaseSyntheticEvent } from 'react';
+
+interface ICellFocusProps {
+  focusType: string;
+  cell: HTMLTableCellElement | undefined;
+}
+interface IHandleResetFocusProps {
+  focusedCellCoord: [number, number];
+  rootElement: HTMLDivElement;
```
haxxmaxx, line 10: i'm not sure this is always a div. Do you know? might be different depending on how you mount it

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -1,71 +1,78 @@
...
 import * as handleAccessibility from '../handle-accessibility';
 
 describe('handle-accessibility', () => {
-  let cell;
-  let rootElement;
-  let focusedCellCoord;
-  let setFocusedCellCoord;
-  let keyboard;
+  let focusType: string;
+  let cell: HTMLTableCellElement | undefined;
```
LiKang6688, line 7: I do not think that undefined is needed here.

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -100,7 +107,7 @@ describe('handle-accessibility', () => {
     });
 
     it('should return null', () => {
-      cell = elementCreator('div', '-1');
+      cell = elementCreator('div', '-1') as HTMLTableCellElement;
```
LiKang6688, line 110: HTMLTableCellElement starts with <table> not <div>. so I think it is HTMLDivElement

src/table/utils/__tests__/handle-accessiblity.spec.ts
haxxmaxx: That is the type that the function wants. And we do a sanity check in there, early return on !cell. So I think it can be undefined

src/table/utils/handle-accessibility.ts
Donya-qlik: Do you mean line 26 `as HTMLTableCellElement` ?

src/table/utils/handle-accessibility.ts
LiKang6688: it is `Announce` form the types file

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -1,71 +1,78 @@
...
+import { stardust } from '@nebula.js/stardust';
 import * as handleAccessibility from '../handle-accessibility';
 
 describe('handle-accessibility', () => {
-  let cell;
-  let rootElement;
-  let focusedCellCoord;
-  let setFocusedCellCoord;
-  let keyboard;
+  let focusType: string;
```
haxxmaxx, line 6: this is only used for updateFocus, so you can move it back to inside that describe

src/table/utils/__tests__/handle-accessiblity.spec.ts
Donya-qlik: Yes, Max explanation is true, there is a sanity check for that.

src/table/hooks/use-focus-listener.ts
```diff
@@ -1,10 +1,12 @@
 import { stardust } from '@nebula.js/stardust';
-import { useEffect } from 'react';
+import React, { useEffect } from 'react';
```
LiKang6688, line 2: Why React is needed in this file?

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -279,28 +290,34 @@ describe('handle-accessibility', () => {
   });
 
   describe('handleFocusoutEvent', () => {
-    let containsRelatedTarget;
-    let evt;
-    let shouldRefocus;
-    let announcement1;
-    let announcement2;
+    let containsRelatedTarget: boolean;
+    let evt0: FocusEvent;
```
haxxmaxx, line 294: Call it focusoutEvent or something, evt0 is a bit arbitrary

src/table/utils/handle-accessibility.ts
LiKang6688: should be `HTMLElement `

src/table/utils/handle-accessibility.ts
LiKang6688: @haxxmaxx I do not think it is necessary. It is implicitly set to HTMLTableCellElement already

LiKang6688: @Donya-qlik 
`export const findCellWithTabStop = (rootElement: Element): HTMLTableCellElement =>`

haxxmaxx: ahh I did not know ts figures that out, nice. Then no need to change anything here :)

haxxmaxx:
I'll have a closer look at the test file when you had a look at the comments and updated

src/table/hooks/use-focus-listener.ts
Donya-qlik: ... I thought I'm adding it because of line 6 but seems it works without importing React anyway. I'll remove it

Donya-qlik:
Not sure what went wrong... I expected git 'files changing part' keeps the transitions from js to ts for reviewing but I see now the js file is totally deleted  and a new file is replaced, sorry for this destruction 😣

src/table/utils/handle-accessibility.ts
Donya-qlik: Fixed.

src/table/utils/handle-accessibility.ts
Donya-qlik: Replaced it with HTMLElement.

src/table/hooks/use-focus-listener.ts
haxxmaxx: this isn't fixed

src/table/utils/handle-accessibility.ts
```diff
@@ -0,0 +1,162 @@
...
+      cell.setAttribute('tabIndex', '-1');
+      break;
+    default:
+      break;
+  }
+};
+
+export const removeAndFocus = (
+  newCoord: [number, number],
+  rootElement: HTMLDivElement,
```
haxxmaxx, line 52:
```suggestion
  rootElement: HTMLElement,
```

src/table/utils/handle-accessibility.ts
```diff
@@ -0,0 +1,162 @@
...
+  keyboard: stardust.Keyboard
+) => {
+  updateFocus({ focusType: 'removeTab', cell: findCellWithTabStop(rootElement) });
+  setFocusedCellCoord(newCoord);
+  keyboard.enabled && keyboard.focus?.();
+};
+
+export const handleClickToFocusBody = (
+  cell: TableCell,
+  rootElement: HTMLDivElement,
```
haxxmaxx, line 63:
```suggestion
  rootElement: HTMLElement,
```

src/table/utils/handle-accessibility.ts
```diff
@@ -0,0 +1,162 @@
...
+  totalsPosition: TotalsPosition
+) => {
+  const { rawRowIdx, rawColIdx } = cell;
+  const adjustedRowIdx = totalsPosition === 'top' ? rawRowIdx + 2 : rawRowIdx + 1;
+  removeAndFocus([adjustedRowIdx, rawColIdx], rootElement, setFocusedCellCoord, keyboard);
+};
+
+export const handleClickToFocusHead = (
+  columnIndex: number,
+  rootElement: HTMLDivElement,
```
haxxmaxx, line 75:
```suggestion
  rootElement: HTMLElement,
```

src/table/utils/handle-accessibility.ts
```diff
@@ -0,0 +1,162 @@
...
+  setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>,
+  keyboard: stardust.Keyboard
+) => {
+  removeAndFocus([0, columnIndex], rootElement, setFocusedCellCoord, keyboard);
+};
+
+export const handleMouseDownLabelToFocusHeadCell = (
+  evt: BaseSyntheticEvent,
+  rootElement: HTMLDivElement,
```
haxxmaxx, line 84:
```suggestion
  rootElement: HTMLElement,
```

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -1,71 +1,79 @@
...
 
 describe('handle-accessibility', () => {
-  let cell;
-  let rootElement;
-  let focusedCellCoord;
-  let setFocusedCellCoord;
-  let keyboard;
+  let cell: HTMLTableCellElement | undefined;
+  let keyboard: stardust.Keyboard;
+  let rootElement: HTMLDivElement;
```
haxxmaxx, line 9:
```suggestion
  let rootElement: HTMLElement;
```

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -279,28 +286,34 @@ describe('handle-accessibility', () => {
...
-    let containsRelatedTarget;
-    let evt;
-    let shouldRefocus;
-    let announcement1;
-    let announcement2;
+    let containsRelatedTarget: boolean;
+    let focusoutEvent: FocusEvent;
+    let shouldRefocus: {
+      current: boolean;
+    };
```
haxxmaxx, line 293: should be `React.MutableRefObject<boolean>`

haxxmaxx:
> Not sure what went wrong... I expected git 'files changing part' keeps the transitions from js to ts for reviewing but I see now the js file is totally deleted and a new file is replaced, sorry for this destruction 😣

That is fine, it has happened to at least one of my PRs. I think it is when you've changed too much it doesn't consider it a rename anymore

src/table/utils/handle-accessibility.ts
haxxmaxx: hmm yes that is true, for the keyboard events as well it seems like. I will change in my PR

haxxmaxx (approved):

src/table/utils/__tests__/handle-accessiblity.spec.ts
LiKang6688: yeah, good

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -157,31 +162,31 @@ describe('handle-accessibility', () => {
...
     });
   });
 
   describe('handleResetFocus', () => {
-    let shouldRefocus;
-    let isSelectionMode;
-    let announce;
+    let shouldRefocus: React.MutableRefObject<boolean>;
+    let isSelectionMode: boolean;
+    let announce: AnnounceFn;
```
LiKang6688, line 189: AnnounceFn is Announce in the main branch. Please merge the main branch to your branch.

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -236,21 +243,21 @@ describe('handle-accessibility', () => {
       rootElement = {
         getElementsByClassName: () => [row, row],
         querySelector: () => cell,
-      };
+      } as unknown as HTMLDivElement;
```
LiKang6688, line 246:
```suggestion
      } as unknown as HTMLElement;
```

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -236,21 +243,21 @@ describe('handle-accessibility', () => {
...
 
     it('should announce cell content and selection status for non selected first cell after focusing on it', () => {
-      cell = { ...cell, textContent: '#something' };
+      cell = { ...cell, textContent: '#something' } as HTMLTableCellElement;
       const row = { getElementsByClassName: () => [cell, cell] };
       rootElement = {
         getElementsByClassName: () => [row, row],
         querySelector: () => cell,
-      };
+      } as unknown as HTMLDivElement;
```
LiKang6688, line 260:
```suggestion
      } as unknown as HTMLElement;
```

src/table/utils/handle-accessibility.ts
haxxmaxx: Looking at it again it seems like the devtools are showing the wrong type, no idea why. If you look at the type definitions for the different react types, the properties on the event match to React.MouseEvent. So set it to that

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -260,15 +267,15 @@ describe('handle-accessibility', () => {
     });
 
     it('should announce cell content and selection status for selected first cell after focusing on it', () => {
-      const tmpCell = global.document.createElement('td');
-      tmpCell.classList.add('selected');
+      const tmpCell = global.document.createElement('td') as HTMLDivElement;
```
LiKang6688, line 270:
```suggestion
      const tmpCell = global.document.createElement('td');
```
I do not think that you need to convert the HTMLTableCellElement to HTMLDivElement here

src/table/utils/__tests__/handle-accessiblity.spec.ts
Donya-qlik: Sorry, didn't get what you mean? It's already available in types in my branch ...

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -260,15 +267,15 @@ describe('handle-accessibility', () => {
...
+      tmpCell?.classList.add('selected');
 
-      cell = { ...cell, classList: tmpCell.classList, textContent: '#something' };
+      cell = { ...cell, classList: tmpCell?.classList, textContent: '#something' } as HTMLTableCellElement;
       const row = { getElementsByClassName: () => [cell, cell] };
       rootElement = {
         getElementsByClassName: () => [row, row],
         querySelector: () => cell,
-      };
+      } as unknown as HTMLDivElement;
```
LiKang6688, line 278:
```suggestion
      } as unknown as HTMLElement;
```

src/table/utils/__tests__/handle-accessiblity.spec.ts
LiKang6688: I mean that your branch needs to be in sync with the main branch. In the main branch, there is no AnnounceFn anymore.

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -309,41 +320,41 @@ describe('handle-accessibility', () => {
...
     it('should not call blur when shouldRefocus is true', () => {
       shouldRefocus.current = true;
 
-      handleAccessibility.handleFocusoutEvent(evt, shouldRefocus, keyboard);
+      handleAccessibility.handleFocusoutEvent(focusoutEvent, shouldRefocus, keyboard);
       expect(keyboard.blur).not.toHaveBeenCalled();
     });
 
-    it('should not call blur when keyboard.enabled is falsey', () => {
+    it('should not call blur when keyboard.enabled is falsy', () => {
```
LiKang6688, line 334:
```suggestion
    it('should not call blur when keyboard.enabled is false', () => {
```

src/table/utils/handle-accessibility.ts
```diff
@@ -0,0 +1,162 @@
+import { BaseSyntheticEvent } from 'react';
+import { stardust } from '@nebula.js/stardust';
+import { AnnounceFn, TableCell, TotalsPosition } from '../../types';
+
+interface ICellFocusProps {
+  focusType: string;
+  cell: HTMLTableCellElement | undefined;
+}
+interface IHandleResetFocusProps {
```
LiKang6688, line 9: It would be very good to put these interfaces to the types.ts file to shorten the line of code of this file.

LiKang6688: And we do not add 'I' at the beginning of an interface.

src/table/utils/__tests__/handle-accessiblity.spec.ts
Donya-qlik: Aha in main it is still Announce. I'll fix it.

LiKang6688: Thanks!

src/table/utils/handle-accessibility.ts
haxxmaxx: Maybe we should though. Probably best to remove `I` now for consistency, but we should look into best practice later

src/table/utils/handle-accessibility.ts
Donya-qlik: Sure, I'll set it to MouseEvent.

src/table/utils/__tests__/handle-accessiblity.spec.ts
haxxmaxx: It might be undefined when it is not true, so falsy should cover both false and undefined

LiKang6688: I do not follow. I see that `keyboard.enabled = false;` below. So 'should not call blur when keyboard.enabled is false' makes sense to me. And I do not see `keyboard.enabled = undefined`.

src/table/utils/handle-accessibility.ts
LiKang6688: MouseEvent extends UIEvent extends SyntheticEvent extends BaseSyntheticEvent. so MouseEvent is more accurate in this case.

src/table/utils/__tests__/handle-accessiblity.spec.ts
Donya-qlik: To be honest what Li said is true. However falsy covers both false and undefined but in our cases undefined hasn't been considered as I checked. It was a written testcase, I just corrected the dictation and was not aware of whether it could be false/undefined. I think we better keep 'false' instead of falsy.

haxxmaxx: yeah it should probably be undefined, since the real enabled is undefined and not false AFAIK. This is not very important though so for this PR it is not super relevant

LiKang6688 (approved):

src/types.ts
```diff
@@ -255,3 +255,16 @@ export type TableData = {
   columns: Column[];
   totalsPosition: TotalsPosition;
```
LiKang6688, line 256: TotalsPosition from 247 line moved to above. Then should be fine.

Donya-qlik: Yes I fixed it in the last commit

PR #626, opened 2022-08-10
refactor: custom hooks

haxxmaxx:


src/table/hooks/use-did-update-effect.ts
```diff
@@ -0,0 +1,14 @@
...
+export default function useDidUpdateEffect(effect: () => void, deps: React.DependencyList) {
+  const didMountRef = useRef(false);
+
+  useEffect(() => {
+    if (didMountRef.current) {
+      effect();
+    } else {
+      didMountRef.current = true;
+    }
+  }, deps);
```
haxxmaxx, line 13: lint does not like the deps being added like this. I tried `[...deps]` and some other things but it says it is not an array literal no matter what. If there is a better way of doing this, pls let me know.

src/table/hooks/use-scroll-listener.ts
```diff
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
 import { handleHorizontalScroll } from '../utils/handle-scroll';
 
-const useScrollListener = (tableContainerRef: React.MutableRefObject<HTMLDivElement>, direction: string) => {
+const useScrollListener = (tableContainerRef: React.MutableRefObject<HTMLElement>, direction: string) => {
```
haxxmaxx, line 4: changing to HTMLElement for consistency

LiKang6688: There are many
Screenshot 2022-08-10 at 16 11 05
To be precise, i think it is  HTMLDivElement

src/table/hooks/use-focus-listener.ts
```diff
@@ -1,12 +1,17 @@
+import { stardust } from '@nebula.js/stardust';
 import { useEffect } from 'react';
 import { handleFocusoutEvent } from '../utils/handle-accessibility';
 
-const useFocusListener = (tableWrapperRef, shouldRefocus, keyboard) => {
+const useFocusListener = (
+  tableWrapperRef: React.MutableRefObject<HTMLElement>,
```
LiKang6688, line 6: i prefer to use HTMLDivElement

src/table/hooks/use-did-update-effect.ts
LiKang6688: yes, here the react-hooks lint cannot know what deps conclude, so it complains.
then I guess that we need to do something for the react-hooks lint

src/table/hooks/use-focus-listener.ts
haxxmaxx: yeah both seem to be divs so we can use that on both container and wrapper

src/table/hooks/use-did-update-effect.ts
haxxmaxx: I've disabled it, so it works at least

haxxmaxx: It seems like you have to use the syntax that you have seperate variable that you put in the array, in the useEffect call. Thus we can't satisfy the rule since we don't know what the deps are. Spreading doesn't help either :/

LiKang6688 (approved):

PR #627, opened 2022-08-10
fix: table moves up when doing selection

LiKang6688:
Head:
when a focused cell is updated, the scroll position is updated accordingly.

abc! (3)

Current:
only when using the keyboard to move up, the scroll position is updated accordingly

https://user-images.githubusercontent.com/4471489/183973096-02159b34-3100-4c7b-80e1-58b718798e20.mov

Limitation:

For a very narrow table, the best we can do is that arrow up scroll behavior is the same as the arrow down scroll behavior
Head:

https://user-images.githubusercontent.com/4471489/183924745-1f1c290b-405a-46bb-8bf1-590a0b49d4ef.mov

Current:

https://user-images.githubusercontent.com/4471489/183974003-9cf9f391-8441-4fa5-82d9-12afa8a91673.mov

### DIFF
--- a/src/table/utils/__tests__/handle-scroll.spec.js
+++ b/src/table/utils/__tests__/handle-scroll.spec.js
@@ -164,7 +164,7 @@ describe('handle-scroll', () => {
         },
       };
       // targetOffsetTop = tableContainer.current.scrollTop - cell.offsetHeight - tableHead.offsetHeight;
-      // 700 - 100 - 128 = 472 => so our scrollTo function might be called with 472
+      // 700 - 100 - 128 = 472 => so our scrollTo function is called with 472
       const targetOffsetTop = 472;
 
       handleNavigateTop(cellCoord, rootElement);
--- a/src/table/utils/__tests__/handle-scroll.spec.js
+++ b/src/table/utils/__tests__/handle-scroll.spec.js
@@ -124,7 +124,7 @@ describe('handle-scroll', () => {
       };
     });
 
-    it('should not do anything when ref is not setup yet', () => {
+    it('should not do anything when rootElement is not setup yet', () => {
       handleNavigateTop(cellCoord, rootElement);
       expect(scrollTo).not.toHaveBeenCalled();
     });
--- a/src/table/utils/__tests__/handle-scroll.spec.js
+++ b/src/table/utils/__tests__/handle-scroll.spec.js
@@ -112,38 +112,43 @@ describe('handle-scroll', () => {
   describe('handleNavigateTop', () => {
     let rowHeight;
     let scrollTo;
-    let tableContainerRef;
-    let focusedCellCoord;
+    let cellCoord;
     let rootElement;
 
     beforeEach(() => {
       rowHeight = 100;
       scrollTo = jest.fn();
-      tableContainerRef = { current: { scrollTo } };
-      focusedCellCoord = [0, 0];
-      rootElement = {};
+      cellCoord = [0, 0];
+      rootElement = {
+        getElementsByClassName: () => [{}],
+      };
     });
 
     it('should not do anything when ref is not setup yet', () => {
-      tableContainerRef.current = {};
-
-      handleNavigateTop({ tableContainerRef, focusedCellCoord, rootElement });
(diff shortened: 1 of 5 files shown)


src/table/utils/handle-key-press.js
```diff
@@ -164,6 +165,7 @@ export const bodyHandleKeyPress = ({
   switch (evt.key) {
     case 'ArrowUp':
     case 'ArrowDown': {
+      evt.key === 'ArrowUp' && handleNavigateTop([cell.rawRowIdx, cell.rawColIdx], rootElement);
```
haxxmaxx, line 168: It is the current cell and not the next cell we should look at?

src/table/utils/__tests__/handle-scroll.spec.js
```diff
@@ -112,38 +112,43 @@ describe('handle-scroll', () => {
...
-      tableContainerRef = { current: { scrollTo } };
-      focusedCellCoord = [0, 0];
-      rootElement = {};
+      cellCoord = [0, 0];
+      rootElement = {
+        getElementsByClassName: () => [{}],
+      };
     });
 
     it('should not do anything when ref is not setup yet', () => {
```
haxxmaxx, line 127: There is no ref anymore, so should be a better description

LiKang6688: right

src/table/utils/handle-key-press.js
LiKang6688: it is the coordinate for the next cell when totals is top and the current cell when totals is bottom and noTotals

LiKang6688: as you see in the video, I think it has a better outcome.

haxxmaxx: but wouldn't it behave a bit differently when you have totals on top or not?

LiKang6688: yes, I also thought it should be different but it is not. I am looking into it.

LiKang6688: do not really know why. I updated the current video.

LiKang6688: Now I understand it and it is correct.

src/table/utils/__tests__/handle-scroll.spec.js
```diff
@@ -159,12 +164,12 @@ describe('handle-scroll', () => {
         },
       };
       // targetOffsetTop = tableContainer.current.scrollTop - cell.offsetHeight - tableHead.offsetHeight;
-      // 700 - 100 - 128 = 472 => so our scrollTo function migth be called with 600
+      // 700 - 100 - 128 = 472 => so our scrollTo function might be called with 472
```
haxxmaxx, line 167: Might be called? I mean it should be called right? otherwise it is not really a test

src/table/utils/handle-key-press.js
haxxmaxx, line 168: Add or extend a test to see if `handleNavigateTop` has been called

LiKang6688:
> If I understand the limitation, it only happens when your header is bigger than the whole body. Should be a really uncommon use case

Right, an uncommon one, and it is good that we know it in advance.

src/table/utils/handle-key-press.js
LiKang6688: Good point. Gonna do

src/table/utils/__tests__/handle-scroll.spec.js
LiKang6688: I think you are right, let me fix it

src/table/utils/handle-key-press.js
LiKang6688: Done

src/table/utils/__tests__/handle-scroll.spec.js
LiKang6688: Done

haxxmaxx (approved):

haxxmaxx:
Really nice that we could solve this without introducing forwardRef, great work

PR #628, opened 2022-08-11
refactor: convert styling-utils to typescript and throw away invalid colors to the columns

haxxmaxx:
- stylingInfo -> stylingIDs since it only contains what props have been set, not the actual value
- isSet -> isColorSet
- remove th jsdocs for converted functions. I think the arg types and return types should be explanatory enough together with a description of the function. Also unsure if typescript and defining args/returns with jsdocs are compatible together
- adding the default padding back. I put it in the object properties first since we are missing it there but that is better to do in a seperate PR
- Sanity checking so we don't add invalid colors to the columns
  - This means that we actually need to update the screenshots for scenario 5. It has invalid props and use to add those to the css. Now it will reject them and keep the other styling


src/object-properties.js
```diff
@@ -147,13 +147,15 @@ const properties = {
  * @property {boolean=} hoverEffect - Toggles hover effect
  * @property {PaletteColor=} hoverColor - Background hover color. Uses `#f4f4f4` if no hover colors are set, is transparent if only `hoverFontColor` is set
  * @property {PaletteColor=} hoverFontColor - When only `hoverColor` is set, this is adjusted to either `#f4f4f4` or `#ffffff` for optimal contrast
+ * @property {string=} padding - padding for the body cells on css format
```
haxxmaxx, line 150: This is available if you write the props yourself, but wont exist in styling panel later. I guess we be documented

src/table/utils/__tests__/styling-utils.spec.ts
```diff
@@ -103,7 +102,7 @@ describe('styling-utils', () => {
       });
     });
     it('should return styling with fontSize and padding', () => {
-      styleObj.fontColor = null;
+      styleObj.fontColor = undefined;
```
haxxmaxx, line 105: this shouldn't be null

src/table/utils/styling-utils.ts
```diff
@@ -27,10 +31,17 @@ export const SELECTION_STYLING = {
   },
 };
 
-// Both index !== -1 and color !== null must be true for the property to be set
-export const isSet = (prop) => prop && JSON.stringify(prop) !== JSON.stringify({ index: -1, color: null });
+/**
+ * Determines if a palette color is set or not. Both index !== -1 and color !== null must be true for it to be unset
+ */
+export const isColorSet = (prop: PaletteColor | undefined): boolean =>
```
LiKang6688, line 37:
```suggestion
export const isColorSet = (prop?: PaletteColor): boolean =>
```

src/table/utils/styling-utils.ts
```diff
@@ -27,10 +31,17 @@ export const SELECTION_STYLING = {
...
+  !!prop && JSON.stringify(prop) !== JSON.stringify({ index: -1, color: null });
```
LiKang6688, line 38:
```suggestion
  prop && JSON.stringify(prop) !== JSON.stringify({ index: -1, color: null });
```

src/table/utils/styling-utils.ts
```diff
@@ -40,34 +51,49 @@ export function getPadding(styleObj, defaultPadding) {
...
   const color = theme.getStyle('object', `straightTable.${objetName}`, 'color');
   const fontSize = theme.getStyle('object', `straightTable.${objetName}`, 'fontSize');
 
-  const baseStyle = {
+  const baseStyle: GeneratedStyling = {
     fontFamily,
-    color: isSet(styleObj?.fontColor) ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj.fontColor) : color,
-    fontSize: styleObj?.fontSize || fontSize,
+    color: isColorSet(styleObj?.fontColor) ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj.fontColor) : color,
+    fontSize: (styleObj?.fontSize && `${styleObj.fontSize}px`) || fontSize,
```
LiKang6688, line 80: Does `theme.getStyle('object', `straightTable.${objetName}`, 'fontSize');` contain 'px' as well?

src/types.ts
```diff
@@ -88,6 +92,7 @@ export interface TableLayout extends EngineAPI.IGenericHyperCubeLayout {
     position: 'top' | 'bottom' | 'noTotals';
     label: string;
   };
+  components?: any;
```
LiKang6688, line 95: why 'any' here?

src/table/utils/__tests__/styling-utils.spec.ts
```diff
@@ -339,64 +342,64 @@ describe('styling-utils', () => {
...
-      stylingInfo = [stylingInfo[0]];
+      stylingIDs = [stylingIDs[0]];
 
-      const columnStyle = getColumnStyle(styling, qAttrExps, stylingInfo);
+      const columnStyle = getColumnStyle(styling, qAttrExps, stylingIDs);
       expect(columnStyle.backgroundColor).toBe('rgb(221,221,221)');
       expect(columnStyle.color).toBe(STYLING_DEFAULTS.FONT_COLOR);
     });
     it('should return styling unchanged', () => {
-      qAttrExps = undefined;
```
LiKang6688, line 376: i think it is also good to check qAttrExps is something then should return styling unchanged

src/table/utils/styling-utils.ts
haxxmaxx: no, need to make sure the return is a boolean. prop can be undefined

LiKang6688: I see, you are right

src/table/utils/styling-utils.ts
haxxmaxx: yes it should be

LiKang6688: ok, good

src/table/utils/styling-utils.ts
haxxmaxx: well, we actually want to enforce that you pass an argument. And sometimes it is undefined and we want to cover that case. But `isColorSet()` does not make sense

src/types.ts
haxxmaxx: hmm I think I forgot this one

src/table/utils/__tests__/styling-utils.spec.ts
haxxmaxx: what do you mean? if qAttrExps is set then you actually have column styling and it should change, that is what the tests above do

haxxmaxx: actually we should probably have another test that tests what happens when you qAttrExps is undefined

LiKang6688: yes, I am considering adding a new test case to cover a case when qAttrExps is not set or set something else then it should return unchanged styling

haxxmaxx: so I tested what the layout becomes if you only add the id but no expression in the properties. You get a layout with qAttrExps and it has qValues as well. but it does not have qText. So need to change the logic a bit so it does not change anything unless there is a qText

haxxmaxx: Turns out that getColumnStyle never runs if stylingIDs is empty. So no need to sanity check for that. We do need to check that qText is there and it resolved to a valid color. So added checks + tests for that instead

src/types.ts
haxxmaxx: added

src/table/utils/__tests__/styling-utils.spec.ts
```diff
@@ -294,7 +297,7 @@ describe('styling-utils', () => {
     });
     // Only checking hover properties from here on
     it('should return styling with no hoverBackgroundColor and the specified hoverFontColor', () => {
-      layout.components[0].content.hoverFontColor.index = 1;
+      if (layout.components?.[0].content?.hoverFontColor) layout.components[0].content.hoverFontColor.index = 1;
```
haxxmaxx, line 300: had to do a sanity check here since components, content and hoverFontColor can be undefined. Note that we only change the layout in the tests, so we don't have to deal with this check in that case.

LiKang6688:
Rendering tests fail?

haxxmaxx:
> Rendering tests fail?

hmm yes looking into it

src/table/utils/__tests__/styling-utils.spec.ts
LiKang6688: Nice

LiKang6688 (approved):
awesome

PR #629, opened 2022-08-12
refactor: convert handle-scroll to typescript

LiKang6688:


src/table/utils/handle-scroll.ts
```diff
@@ -35,18 +35,18 @@ export const handleNavigateTop = (cellCoord, rootElement) => {
   if (x < MIN_ROW_COUNT) {
     tableContainer.scrollTo({
       top: 0,
-      behavior: 'instant',
+      behavior: 'auto',
```
haxxmaxx, line 38: how does the behavior differ here?

src/table/utils/handle-scroll.ts
```diff
@@ -35,18 +35,18 @@ export const handleNavigateTop = (cellCoord, rootElement) => {
...
     });
   } else {
-    const tableHead = rootElement.getElementsByClassName('sn-table-head-cell')[0];
+    const tableHead = rootElement.getElementsByClassName('sn-table-head-cell')[0] as HTMLElement;
     const rowElements = rootElement.getElementsByClassName('sn-table-row');
-    const cell = rowElements[x]?.getElementsByClassName('sn-table-cell')[y];
+    const cell = rowElements[x]?.getElementsByClassName('sn-table-cell')[y] as HTMLElement;
```
haxxmaxx, line 43: the sanity check here is a bit strange. There is a `?.` and then assume that cell exist in the if block below. Is there a case where cell is undefined?

src/table/utils/selections-utils.ts
```diff
@@ -35,7 +35,7 @@ type AddSelectionListenersArgs = {
   selectionDispatch: React.Dispatch<Action>;
   setShouldRefocus(): void;
   keyboard: stardust.Keyboard;
-  tableWrapperRef: React.MutableRefObject<HTMLElement>;
+  tableWrapperRef: React.MutableRefObject<HTMLDivElement>;
```
haxxmaxx, line 38: 👍

src/table/utils/handle-scroll.ts
LiKang6688: no difference. 'instant' is just not available in typescript
Screenshot 2022-08-12 at 11 11 30

haxxmaxx: ok

src/table/utils/handle-scroll.ts
LiKang6688: rowElements = [rootElement,  rootElement];
rowElements[2] is undefined
undefiiend.?getElementsByClassName('sn-table-cell')[y] is undefined

Gonna fix

haxxmaxx (approved):

PR #630, opened 2022-08-12
fix: type error

haxxmaxx:
Also added a test for null

### DIFF
--- a/src/table/utils/color-utils.ts
+++ b/src/table/utils/color-utils.ts
@@ -46,8 +46,8 @@ export function resolveToRGBAorRGB(input: string): string {
   return 'none';
 }
 
-export function isDarkColor(color: string | undefined): boolean {
-  if (color === undefined) return false;
+export function isDarkColor(color: string | undefined | null): boolean {
+  if (color == null) return false;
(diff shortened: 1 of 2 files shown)


src/table/utils/color-utils.ts
```diff
@@ -46,8 +46,8 @@ export function resolveToRGBAorRGB(input: string): string {
   return 'none';
 }
 
-export function isDarkColor(color: string | undefined): boolean {
-  if (color === undefined) return false;
+export function isDarkColor(color: string | undefined | null): boolean {
+  if (color == null) return false;
```
LiKang6688, line 50: `undefined == null` is true, not sure if there is any side effect.

LiKang6688 (approved):

src/table/utils/color-utils.ts
haxxmaxx: that is the intention, so it catches both undefined and null

LiKang6688: Yes, i know. (color === undefined && color === null) is more common. But it should be fine

PR #634, opened 2022-08-12
refactor: convert index to typescript

LiKang6688:


src/types.ts
```diff
@@ -23,6 +23,16 @@ export interface ExtendedTranslator extends stardust.Translator {
   language(): string;
 }
 
+export interface Galaxy {
```
haxxmaxx, line 26: is galaxy a term that we use for this in other places? I don't think we should introduce terms like this in the nebula eco-system unless it is thought through.

src/types.ts
```diff
@@ -23,6 +23,16 @@ export interface ExtendedTranslator extends stardust.Translator {
...
+  translator: ExtendedTranslator;
```
haxxmaxx, line 27: is the translator on here? I thought we got it from the useTheme hook?

src/types.ts
```diff
@@ -101,3 +111,15 @@ export interface Column {
   stylingInfo: string[];
   sortDirection: string;
 }
+
+export interface RenderWithCarbonArguments {
+  env: Galaxy;
+  rootElement: HTMLElement;
+  model: EngineAPI.IGenericObject | undefined;
```
haxxmaxx, line 118: Could this be optional instead of `| undefined`?

LiKang6688: Yes

src/types.ts
LiKang6688: Yes, https://qlik.dev/libraries-and-tools/nebulajs/in-qlik-sense#exposed-from-qlik-sense and https://github.com/qlik-oss/sn-pivot-table/blob/main/src/index.ts#L15

src/types.ts
LiKang6688: Yes, the translator is from Qlik Sense https://qlik.dev/libraries-and-tools/nebulajs/in-qlik-sense

src/index.ts
```diff
@@ -62,12 +73,14 @@ export default function supernova(env) {
...
-      }, [layout, pageInfo]);
+      const [tableData] = usePromise(
+        async () => (env.carbon ? nothing() : manageData(model, layout, pageInfo, setPageInfo)),
+        [layout, pageInfo]
+      );
 
       useEffect(() => {
-        if (!env.carbon && reactRoot && layout && tableData && announce && changeSortOrder && theme) {
+        const isReadyToRender = !env.carbon && reactRoot && layout && tableData && announce && changeSortOrder && theme;
+        isReadyToRender &&
```
haxxmaxx, line 83: was this to fix the problem with too many things in the if condition?

src/types.ts
haxxmaxx: great

src/index.ts
LiKang6688: yes

LiKang6688: codeclimate complains

src/types.ts
```diff
@@ -61,7 +71,7 @@ export interface PaginationColors {
 }
 export interface TableThemeColors {
   tableBackgroundColorFromTheme: string;
-  backgroundColor: string | undefined;
+  backgroundColor?: string;
```
haxxmaxx, line 74: Are you sure this is the same thing? This is set to undefined sometimes, not unset. It might be the same thing but I am not sure

haxxmaxx (approved):

src/types.ts
LiKang6688: undefined happens when we don’t assign a value to a variable, so unset === set to undefined

haxxmaxx: ok

PR #637, opened 2022-08-13
refactor: convert root to typescript

LiKang6688:


src/types.ts
```diff
@@ -101,3 +101,45 @@ export interface Column {
...
 }
+
+export type TableData =
+  | {
+      totalColumnCount: number;
+      totalRowCount: number;
+      totalPages: number;
+      paginationNeeded: boolean;
+      rows: number[];
+      columns: number[];
```
haxxmaxx, line 112: Pretty sure they are not arrays with numbers. There is already a type for columns, might have to define the rows one

src/types.ts
```diff
@@ -101,3 +101,45 @@ export interface Column {
...
+      totalsPosition: string;
+    }
+  | null
```
haxxmaxx, line 115: when is this null?

src/typings.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module 'styled-components' {
+  export type StylisPlugin = (context: number, content: string) => string | void;
+
+  export interface StyleSheetManagerProps {
+    children?: React.ReactNode;
+    disableCSSOMInjection?: boolean | undefined;
+    disableVendorPrefixes?: boolean | undefined;
+    stylisPlugins?: StylisPlugin[] | undefined;
+    sheet?: ServerStyleSheet | undefined;
+    target?: HTMLElement | ShadowRoot | undefined;
```
haxxmaxx, line 10: Do they need undefined here if they are optional?

LiKang6688: I do not think that they need it. I will remove them.

src/types.ts
LiKang6688: I realized that the type TableData is not needed in this PR. Gonna remove it.

src/types.ts
```diff
@@ -101,3 +101,32 @@ export interface Column {
...
+  rowsPerPageOptions: number[];
+}>;
+
+export type Announce = ((arg0: AnnounceArgs) => void) | undefined;
+
+export interface RenderProps {
+  direction: 'ltr' | 'rtl';
+  selectionsAPI: stardust.ObjectSelections;
+  rootElement: HTMLElement;
+  tableData: object;
```
haxxmaxx, line 123: if you are not doing this in this PR we need to do it at a later point

src/types.ts
```diff
@@ -101,3 +101,32 @@ export interface Column {
...
+  rowsPerPageOptions: number[];
+}
+
+export type SetPageInfo = stardust.SetStateFn<{
+  page: number;
+  rowsPerPage: number;
+  rowsPerPageOptions: number[];
+}>;
+
+export type Announce = ((arg0: AnnounceArgs) => void) | undefined;
```
haxxmaxx, line 117: this should already exist

src/types.ts
```diff
@@ -101,3 +101,32 @@ export interface Column {
...
+  rootElement: HTMLElement;
+  tableData: object;
+  pageInfo: PageInfo;
+  setPageInfo: SetPageInfo;
+  constraints: stardust.Constraints;
+  translator: ExtendedTranslator;
+  theme: ExtendedTheme;
+  keyboard: stardust.Keyboard;
+  footerContainer: HTMLElement;
+  announce: (...args: any[]) => any;
```
haxxmaxx, line 131: should be more strict, should be a type for it already

src/typings.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module 'styled-components' {
```
haxxmaxx, line 1: why is this in a seperate file?

LiKang6688: I tried with @types/styled-components which does not work for me. So I created it by myself.

src/types.ts
LiKang6688: I tried with `announce: Announce` which does not work. But I gonna look into it again.

src/types.ts
LiKang6688: I tried with `announce: Announce` which does not work. But I gonna look into it again.

src/types.ts
LiKang6688: Will get it done when working on handle-data.js

src/types.ts
LiKang6688: Will get it done when working on TableWrapper.jsx

LiKang6688: Done, may need to do some change working on TableWrapper.jsx

src/types.ts
LiKang6688: Done, may need to do some change working on handle-data.js

src/types.ts

src/typings.d.ts
LiKang6688: It has to be in a .d.ts file, as far as know

src/types.ts
```diff
@@ -101,3 +101,40 @@ export interface Column {
...
+  rowsPerPageOptions: number[];
+}>;
+
+export type TableData = {
+  totalColumnCount: number;
+  totalRowCount: number;
+  totalPages: number;
+  paginationNeeded: boolean;
+  rows: Record<string, number>;
+  columns: number[];
```
haxxmaxx, line 123: This is wrong. columns should be `Column[]`. If you don't want to do rows properly now set it to `Record<srting, any>` for now. Better to be loosely defined than incorrect

src/typings.d.ts
haxxmaxx: I don't understand what is going on here. You are declaring the types for the styled-components package? What happens without this? Does typescript complain?

src/nebula-hooks/use-announce-and-translations.ts
```diff
@@ -47,7 +47,7 @@ export const announcementFactory = (
 };
 
 const useAnnounceAndTranslations = (rootElement: HTMLElement, translator: ExtendedTranslator) => {
-  const [announce, setAnnounce] = useState<undefined | ((arg0: AnnounceArgs) => void)>(undefined);
+  const [announce, setAnnounce] = useState<undefined | Announce>(undefined);
```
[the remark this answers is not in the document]
LiKang6688: Right, fixed

src/types.ts
LiKang6688: Right, looked at handle-data and fixed in https://github.com/qlik-oss/sn-table/pull/637/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR17

src/typings.d.ts
LiKang6688: yes, typescript complains.
So I started with installing @types/styled-components which does not work with rtlPluginSc from 'stylis-plugin-rtl-sc'. 
As rtlPluginSc only takes `(context: number, content: string)`.
So I solved it by declaring the types for the styled-components package by myself.

src/types.ts
```diff
@@ -121,5 +128,67 @@ export interface RenderWithCarbonArguments {
...
+  setPageInfo?: SetPageInfo;
+  constraints?: stardust.Constraints;
+  translator?: ExtendedTranslator;
+  theme: ExtendedTheme;
+  keyboard?: stardust.Keyboard;
+  footerContainer?: HTMLElement;
+  announce?: Announce;
+  model?: EngineAPI.IGenericObject;
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  manageData?: Function;
```
LiKang6688, line 174: Will be fixed working on handle-data.js

src/types.ts
```diff
@@ -212,7 +216,46 @@ export type TableData = {
...
+  pageInfo?: PageInfo;
+  setPageInfo?: SetPageInfo;
+  constraints?: stardust.Constraints;
+  translator?: ExtendedTranslator;
+  theme: ExtendedTheme;
+  keyboard?: stardust.Keyboard;
+  footerContainer?: HTMLElement;
+  announce?: Announce;
+  model?: EngineAPI.IGenericObject;
+  manageData?: typeof manageData;
```
haxxmaxx, line 241: manageData is a static function, why is it on the renderProps? It also creates a circular dependency importing manageData to this file.

LiKang6688: Good question. Because in https://github.com/qlik-oss/sn-table/blob/main/src/index.ts#L48, it also calls render. 

RIght, gonna fix that.

LiKang6688: Done

haxxmaxx:
> @haxxmaxx I am considering removing the types.ts file in the codeclimate. What do you think?

Not sure it is good idea. As I mentioned before, we should probably split them up so you have maybe types for the components and types that are used everywhere in seperate files. There is a good reason why a file shouldn't be too big

haxxmaxx:
you can ignore it for now though

src/index.ts
```diff
@@ -45,7 +45,7 @@ const renderWithCarbon = ({
   changeSortOrder,
 }: RenderWithCarbonArguments) => {
   if (env.carbon && changeSortOrder && theme) {
-    render(rootElement, { layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect });
+    render(rootElement, { layout, model, theme, selectionsAPI, changeSortOrder, app, rect });
```
haxxmaxx, line 48: I'm pretty sure they use the instance of manageData that is passed here. If you remove it here you would have to change so the native implementation imports the function from file instead

src/table/Root.tsx
```diff
@@ -1,31 +1,36 @@
...
 import { ThemeProvider } from '@mui/material/styles';
 import rtlPluginSc from 'stylis-plugin-rtl-sc';
+
 import TableWrapper from './components/TableWrapper';
 import { TableContextProvider } from './context';
 import muiSetup from './mui-setup';
+import { RenderProps, RootProps } from '../types';
 
-export function render(reactRoot, props) {
+export function render(reactRoot: ReactDom.Root | HTMLElement, props: RenderProps) {
```
haxxmaxx, line 12: So since we have to take the native implementation into account it could be HTMLElement as well, and then cast it to ReactDom.Root? Now the file overriding thing doesn't seem like a good idea anymore

src/index.ts
LiKang6688: ops, that's not my intention to remove managedata. Thanks for finding it out.

src/table/Root.tsx
```diff
@@ -1,31 +1,36 @@
...
   const { direction, selectionsAPI } = props;
   const muiTheme = muiSetup(direction);
+  const root = reactRoot as ReactDom.Root;
+  const rootProps = props as RootProps;
```
haxxmaxx, line 16: Is it the same thing here, that you have to take both implementations into account?

src/table/Root.tsx
LiKang6688: When the native implementation runs:
`render(rootElement, { layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect });`
reactRoot is HTMLElement

when our code runs:
```
 render(reactRoot, {
            rootElement,
            layout,
            tableData,
            direction,
            pageInfo,
            setPageInfo,
            constraints,
            translator,
            selectionsAPI,
            theme,
            changeSortOrder,
            keyboard,
            rect,
            footerContainer,
            announce,
          });
```
reactRoot is ReactDom.Root

No, it is really not.

src/table/Root.tsx
LiKang6688: Right

LiKang6688: In practice, HTMLElement never reaches reactRoot here. But in typescript, we need to take care it.

src/table/Root.tsx
haxxmaxx: Well I guess we can do it like this for now. Better to write seperate functions with different names (like `renderNative` since they take different arguments at this point. Maybe we can start with your conversion in this PR and then we need to discuss a smarter way to use root

haxxmaxx: Well I guess we can do it like this for now. Better to write seperate functions with different names (like `renderNative` since they take different arguments at this point. Maybe we can start with your conversion in this PR and then we need to discuss a smarter way to use root

LiKang6688: I tried your way before, it does not work for me.

LiKang6688: Actually, I tried many other ways and I realized that nebula implementation only do the current way.

haxxmaxx: Not sure what you mean, I don't see why it wouldn't be possible. We can run different render functions depending on what env.carbon is. But that is not for this PR

haxxmaxx (approved):
We will have to revisit this but fine for now

src/table/Root.tsx
LiKang6688: then when running `yarn build`, it complains.

PR #640, opened 2022-08-16
refactor: convert handle-key-press to typescript

haxxmaxx:
- convert handle-keypress and test file
- rename key handler functions to handleBody/Head/Total/WrapperKeyDown and handleBodyKeyUp for consistency and more explanatory to what they do
- rename `arrowKeysNavigation` to `getNextCellCoord`
- removed the `getRowAndColumnCount` it was only used in `getNextCellCoord`. Bit strange to to have the cell counts on the outer scope when it is not needed there,


src/table/utils/selections-utils.ts
```diff
@@ -129,8 +129,11 @@ export const getSelectedRows = (
   if (evt.shiftKey && key.includes('Arrow')) {
     selectedRows[qElemNumber] = rowIdx;
     // add the next or previous cell to selectedRows, based on which arrow is pressed
-    selectedRows[key === 'ArrowDown' ? cell.nextQElemNumber : cell.prevQElemNumber] =
-      key === 'ArrowDown' ? rowIdx + 1 : rowIdx - 1;
+    if (key === 'ArrowDown') {
+      selectedRows[cell.nextQElemNumber as number] = rowIdx + 1;
+    } else {
+      selectedRows[cell.prevQElemNumber as number] = rowIdx - 1;
+    }
```
haxxmaxx, line 136: This was cleaner when  you had to add the casting + one less check for key === 'ArrowUp'

src/types.ts
```diff
@@ -10,8 +10,8 @@ export interface TableCell {
   isSelectable: boolean;
   rawRowIdx: number;
   rawColIdx: number;
-  prevQElemNumber: number;
-  nextQElemNumber: number;
+  prevQElemNumber: number | undefined;
+  nextQElemNumber: number | undefined;
```
haxxmaxx, line 14: this is always set so should not be optional, but first and last row they can be undefined

src/table/utils/handle-key-press.ts
```diff
@@ -1,15 +1,30 @@
+import { stardust } from '@nebula.js/stardust';
+import React from 'react';
```
LiKang6688, line 2: We normally import react first

src/table/utils/handle-key-press.ts
```diff
@@ -1,15 +1,30 @@
...
+const isCtrlShift = (evt: React.KeyboardEvent) => evt.shiftKey && (evt.ctrlKey || evt.metaKey);
 
-export const preventDefaultBehavior = (evt) => {
+export const preventDefaultBehavior = (evt: React.KeyboardEvent) => {
   evt.stopPropagation();
   evt.preventDefault();
 };
 
-export const handleTableWrapperKeyDown = ({
+interface HandleWrapperProps {
```
LiKang6688, line 16: why not in the types file?

haxxmaxx: good question. I haven't worried too much where the types are at this point. My innitial thought was to put types only used in one file in that file and then put types used in multiple files in the types file.

src/table/utils/handle-key-press.ts
haxxmaxx: ah yes, this was added automatically. I'll move it

src/table/utils/handle-key-press.ts
LiKang6688: My thought is that it is good not to make the file larger because of defining types.

src/table/utils/handle-key-press.ts
```diff
@@ -156,11 +214,11 @@ export const bodyHandleKeyPress = ({
   paginationNeeded,
   totalsPosition,
   selectionsAPI,
-}) => {
+}: handleBodyProps) => {
```
LiKang6688, line 217: we have `bodyHandleKeyUp` in the handle-key-press file. So it is better to rename it to be `handleBodyKeyDownProps`.

LiKang6688:
It would be good to spread rename and typescript conversation into different PRs next time.

src/types.ts
LiKang6688: in theory, no difference. But sure.

src/table/utils/handle-key-press.ts
haxxmaxx: yeah, I see that too. This file even surpasses the codeclimate threshold

src/types.ts
haxxmaxx: hmm well maybe that is only for function arguments it makes a difference then. anyway just a side note

src/table/utils/selections-utils.ts
LiKang6688: we can remove {}
```
if (key === 'ArrowDown') selectedRows[cell.nextQElemNumber as number] = rowIdx + 1;
else selectedRows[cell.prevQElemNumber as number] = rowIdx - 1;
```

src/table/utils/handle-key-press.ts
```diff
@@ -76,26 +102,41 @@ export const arrowKeysNavigation = (evt, rowAndColumnCount, cellCoord, topAllowe
...
   return nextCell;
 };
 
-export const headHandleKeyPress = ({
+interface handleHeadProps {
+  evt: React.KeyboardEvent;
+  rootElement: HTMLElement;
+  cellCoord: [number, number];
+  column: Column;
+  changeSortOrder: (layout: TableLayout, column: Column) => Promise<void>;
```
LiKang6688, line 130: changeSortOrder used in both handleHeadProps and RenderWithCarbonArguments. We can extract it to be ChangeSortOrder.

src/table/utils/handle-key-press.ts
```diff
@@ -76,26 +102,41 @@ export const arrowKeysNavigation = (evt, rowAndColumnCount, cellCoord, topAllowe
...
-
-export const moveFocus = (evt, rootElement, cellCoord, setFocusedCellCoord, topAllowedRow) => {
+/**
+ * resets and adds new focus to a table cell based which key is pressed
+ */
+export const moveFocus = (
+  evt: React.KeyboardEvent,
+  rootElement: HTMLElement,
+  cellCoord: [number, number],
+  setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>,
```
LiKang6688, line 112: same as changeSortOrder. We can extract it to be SetFocusedCellCoord.

haxxmaxx: yeah there are a bunch of these. Since they are so long, it could be good to extract them

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -553,25 +566,25 @@ describe('handle-key-press', () => {
...
+      expect((evt.target as HTMLElement).setAttribute).toHaveBeenCalledTimes(1);
       expect(setFocusedCellCoord).toHaveBeenCalledTimes(1);
       expect(selectionDispatch).toHaveBeenCalledTimes(1);
       expect(handleAccessibility.announceSelectionState).not.toHaveBeenCalled();
       expect(handleScroll.handleNavigateTop).toHaveBeenCalledTimes(1);
     });
 
     it('when press shift + arrow up key on the second row cell, should prevent default behavior, remove current focus and set focus and attribute to the next cell, but not select values for dimension', () => {
-      cell.prevQElemNumber = undefined;
+      cell.prevQElemNumber;
```
LiKang6688, line 580: it does not look correct

src/table/utils/handle-key-press.ts
haxxmaxx: hmm I think it is better to do in a different PR though

LiKang6688: sure

src/table/utils/__tests__/handle-key-press.spec.ts
haxxmaxx: ah yes, suppose to add a delete before it

haxxmaxx: actually the row is not needed at all. There is no `prevQElemNumber` on the cell by default

haxxmaxx: removed

src/table/utils/handle-key-press.ts
haxxmaxx: done

src/table/utils/handle-key-press.ts
haxxmaxx: done

src/table/utils/handle-key-press.ts
haxxmaxx: moved. I think when all files are done we would need to split up the types file in a smart way

src/table/utils/handle-key-press.ts
haxxmaxx: done

src/table/utils/selections-utils.ts
haxxmaxx: hmm yeah more compact but I would say it hurts readability, either or for me

LiKang6688 (approved):
Looks good

PR #641, opened 2022-08-16
fix: block left/right movement in selection mode

haxxmaxx:
Broken by me from a refactor a while back. We break if topAllowedRow > 0, and it was undefined when moving left/right


LiKang6688 (approved):
Had a manual test and it fixes the issue.

PR #642, opened 2022-08-16
feat: enabled new styling panel for sn table

rajani-varpe:
A new styling panel has been designed to be created in react, replacing the old style editor. Therefore, the sn-table- properties has been updated to allow new react components to be used to construct the new styling panel in React. This PR is about that particular property definition.This Feature is not under Feature Flag.
Also added translations for labels i.e for Header, Content and Row Hover.

image


src/ext.js
```diff
@@ -104,120 +105,9 @@ const getStyleSettings = (env) => {
...
-                  component: 'color-picker',
-                  dualOutput: true,
-                },
-              },
-            },
-          },
-        },
+        env.flags.isEnabled('SENSECLIENT_PS_17478_STYLINGPANEL_STRAIGHT_TABLE_OSS')
+          ? createStylingPanelDefinition()
+          : createStylingEditorDefinition(),
```
haxxmaxx, line 110: `createStylingEditorDefinition` seems to be the old definition, right? The names doesn't really differ to me. they are both editors and panels in a sense.

haxxmaxx, line 110: Is the naming here that the component prop is `panel-section` for the new one and `style-editor` for the old?

src/styling-panel-property-definition.js
```diff
@@ -0,0 +1,264 @@
+const createStylingPanelDefinition = () => {
```
haxxmaxx, line 1: This is static at this point so does not need to be a function, same for the old definition

src/ext.js
```diff
@@ -104,120 +105,9 @@ const getStyleSettings = (env) => {
...
-                  translation: 'ThemeStyleEditor.style.hoverFontStyle',
-                  type: 'object',
-                  component: 'color-picker',
-                  dualOutput: true,
-                },
-              },
-            },
-          },
-        },
+        env.flags.isEnabled('SENSECLIENT_PS_17478_STYLINGPANEL_STRAIGHT_TABLE_OSS')
```
haxxmaxx, line 108: This is failing in the unit tests since isEnabled is false. Easiest would be to add optional chaining and the tests would run with `createStylingEditorDefinition`

LiKang6688:
@rajani-varpe To be clear, the straight table and the sn-table (supernova table) are different.

src/ext.js
rajani-varpe: For the new one it is `styling-panel`.

src/ext.js
rajani-varpe: Added optional chaining

src/styling-panel-property-definition.js
rajani-varpe: Removed functions from both old and new definition.

src/ext.js
```diff
@@ -104,120 +105,9 @@ const getStyleSettings = (env) => {
...
-                  translation: 'ThemeStyleEditor.style.hoverFontStyle',
-                  type: 'object',
-                  component: 'color-picker',
-                  dualOutput: true,
-                },
-              },
-            },
-          },
-        },
+        env?.flags?.isEnabled('SENSECLIENT_PS_17478_STYLINGPANEL_SN_TABLE_OSS')
```
niekvanstaveren, line 108: I don't think there is a need to have this behind a feature flag. The sn-table is not available in the client yet (or at least not the property panel for it).

src/styling-panel-property-definition.js
```diff
@@ -0,0 +1,256 @@
+const createStylingPanelDefinition = {
```
niekvanstaveren, line 1: This is not a function anymore, so should probably have a different name

src/styling-panel-property-definition.js
```diff
@@ -0,0 +1,256 @@
...
+              component: 'inline-wrapper',
+              items: {
+                headerFontSize: {
+                  component: 'integer',
+                  ref: 'header.fontSize',
+                  type: 'integer',
+                  width: 9,
+                  min: 5,
+                  max: 300,
+                  // placeholder: () => parseInt(styleService.getStyle('header', 'fontSize'), 10), // Not sure about placeholder, as qlik-oss do not have theme integrated in it
```
niekvanstaveren, line 30: Theme should be available through the environment. Have a look if you can access the styleService:
Theme.initializeService

src/styling-panel-property-definition.js
rajani-varpe: Made required changes

src/ext.js
niekvanstaveren: No when we add this to the client we will use the new panel directly, so we won't have to keep this behind a feature flag. We can directly clean up the old parts as well

src/styling-panel-property-definition.js
rajani-varpe: We do not get the initializeService() access from env.

src/ext.js
rajani-varpe: Removed the old styling panel def along with the feature flag

src/styling-panel-property-definition.js
haxxmaxx: IMO this can stay commented for now, and fixed in a future PR

haxxmaxx:
Tested quickly by building from this branch and importing it as an extension, had the flag on as well. Doesn't look right to me
Screenshot 2022-08-29 at 14 05 34

rajani-varpe:
> Tested quickly by building from this branch and importing it as an extension, had the flag on as well. Doesn't look right to me
> Screenshot 2022-08-29 at 14 05 34

Actually the "Integer" component is implemented in Analytics repo and we had already raised a PR for it there which is not yet merged due to which you are getting the error of "No component found as Integer".

haxxmaxx:
> Actually the "Integer" component is implemented in Analytics repo and we had already raised a PR for it there which is not yet merged due to which you are getting the error of "No component found as Integer".

Ok so should we wait for that to be merged before we merge this PR?

rajani-varpe:
> > Actually the "Integer" component is implemented in Analytics repo and we had already raised a PR for it there which is not yet merged due to which you are getting the error of "No component found as Integer".
> 
> Ok so should we wait for that to be merged before we merge this PR?

Yes, as this table is not under any feature flag I think we should wait for the Integer component PR to get merge else it will give "Integer component not found" error.

Will inform you once the Integer component PR gets merged.

rajani-varpe:
> > > Actually the "Integer" component is implemented in Analytics repo and we had already raised a PR for it there which is not yet merged due to which you are getting the error of "No component found as Integer".
> > 
> > 
> > Ok so should we wait for that to be merged before we merge this PR?
> 
> Yes, as this table is not under any feature flag I think we should wait for the Integer component PR to get merge else it will give "Integer component not found" error.
> 
> Will inform you once the Integer component PR gets merged.

Integer Component PR is merged, can you please check and confirm, so that I will squash and merge this PR as well.

src/ext.js
```diff
@@ -105,114 +105,139 @@ const getStyleSettings = (env) => {
       type: 'items',
       items: [
         {
-          component: 'style-editor',
+          component: 'styling-panel',
+          chartTitle: 'Object.StraightTable',
```
haxxmaxx, line 109: This might be changed in the future, since the customer-facing name for sn-table is not decided yet. But this is fine for now

haxxmaxx (approved):

AmeliaAndersson (changes requested):
Screenshot 2022-09-09 at 16 09 31

The default values/placeholders are incorrect. These should not be 0 and white but instead the actual font size that is being used for header and content and the colour used for the text. If qlik-oss doesnt use theme Id rather see these values be hardcoded than to be 0 and white. Maybe we can look in the sn-table code and see how they retrieve the default values.

AmeliaAndersson:
In the old one its just empty in the integer and the colorpickerbox is set to "none", could we have that instead? I think that looks better than 0 and white.

rajani-varpe:
> Screenshot 2022-09-09 at 16 09 31
> 
> The default values/placeholders are incorrect. These should not be 0 and white but instead the actual font size that is being used for header and content and the colour used for the text. If qlik-oss doesnt use theme Id rather see these values be hardcoded than to be 0 and white. Maybe we can look in the sn-table code and see how they retrieve the default values.

As mentioned in previous review comment in this PR, we do not have theme access in oss repo due to which we have commented the placeholder line as it is which is already confirmed by Niek.

image

src/ext.js
rajani-varpe: Yes sure

haxxmaxx:
But the old one didn't show the wrong value, and the color picker showed the "none" color. @AmeliaAndersson is the none color available in the new picker yet?
Screenshot 2022-09-12 at 08 23 07

AmeliaAndersson:
> But the old one didn't show the wrong value, and the color picker showed the "none" color. @AmeliaAndersson is the none color available in the new picker yet? Screenshot 2022-09-12 at 08 23 07

Yes, you can have none in the new colorpicker. Its the default and then you can chose "disablenone" if you dont want it.

rajani-varpe:
> disablenone

@AmeliaAndersson I am not able to use defaultNone in color picker in oss/sn-table definition, it does not shows any effect on color-picker look. And in the placeholder/value do hardcoded value work ?

AmeliaAndersson:
I tried it out and seems theres a bug in the hook for disableNone in the colorpicker component in analytics folder. I will look at it and create a Jira bug for it. We can merge this in the meanwhile and Ill fix it in the future when I fix the colorpicker.

AmeliaAndersson (approved):

PR #643, opened 2022-08-16
refactor: convert handle-data to typescript

LiKang6688:


src/__test__/generate-test-data.ts
```diff
@@ -20,8 +20,8 @@ export function generateLayout(
   nDims: number,
   nMeas: number,
   nRows: number,
-  qColumnOrder = [],
-  qGrandTotalRow = []
+  qColumnOrder = [] as number[],
+  qGrandTotalRow = [] as Record<string, string>[]
```
haxxmaxx, line 24: This should work right?
```suggestion
  qColumnOrder: number[] = [],
  qGrandTotalRow:  Record<string, string>[] = []
```

src/handle-data.ts
```diff
@@ -23,19 +35,27 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
  * @param {Number} numDims
```
haxxmaxx, line 35: you can remove the jsdocs now since the typing should add the same information. It is nice to keep a small description though

src/handle-data.ts
```diff
@@ -23,19 +35,27 @@ export function getColumnOrder({ qColumnOrder, qDimensionInfo, qMeasureInfo }) {
...
  * @returns dimensions and measures total cell values as strings
  */
-export function getTotalInfo(isDim, layout, colIndex, numDims, columnOrder) {
+export function getTotalInfo(
+  isDim: boolean,
+  layout: TableLayout,
+  colIndex: number,
+  numDims: number,
+  columnOrder: number[]
+) {
```
haxxmaxx, line 44: So the return type is now implied as `string | undefined` we probably want to enforce that it becomes a string

src/types.ts
```diff
@@ -88,12 +89,22 @@ export interface ExtendedTheme extends stardust.Theme {
...
+  rowsPerPage: number;
+  rowsPerPageOptions: number[];
 }
 
-export type TotalsPosition = 'top' | 'bottom' | 'noTotals';
+export type SetPageInfo = stardust.SetStateFn<{
+  page: number;
+  rowsPerPage: number;
+  rowsPerPageOptions: number[];
+}>;
```
haxxmaxx, line 102:
```suggestion
export type SetPageInfo = stardust.SetStateFn<PageInfo>;
```

src/types.ts
```diff
@@ -112,6 +123,37 @@ export interface PaletteColor {
...
+  qAttrExprInfo: ExtendedNxAttrExprInfo[];
+}
+
+export interface ExtendedNxMeasureInfo extends EngineAPI.INxMeasureInfo {
+  textAlign: TextAlign;
+  qAttrExprInfo: ExtendedNxAttrExprInfo[];
+}
+
+export interface HyperCube extends Omit<EngineAPI.IHyperCube, 'qDimensionInfo' | 'qMeasureInfo'> {
+  qColumnOrder: number[];
```
haxxmaxx, line 151: does this not exist on the EngineAPI.IHyperCube ?

src/types.ts
```diff
@@ -170,3 +212,15 @@ export interface RenderWithCarbonArguments {
...
+export type TotalsPosition = 'top' | 'bottom' | 'noTotals';
+
+export type TableData = {
+  totalColumnCount: number;
+  totalRowCount: number;
+  totalPages: number;
+  paginationNeeded: boolean;
+  rows: Row[];
+  columns: Column[];
+  totalsPosition: string;
```
haxxmaxx, line 225:
```suggestion
  totalsPosition: TotalsPosition;
```

src/types.ts
```diff
@@ -112,6 +123,37 @@ export interface PaletteColor {
...
+  align: string;
+}
+
+export interface ExtendedNxAttrExprInfo extends EngineAPI.INxAttrExprInfo {
+  id: string;
+  qMin: number;
+  qMax: number;
+  qContinuousAxes: boolean;
+  qIsCyclic: boolean;
+  qFallbackTitle: string;
```
haxxmaxx, line 137: I thought these were already on here? that the only thing you needed to extend with is the id prop? https://help.qlik.com/en-US/sense-developer/May2022/Subsystems/EngineJSONAPI/Content/models-nxattrexprinfo.htm

src/types.ts
```diff
@@ -112,6 +123,37 @@ export interface PaletteColor {
...
+}
+
+export interface ExtendedNxDimensionInfo extends Omit<EngineAPI.INxDimensionInfo, 'qAttrExprInfo'> {
```
haxxmaxx, line 140: I've never seen this Omit syntax. isn't qAttrExprInfo: ExtendedNxAttrExprInfo[]; overriding it anyway?

src/handle-data.ts
LiKang6688: sure

src/types.ts
LiKang6688: no, qAttrExprInfo already exists in EngineAPI.INxDimensionInfo. We can only extend (add) a new one.

src/types.ts
LiKang6688: No, qColumnOrder does not exist on the EngineAPI.IHyperCube

src/types.ts
LiKang6688: right, fixed

src/handle-data.ts
LiKang6688: I see that it is fine to have totalInfo in Column to be undefined

src/types.ts
haxxmaxx: ok so you have to omit then add it back? otherwise you have a two props with the same name?

src/handle-data.ts
haxxmaxx: hmm yeah I guess it is would work too. It needs to be added to the column type though, i think I saw it missing there

src/types.ts
haxxmaxx: hmm surprising

src/handle-data.ts
LiKang6688: yes, I added in my previous commit.

src/types.ts
LiKang6688: Yes, otherwise typescript complains

haxxmaxx (approved):

PR #646, opened 2022-08-17
refactor: convert HOCs to typescript

haxxmaxx:
- Converting the HOCs to typescript
- Creating one common props type for the column styling and selection one. The last one will always take a standard mui TableCell so that one has the props for that one. This means that we keep all props on the properrty object, so not removing props once they are "used". I tried with React.StrictMode and got no warnings so seems god to me
- Removed the test for withStyling. It basically just mocks an element and then renders that. So I don't think it even tests anything. Should not affect the quality of the test suite, just the coverage a little bit
- Changing the argument for `getCellSelectionState` to not get the whole context, just the selectionState


src/table/components/withSelections.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default function withSelections(CellComponent: (props: hocProps) => JSX.Element) {
+  const HOC = (props: hocProps) => {
+    const { cell, styling, announce } = props;
+    const cellSelectionState = useContextSelector(TableContext, (value) =>
+      getCellSelectionState(cell, value.selectionState as SelectionState)
+    );
+    const selectionDispatch = useContextSelector(
+      TableContext,
+      (value) => value.selectionDispatch as React.Dispatch<TSelectionActions>
+    );
```
haxxmaxx, line 16: This isn't super pretty with the casting. The other options is to sanity check them

src/table/components/withSelections.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useContextSelector, TableContext } from '../context';
+import { getSelectionStyle } from '../utils/styling-utils';
+import { getCellSelectionState, SelectionActions, TSelectionActions } from '../utils/selections-utils';
+import { hocProps, SelectionState } from '../../types';
+
+export default function withSelections(CellComponent: (props: hocProps) => JSX.Element) {
+  const HOC = (props: hocProps) => {
+    const { cell, styling, announce } = props;
+    const cellSelectionState = useContextSelector(TableContext, (value) =>
+      getCellSelectionState(cell, value.selectionState as SelectionState)
```
LiKang6688, line 11: value is possible to be {}, as TableContext initially is {}. {}.selectionState goes wrong. 
So I think value.?selectionState is better.

haxxmaxx: Well according to the typing it can be {}, but don't think it will ever be in practice. Once root runs the first time the values are initialized. So I think it is fine to cast here, but not sure. The other alternative is to sanity check everytime we use something from the context

LiKang6688: that's true. I debug it and find that it never be {}. Then I suggest 
`export const TableContext = createContext<AppContextInterface>({} as AppContextInterface);`

src/table/context/TableContext.tsx
```diff
@@ -3,9 +3,9 @@ import { stardust } from '@nebula.js/stardust';
 import PropTypes from 'prop-types';
 import { createSelectorProvider } from './createSelectorProvider';
 import { reducer } from '../utils/selections-utils';
-import { ExtendedSelectionAPI } from '../../types';
+import { ContextValue, ExtendedSelectionAPI } from '../../types';
 
-export const TableContext = createContext({});
+export const TableContext = createContext<ContextValue>({});
```
LiKang6688, line 8:
```suggestion
export const TableContext = createContext<ContextValue>({} as ContextValue);
```

src/table/components/withSelections.tsx
haxxmaxx: ah that actually works with the props on ContextValue not being optional!

src/table/components/withSelections.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { useContextSelector, TableContext } from '../context';
+import { getSelectionStyle } from '../utils/styling-utils';
+import { getCellSelectionState, SelectionActions, TSelectionActions } from '../utils/selections-utils';
+import { hocProps, SelectionState } from '../../types';
+
+export default function withSelections(CellComponent: (props: hocProps) => JSX.Element) {
+  const HOC = (props: hocProps) => {
+    const { cell, styling, announce } = props;
+    const cellSelectionState = useContextSelector(TableContext, (value) =>
```
LiKang6688, line 10: It is strange that `const cellSelectionState = useContextSelector(TableContext, (value) => getCellSelectionState(cell, value));` worked before.

LiKang6688: Never mind, now I see.

src/table/components/withSelections.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { useContextSelector, TableContext } from '../context';
+import { getSelectionStyle } from '../utils/styling-utils';
+import { getCellSelectionState, SelectionActions, TSelectionActions } from '../utils/selections-utils';
+import { hocProps, SelectionState } from '../../types';
```
LiKang6688, line 5: TSelectionActions and SelectionState can be removed

src/table/context/TableContext.tsx
```diff
@@ -4,6 +4,10 @@ import { createSelectorProvider } from './createSelectorProvider';
 import { reducer } from '../utils/selections-utils';
 import { ExtendedSelectionAPI, ContextValue, ContextProviderProps } from '../../types';
 
+// In order to not have typing issues when using properties on the context,
+// the initial value for the context is casted to ContextValue.
+// In practice it will always be populated since TableContextProvider
+// runs before anything using props on the context
```
LiKang6688, line 10: Nice one

src/types.ts
```diff
@@ -320,6 +321,12 @@ export interface ContextProviderProps {
   cellCoordMock?: [number, number];
   selectionDispatchMock?: jest.Mock<any, any>;
 }
+export interface hocProps extends TableCellProps {
```
LiKang6688, line 324: I prefer HOCProps

src/table/components/withStyling.tsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import { TableCellProps } from '@mui/material';
+import { hocProps } from '../../types';
 
-export default function withStyling(CellComponent) {
-  const HOC = (props) => {
+export default function withStyling(CellComponent: (props: TableCellProps) => JSX.Element) {
```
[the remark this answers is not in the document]
LiKang6688: oh, never mind, there is hocProps that extends TableCellProps

src/table/components/withSelections.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default function withSelections(CellComponent: (props: hocProps) => JSX.Element) {
```
LiKang6688, line 7: (props: hocProps) => JSX.Element are used twice. 
I think it is better to put it into types.
export type CellComponentProps = (props: hocProps) => JSX.Element

haxxmaxx: sure

src/table/components/withSelections.tsx
haxxmaxx: done

src/table/components/withStyling.tsx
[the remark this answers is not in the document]
haxxmaxx: ah nice, done

src/table/components/withSelections.tsx
haxxmaxx: done

haxxmaxx:
@LiKang6688 I fixed the get-cell-renderer test as well

src/table/utils/__tests__/get-cell-renderer.spec.ts
```diff
@@ -5,15 +5,15 @@ import * as withStyling from '../../components/withStyling';
...
 
     beforeEach(() => {
       selectionsEnabled = false;
       hasColumnStyling = false;
-      jest.spyOn(withSelections, 'default').mockImplementation(() => jest.fn());
-      jest.spyOn(withColumnStyling, 'default').mockImplementation(() => jest.fn());
-      jest.spyOn(withStyling, 'default').mockImplementation(() => jest.fn());
+      jest.spyOn(withSelections, 'default');
+      jest.spyOn(withColumnStyling, 'default');
+      jest.spyOn(withStyling, 'default');
```
LiKang6688, line 16: nice. the pattern can be used in many other places.

haxxmaxx: yea I thought I would try the other places too in a different pr

PR #647, opened 2022-08-17
refactor: add types for object-properties and move files into qae folder

LiKang6688:


spec-configs/props.conf.js
```diff
@@ -3,7 +3,7 @@ const path = require('path');
 const pkg = require(path.resolve(__dirname, '../package.json')); // eslint-disable-line
 
 module.exports = {
-  glob: ['./src/object-properties.js'],
+  glob: ['./src/qlik-associative-engine/object-properties.js'],
```
haxxmaxx, line 6: I think we should stick to qae as other repos do, easier to understand if you are use to workin in a different repo

LiKang6688: Sure

haxxmaxx (approved):

PR #648, opened 2022-08-17
docs: add yarn spec

LiKang6688:


.github/CONTRIBUTION.md
```diff
@@ -44,6 +44,14 @@ Run type checking with:
 
     yarn types:check
 
+## Spec checking
+
+Run spec checking with:
```
haxxmaxx, line 49: I mean what you are doing is not checking something, the command generates a JSON file for the API spec. And then  it tells you if something is off. If we want to add this to the documentation it should explain what it does, and that you can use it to make sure the props definition is correct

LiKang6688: That's better. I will try to write them down as you said.

.github/CONTRIBUTION.md
```diff
@@ -9,6 +9,16 @@
...
 
+## API spec generating and verification
+
+Run API spec generating and verification with:
+
+    yarn spec
+
+It verifies the object-properties.js file (a JSDoc file) and generates a [Scriptappy](https://github.com/qlik-oss/scriptappy) definition documentation (a API spec) from that file.
+
+> Note: When there is any change for the [API spec](../api-specifications/properties.json), DO NOT commit the file. The change will be merged into the main branch when doing the Releasing.
```
haxxmaxx, line 20:
```suggestion
> Note: When there is any change in the [API spec](../api-specifications/properties.json), DO NOT commit the file. The change will be updated and merged into the main branch when doing the next release
```

haxxmaxx (approved):

PR #651, opened 2022-08-18
feat: add dragging table columns

jjxie:
Add dragging table columns to change columns order in table under feature flag `DRAGGING_COLUMNS_IN_TABLE`. 

1. There are files and logic I copied from sense-client. They are not converted to ts files in the PR and I didn't add tests to them. I think we can convert and add tests later since this PR is getting bigger.
2. I am not sure why column's width is not stored correctly to columns data, it is always default 200 that why I have to update width in `TableHeadWrapper.jsx`. I saw there are some logic related to columns width. If no one knows why immediate here, I will try to debug more.
3. ~~Known issue. After dragging, columns focus style sometimes are not correct. Ideally should focus on dragged column~~. Removed focus style since it re-rendered the table, so make sense without any focus style. To be honest, I tried a lot but can't make it focus on the new column position. 

https://user-images.githubusercontent.com/25456307/186172104-7faf9f68-c214-49a8-9a40-14ef287d11bd.mov


Donya-qlik:
Almost all sn-table files are converted to Typescript now so after rebasing your branch I guess there will be no option than converting yours to ts.

package.json
```diff
@@ -101,6 +101,7 @@
     "eslint-plugin-react": "7.30.1",
     "eslint-plugin-react-hooks": "4.6.0",
     "eslint-plugin-testing-library": "5.6.0",
+    "extend": "^3.0.2",
```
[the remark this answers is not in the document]
jjxie: It means Compatible with version, i.e., accept new minor and patch versions. I think when I add the package, it added like that

jjxie:
> Almost all sn-table files are converted to Typescript now so after rebasing your branch I guess there will be no option than converting yours to ts.

I added my main files as ts files. I agree to convert them to ts files, but I don't think will be good to do it in this PR. They are copied from sense client, need some time to converting and adding tests. This PR is already getting bigger and have to keep merging typescript changes.

package.json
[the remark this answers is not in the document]
LiKang6688: in sn-table, we omit the caret symbol as we bump dependencies by ourselves or by renovate-bot.

src/global.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*';
```
LiKang6688, line 1: what is for?

src/table/components/TableBodyWrapper.tsx
```diff
@@ -38,7 +39,14 @@ function TableBodyWrapper({
...
   const bodyCellStyle = useMemo(() => getBodyCellStyle(layout, theme), [layout, theme]);
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
-  const cellStyle = { color: bodyCellStyle.color, backgroundColor: theme.table.backgroundColor };
+  const getCellStyle = (columnIndex: number) => {
+    return {
+      color: bodyCellStyle.color,
+      backgroundColor: theme.table.backgroundColor,
+      opacity: engagedColumn !== undefined && engagedColumn !== columnIndex ? '50%' : null,
+    };
+  };
```
LiKang6688, line 48:
```suggestion
const getCellStyle = (columnIndex: number) => ({
    color: bodyCellStyle.color,
    backgroundColor: theme.table.backgroundColor,
    opacity: engagedColumn !== undefined && engagedColumn !== columnIndex ? '50%' : null,
  });
```
Not big deal, just looks cleaner.

src/table/utils/handle-drag-columns.ts
```diff
@@ -0,0 +1,124 @@
...
+  rtl: boolean;
+  columns: Column[];
+  setEngagedColumn: (i: number) => void;
+};
+
+type DragEndProps = {
+  event: DragEvent;
+  model: EngineAPI.IGenericObject;
+  setEngagedColumn: (arg0: undefined) => void;
+};
```
LiKang6688, line 36: I suggest moving those types into the types.ts file

src/global.d.ts
jjxie: It used to fix error from typescript of using extend 
```
Could not find a declaration file for module 'extend'. '/Users/fcy/Documents/GitHub/sn-table/node_modules/extend/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/extend` if it exists or add a new declaration (.d.ts) file containing `declare module 'extend';`
```

src/table/utils/handle-drag-columns.ts
jjxie: Sure

src/table/components/TableBodyWrapper.tsx
jjxie: sure

src/global.d.ts
LiKang6688: https://www.npmjs.com/package/@types/extend 
`yarn add -D @types/extend` should work, right?

jjxie: Yes, adding @types/extend fixed the issue.

haxxmaxx:
Hi @jjxie. Our ts migration is done, so if you fix your merge conflicts there shouldn't be any new big ones. But we realized there is a different problem. Dragging and dropping columns is not available for users that can't use a mouse. That breaks the principle that every feature available using a mouse should be available using the keyboard. There are discussions on how to treat this coming up soon

jjxie:
> Hi @jjxie. Our ts migration is done, so if you fix your merge conflicts there shouldn't be any new big ones. But we realized there is a different problem. Dragging and dropping columns is not available for users that can't use a mouse. That breaks the principle that every feature available using a mouse should be available using the keyboard. There are discussions on how to treat this coming up soon

@haxxmaxx Thanks for letting me know. The moving columns feature is behind feature flag. I think we can review and merge this PR first. When the design or discussion is done, we can make changes to support keyboard users.

haxxmaxx:
> @haxxmaxx Thanks for letting me know. The moving columns feature is behind feature flag. I think we can review and merge this PR first. When the design or discussion is done, we can make changes to support keyboard users.

Well, that is the thing, We don't know what a keyboard solution for this would be. And I don't think it is a good idea to implement a feature and the retrofit the a11y for it. We have worked like that before and almost always regretted it. Hopefully there will be some discussion around this during the week. Multiple upcoming features have the same issue

jjxie:
> > @haxxmaxx Thanks for letting me know. The moving columns feature is behind feature flag. I think we can review and merge this PR first. When the design or discussion is done, we can make changes to support keyboard users.
> 
> Well, that is the thing, We don't know what a keyboard solution for this would be. And I don't think it is a good idea to implement a feature and the retrofit the a11y for it. We have worked like that before and almost always regretted it. Hopefully there will be some discussion around this during the week. Multiple upcoming features have the same issue

Ok, I will resolve the conflicts and leave the PR here. When the discussion is done, we can decide to use it or rewrite it.

haxxmaxx:
> Ok, I will resolve the conflicts and leave the PR here. When the discussion is done, we can decide to use it or rewrite it.
Sounds good, thanks for the understanding

haxxmaxx:
Closing this since the rearrange columns feature for sn-table can be achieved by the table explorations

PR #653, opened 2022-08-18
refactor: migrate custom hook pattern

LiKang6688:
Migrate from
```
const [filteredData, setFilteredData] = useState();
useEffect(() => {
  setFilteredData(data.filter(x => isOfSomeType(x)));
}, [data]);
return <List data={filteredData} />;
```

to it
```
const filteredData = useMemo(() => {
  return data.filter(x => isOfSomeType(x));
}, [data]);
return <List data={filteredData} />;
```
Reference: https://jnystad.no/post/not-using-usecallback-is-premature-optimization

### DIFF
--- a/src/nebula-hooks/use-extended-theme.ts
+++ b/src/nebula-hooks/use-extended-theme.ts
@@ -44,12 +44,10 @@ export const tableThemeColors = (theme: ExtendedTheme, rootElement: HTMLElement)
 const useExtendedTheme = (rootElement: HTMLElement): ExtendedTheme => {
   // TODO: add name method to stardust.Theme
   const nebulaTheme = useTheme() as ExtendedTheme;
-  const theme = useMemo(
+  return useMemo(
     () => ({ ...nebulaTheme, table: tableThemeColors(nebulaTheme, rootElement) } as ExtendedTheme),
     [nebulaTheme.name(), rootElement]
   );
-
-  return theme;
 };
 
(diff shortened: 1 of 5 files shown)


src/nebula-hooks/use-extended-theme.ts
```diff
@@ -44,12 +44,10 @@ export const tableThemeColors = (theme: ExtendedTheme, rootElement: HTMLElement)
 const useExtendedTheme = (rootElement: HTMLElement): ExtendedTheme => {
   // TODO: add name method to stardust.Theme
   const nebulaTheme = useTheme() as ExtendedTheme;
-  const theme = useMemo(
+  return useMemo(
```
haxxmaxx, line 47: this works the same? never seen a useMemo been directly returned

LiKang6688: yes, typescript can help to verify that the type is the same. I will look into how to write unit tests for the custom hook.

haxxmaxx (approved):

src/nebula-hooks/use-extended-theme.ts
haxxmaxx: yeah I mean if there is a simple way to do it. We have a really high test coverage so more tests does not necessarily add more value

LiKang6688: Okay, it is not easy. then I will not do it.

PR #655, opened 2022-08-19
refactor: rename TableCell type to Cell

haxxmaxx:
This was conflicting with mui TableCell and more aligned with the naming of `Column` and `Rox`.  Also fixed the imports for styles file

### DIFF
--- a/src/table/styles.ts
+++ b/src/table/styles.ts
@@ -1,14 +1,12 @@
-import {
-  styled,
-  Paper,
-  TableContainer,
-  TableRow,
-  TableSortLabel,
-  TableBody,
-  Select,
-  IconButton,
-  TableCell,
-} from '@mui/material';
+import styled from '@mui/system/styled';
+import Paper from '@mui/material/Paper';
+import TableContainer from '@mui/material/TableContainer';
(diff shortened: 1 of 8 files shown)


src/table/styles.ts
```diff
@@ -1,14 +1,12 @@
...
-} from '@mui/material';
+import styled from '@mui/system/styled';
+import Paper from '@mui/material/Paper';
+import TableContainer from '@mui/material/TableContainer';
+import TableRow from '@mui/material/TableRow';
+import TableSortLabel from '@mui/material/TableSortLabel';
+import TableBody from '@mui/material/TableBody';
+import Select from '@mui/material/Select';
+import IconButton from '@mui/material/IconButton';
+import TableCell from '@mui/material/TableCell';
```
LiKang6688, line 9: Does it reduce the bundle size?

LiKang6688 (approved):

src/table/styles.ts
haxxmaxx: I think so. We have been importing the components like this in the other files, so might as well stick to it

LiKang6688: good

PR #657, opened 2022-08-19
refactor: convert TableBodyWrapper to typescript

Donya-qlik:


src/table/components/TableBodyWrapper.tsx
```diff
@@ -20,7 +20,7 @@ function TableBodyWrapper({
   tableWrapperRef,
   announce,
   children,
-}) {
+}: RootProps) {
```
LiKang6688, line 23: Create a BodyWrapperProps and extends RootProps

src/types.ts
```diff
@@ -312,6 +312,9 @@ export interface RootProps {
   keyboard: stardust.Keyboard;
   footerContainer?: HTMLElement;
   announce: Announce;
+  setShouldRefocus(): void;
+  tableWrapperRef: React.MutableRefObject<HTMLDivElement | undefined>;
+  children: JSX.Element;
```
LiKang6688, line 317:
```suggestion
export interface BodyWrapperProps extends Omit<RootProps, 'selectionsAPI'> {
  selectionsAPI: ExtendedSelectionAPI;
  setShouldRefocus(): void;
  tableWrapperRef: React.MutableRefObject<HTMLDivElement | undefined>;
  children: JSX.Element;
}
```

src/types.ts
```diff
@@ -298,7 +298,7 @@ export interface RenderProps {
 
 export interface RootProps {
   direction?: 'ltr' | 'rtl';
-  selectionsAPI: stardust.ObjectSelections;
+  selectionsAPI: ExtendedSelectionAPI;
```
LiKang6688, line 301: I am not sure when the stardust.ObjectSelections becomes the ExtendedSelectionAPI. Looking into it.

LiKang6688:
I got the type of selectionsAPI wrong in the index.ts, should be:
`const selectionsAPI = useSelections() as ExtendedSelectionAPI;`

src/table/components/TableBodyWrapper.tsx
```diff
@@ -31,15 +31,22 @@ function TableBodyWrapper({
...
     [columnsStylingIDsJSON, isSelectionsEnabled]
   );
   const bodyCellStyle = useMemo(() => getBodyCellStyle(layout, theme), [layout, theme]);
   const hoverEffect = layout.components?.[0]?.content?.hoverEffect;
   const cellStyle = { color: bodyCellStyle.color, backgroundColor: theme.table.backgroundColor };
-
   useEffect(() => {
-    addSelectionListeners({ api: selectionsAPI, selectionDispatch, setShouldRefocus, keyboard, tableWrapperRef });
+    addSelectionListeners({
+      api: selectionsAPI as ExtendedSelectionAPI,
```
Donya-qlik, line 44: @LiKang6688  set `api `as `selectionAPI ` is the reason of using ExtendedSelectionAPI here.

src/types.ts
Donya-qlik: I explained it at above on the right file.

Donya-qlik:
> I got the type of selectionsAPI wrong in the index.ts, should be: `const selectionsAPI = useSelections() as ExtendedSelectionAPI;` all `stardust.ObjectSelections` should be replaced with `ExtendedSelectionAPI`

@LiKang6688 index.d.ts is the file that should be modified by the new type and as you know is not editable via the current PR, could you give me a hint how I could I fix those in nebula? there are couple of more things that should be change there that already has ts-ignore tag on.

LiKang6688:
> > I got the type of selectionsAPI wrong in the index.ts, should be: `const selectionsAPI = useSelections() as ExtendedSelectionAPI;` all `stardust.ObjectSelections` should be replaced with `ExtendedSelectionAPI`
> 
> @LiKang6688 index.d.ts is the file that should be modified by the new type and as you know is not editable via the current PR, could you give me a hint how I could I fix those in nebula? there are couple of more things that should be change there that already has ts-ignore tag on.

@Donya-qlik You do not need to fix those in the nebula. 
We could do the fix in the src/index.ts(not index.d.ts), 
from
`const selectionsAPI = useSelections()`
to
`const selectionsAPI = useSelections() as ExtendedSelectionAPI;`

Donya-qlik:
> as ExtendedSelectionAPI

Ah, true! Will fix it

LiKang6688:
stardust.ObjectSelections should also be migrated to ExtendedSelectionAPI everywhere, for example, https://github.com/qlik-oss/sn-table/blob/main/src/types.ts#L276

src/types.ts
```diff
@@ -315,9 +315,16 @@ export interface RootProps {
   announce: Announce;
 }
 
+export interface BodyWrapperProps
+  extends Omit<RootProps, 'changeSortOrder' | 'rect' | 'pageInfo' | 'setPageInfo' | 'footerContainer' | 'translator'> {
```
haxxmaxx, line 319: Not sure about this.  Almost half the props on RootProps are being omitted and three are added. As I mentioned in here https://github.com/qlik-oss/sn-table/pull/663 I'm not sure we should omit these as they are actually passed.

src/table/components/__tests__/TableBodyWrapper.spec.tsx
```diff
@@ -38,49 +55,58 @@ describe('<TableBodyWrapper />', () => {
...
+    tableFirstRow = tableData.rows[0]['col-0'] as Cell;
   });
 
   afterEach(() => jest.clearAllMocks());
 
   it('should render 2x2 table body and call CellRenderer', () => {
     jest.spyOn(getCellRenderer, 'default').mockImplementation(() => {
       cellRendererSpy();
-      return 'td';
+      return 'td' as unknown as CellHOC;
```
haxxmaxx, line 92: I think this is way more complicated than needed. Try just `jest.spyOn(getCellRenderer, 'default')` and then `    expect(getCellRenderer.default).toHaveBeenCalledTimes(2);`

haxxmaxx: Goes for the other tests below as well

haxxmaxx:
Looks good but seems like you need to fix those rendering tests

src/types.ts
Donya-qlik: First I added them to RootProps but Li suggested the extended ones. The only benefit of omitting the unnecessary ones is we don't need to add every single RootProps in TableBodyWraper test component.

Donya-qlik:
> Looks good but seems like you need to fix those rendering tests

@haxxmaxx rendering tests? There is no test failure as far as I see ...

src/table/components/__tests__/TableBodyWrapper.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+    jest.spyOn(handleKeyPress, 'handleBodyKeyDown');
+
+    const { getByText } = renderTableBody();
+    fireEvent.keyDown(getByText(tableFirstRow.qText as string));
+
+    expect(handleKeyPress.handleBodyKeyDown).toHaveBeenCalledTimes(1);
+  });
+
+  it('should call handleClickToFocusBody on mouseDown', () => {
+    jest.spyOn(handleAccessibility, 'handleClickToFocusBody').mockImplementation(() => jest.fn());
```
haxxmaxx, line 108:
```suggestion
    jest.spyOn(handleAccessibility, 'handleClickToFocusBody');
```

src/table/components/__tests__/TableBodyWrapper.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+          tableWrapperRef={tableWrapperRef}
+          announce={announce}
+        >
+          {children}
+        </TableBodyWrapper>
+      </TableContextProvider>
+    );
+
+  beforeEach(async () => {
+    jest.spyOn(selectionsUtils, 'addSelectionListeners').mockImplementation(() => jest.fn());
```
haxxmaxx, line 63:
```suggestion
    jest.spyOn(selectionsUtils, 'addSelectionListeners');
```

src/types.ts
haxxmaxx: ok go for this now and we might refactor it later

haxxmaxx:
> > Looks good but seems like you need to fix those rendering tests
> 
> @haxxmaxx rendering tests? There is no test failure as far as I see ...

I must have seen a ghost. Tests looks fine!

src/table/components/__tests__/TableBodyWrapper.spec.tsx
Donya-qlik: There is a bunch of `api.on( ... )` lines in `addSelectionListeners` that by this change,  are not functioning, seems this line is needed.

haxxmaxx: that makes sense actually. Try with
```suggestion
    jest.spyOn(selectionsUtils, 'addSelectionListeners').mockImplementation();
```

haxxmaxx (approved):

PR #660, opened 2022-08-21
fix: object properties delcaration file

LiKang6688:


src/qae/object-properties.js
```diff
@@ -16,20 +16,14 @@ const properties = {
...
-     * @type {MeasureProperties[]}
+     * @type {MeasureProperty[]}
      */
     qMeasures: [],
-    /** @type {schemasNxHypercubeMode} */
-    qMode: 'S',
-    /** @type {boolean} */
-    qSuppressZero: false,
-    /** @type {boolean} */
-    qSuppressMissing: true,
```
LiKang6688, line 32: They are already in the EngineAPI.IHyperCubeDef

src/qae/object-properties.js
```diff
@@ -78,25 +72,25 @@ const properties = {
   },
   /**
    * Holds general styling
-   * @type {?Styling[]}
```
LiKang6688, line 81: I think Component is better, so I choose to match Component with    the declaration file

src/qae/object-properties.js
```diff
@@ -16,20 +16,14 @@ const properties = {
   qHyperCubeDef: {
     /**
      * The maximum amount of dimensions is 1000
-     * @type {DimensionProperties[]}
+     * @type {DimensionProperty[]}
```
haxxmaxx, line 19: For one dimension, it still is a collection of properties. I think this should be reverted

LiKang6688: Sure

LiKang6688: Done

src/types.ts
```diff
@@ -156,6 +181,18 @@ export interface StylingLayout {
...
+  fontSize?: number;
+}
+
+interface ContentStyling extends HeaderStyling {
+  hoverColor?: PaletteColor;
+  hoverEffect?: boolean;
+  hoverFontColor?: PaletteColor;
+  padding?: string; // not available as option in the styling panel
+}
+
```
haxxmaxx, line 195: Is this used anywhere yet?

src/types.ts
```diff
@@ -178,10 +215,10 @@ export interface CellStyle {
   selectedCellClass?: string;
 }
 
-export interface Components {
+export interface Component {
   key: string;
-  content?: StylingLayout;
-  header?: StylingLayout;
+  content?: ContentStyling;
+  header?: HeaderStyling;
```
haxxmaxx, line 221: These are not the types the `styleObj` has in `getBaseStyling`. I intentionally made the the same type so that you wouldn't need a union to type `getBaseStyling`

src/types.ts
```diff
@@ -147,6 +148,30 @@ export interface HyperCube extends Omit<EngineAPI.IHyperCube, 'qDimensionInfo' |
   qMeasureInfo: ExtendedNxMeasureInfo[];
 }
 
+interface InlineDimensionDef extends EngineAPI.INxInlineDimensionDef {
+  textAlign: TextAlign;
+}
+interface InlineMeasureDef extends EngineAPI.INxInlineMeasureDef {
+  textAlign: TextAlign;
+}
+interface AttributeExpressionProperty extends EngineAPI.INxAttrExprDef {
```
haxxmaxx, line 157: same here, it is a collection of properties

src/types.ts
haxxmaxx: but `StylingLayout` is not the same as `ContentStyling` and `HeaderStyling`. Does that work even though they do not  have anything to do with eachother?

LiKang6688: Good question. I see your point.
Screenshot 2022-08-22 at 11 26 57
 It works. As of now all properties in ContentStyling, HeaderStyling, and StylingLayout are optional. We probably want to change it.

LiKang6688: We need to think about whether to use optional or undefined in types.ts. 
For this specific case, I think they can be optional, as it is okay not to provide any value.

haxxmaxx: Yes they are actually optional. Hmm I think either we do one generic as before just `StylingLayout` or we use yours and then set the type for getBaseStyle to `styleObj: ContentStyling | HeaderStyling`. I made it just one type since they overlap a lot but everything is optional so you could use one type for both

LiKang6688: I would like to match Component in src/qae/object-properties.js with Component in src/types.ts. So I choose to `styleObj: ContentStyling | HeaderStyling`

haxxmaxx: sounds good

LiKang6688: Dona,  please have a look

LiKang6688:
Funny things happened, will look into it.

src/table/utils/styling-utils.ts
```diff
@@ -81,7 +89,7 @@ export const getBaseStyling = (
     fontFamily,
     color: isColorSet(styleObj?.fontColor) ? getColor(STYLING_DEFAULTS.FONT_COLOR, theme, styleObj?.fontColor) : color,
     fontSize: (styleObj?.fontSize && `${styleObj.fontSize}px`) || fontSize,
-    padding: getPadding(styleObj),
+    padding: styleObj && 'padding' in styleObj ? getPadding(styleObj) : undefined,
```
haxxmaxx, line 92: once we add padding to the API, I think it should be a string. Therefor we could use optional chainging here
```suggestion
    padding: styleObj?.padding ? getPadding(styleObj) : undefined,
```

LiKang6688: Unfortuanlly typescript complains `styleObj?.padding`

haxxmaxx: hmm ok I have no better suggestion then

LiKang6688: I have a question. When we set a fontSize for the header, the header should get padding based on that font size, right?

haxxmaxx: ah yes, we need to run it even if there is no padding. Now you are basically doing the logic that is in the getPadding function before you call it. I would say better to rewrite getPadding to do better sanity checking

LiKang6688: sounds good. Doing it

LiKang6688: Done

haxxmaxx (approved):

PR #661, opened 2022-08-21
refactor: remove default padding in styling utils

LiKang6688:


src/table/utils/styling-utils.ts
```diff
@@ -15,7 +15,6 @@ export const STYLING_DEFAULTS = {
   EXCLUDED_BACKGROUND:
     'repeating-linear-gradient(-45deg, rgba(200,200,200,0.08), rgba(200,200,200,0.08) 2px, rgba(200,200,200,0.3) 2.5px, rgba(200,200,200,0.08) 3px, rgba(200,200,200,0.08) 5px)',
   WHITE: '#fff',
-  PADDING: '7px 14px',
```
LiKang6688, line 18: As we already had the default padding in the muiTheme https://github.com/qlik-oss/sn-table/blob/main/src/table/mui-setup.js#L13

src/table/utils/styling-utils.ts
```diff
@@ -34,15 +33,15 @@ export const SELECTION_STYLING = {
 /**
  * Determines if a palette color is set or not. Both index !== -1 and color !== null must be true for it to be unset
  */
-export const isColorSet = (prop: PaletteColor | undefined): boolean =>
+export const isColorSet = (prop?: PaletteColor): boolean =>
```
haxxmaxx, line 36: Not a fan of this. This makes `isColorSet()` a valid call, which it is not. There is no point of calling this function if you are not passing an argument

src/table/utils/styling-utils.ts
```diff
@@ -34,15 +33,15 @@ export const SELECTION_STYLING = {
...
   !!prop && JSON.stringify(prop) !== JSON.stringify({ index: -1, color: null });
 
 /**
  * Gets specified padding from layout if defined, otherwise calculates it based on specified font size if defined, otherwise returns default padding.
  * Note that the padding property can't be set in the styling panel
  */
-export function getPadding(styleObj: StylingLayout | undefined, defaultPadding: string): string {
-  let padding = defaultPadding;
+export function getPadding(styleObj?: StylingLayout): string | undefined {
```
haxxmaxx, line 43: Same here, the styleObj argument is not optional.

src/table/utils/styling-utils.ts
```diff
@@ -69,11 +68,7 @@ export const getAutoFontColor = (backgroundColor: string): string =>
 /**
  * Gets base styling for either header or body taking table theme settings into account
  */
-export const getBaseStyling = (
-  styleObj: StylingLayout | undefined,
-  objetName: string,
-  theme: ExtendedTheme
-): GeneratedStyling => {
+export const getBaseStyling = (objetName: string, theme: ExtendedTheme, styleObj?: StylingLayout): GeneratedStyling => {
```
haxxmaxx, line 71: Same here

src/table/utils/styling-utils.ts
LiKang6688: okay, gonna revert it.

LiKang6688: Done.
But when `export const isColorSet = (prop: PaletteColor | undefined): boolean =>`,  isColorSet() is also a valid call.

haxxmaxx (approved):

PR #663, opened 2022-08-22
refactor: convert PaginationContent to ts

haxxmaxx:


src/table/components/__tests__/PaginationContent.spec.tsx
```diff
@@ -36,13 +52,17 @@ describe('<PaginationContent />', () => {
         rect={rect}
         handleChangePage={handleChangePage}
         announce={announce}
+        selectionsAPI={SelectionsAPI}
+        rootElement={rootElement}
+        layout={layout}
+        changeSortOrder={changeSortOrder}
```
haxxmaxx, line 58: These are needed since they are on RootProps, that we extend. Could be omitted but not sure it is worth it. In production these props are actually passed since we are spreading props from tableWrapper

Donya-qlik: @haxxmaxx What if you use Pick<T, K> from RootProps instead ?

haxxmaxx: yeah I don't know, none of the solutions are good IMO. I'm thinking we might have a type called `commonProps` or something. That one has the props that all components have, then you extend it with props needed for each component. The common props is completely arbitrary, it is just the overlap of all components. For now that is the best solution I can think of.

src/table/components/PaginationContent.tsx
```diff
@@ -143,7 +155,7 @@ function PaginationContent({
 
   const pageOptions = (
     <>
-      {[...Array(totalPages).keys()].map((pageIdx, index) => (
+      {Array.from(Array(totalPages).keys()).map((pageIdx, index) => (
```
haxxmaxx, line 158: ts was giving me hell here. The error message told me do add a flag to the tsconfig. I did and the error was still there. This is basically the same to me so I didn't bother digging deeper

src/table/components/PaginationContent.tsx
```diff
@@ -70,19 +70,26 @@ function PaginationContent({
   const showRowsPerPage = !isSelectionMode && shouldShow('rppOptions', width) && totalColumnCount <= 100;
   const displayedRowsText = translator.get('SNTable.Pagination.DisplayedRowsLabel', [
     `${page * rowsPerPage + 1} - ${Math.min((page + 1) * rowsPerPage, totalRowCount)}`,
-    totalRowCount,
+    totalRowCount.toString(),
```
Donya-qlik, line 73: Isn't is a number ?

haxxmaxx: yes

Donya-qlik: .. so why have you converted it to string ?

haxxmaxx: because translator wants strings

Donya-qlik: Ok, I see.

Donya-qlik:
...except the left type error you mentioned, that I'm not good at , the rest lgtm. : )

Donya-qlik (approved):

src/table/components/PaginationContent.tsx
Donya-qlik: Mario might be able to help, although not sure if he has access to this repo,

LiKang6688: I do not see any error or warning here. Sometimes rebooting VScode helps.

haxxmaxx: with the old implementation I mean, with Array.from() there is no error

PR #664, opened 2022-08-22
refactor: simplify test files + root arguments

haxxmaxx:
- use `getBy...` instead of `queryBy...` when we assume that the element we are getting exists. Removes the need for cast
- refactor root arguments etc. Makes more sense to me to have the reactRoot as an optional argument, instead of one argument that could have completely different types
- convert AnnounceElements.spec to tsx. just a rename needed


src/table/Root.tsx
```diff
@@ -7,19 +7,17 @@ import rtlPluginSc from 'stylis-plugin-rtl-sc';
 import TableWrapper from './components/TableWrapper';
 import { TableContextProvider } from './context';
 import muiSetup from './mui-setup';
-import { RenderProps, RootProps } from '../types';
+import { RenderProps, TableWrapperProps } from '../types';
```
Donya-qlik, line 10: I also think it is a more related name than RootProps  👍

Donya-qlik (approved):
lgtm!

PR #666, opened 2022-08-23
refactor: convert TableHeadWrapper to typescript

Donya-qlik:


src/table/components/TableHeadWrapper.tsx
```diff
@@ -19,16 +19,16 @@ function TableHeadWrapper({
...
   const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const headerStyle = useMemo(() => getHeaderStyle(layout, theme), [layout, theme]);
-  const headRowRef = useRef();
+  const headRowRef = useRef<HTMLElement>();
 
   useEffect(() => {
-    setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
+    headRowRef.current !== undefined && setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
```
haxxmaxx, line 31:
```suggestion
    headRowRef.current && setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
```

haxxmaxx, line 31:
```suggestion
    headRowRef.current && setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
```

src/table/components/__tests__/TableHeadWrapper.spec.tsx
```diff
@@ -124,33 +133,32 @@ describe('<TableHeadWrapper />', () => {
...
+    const secondColQuery = getByText(tableData.columns[1].label).closest('th');
 
-    expect(firstColQuery.getAttribute('aria-sort')).toBeNull();
-    expect(firstColQuery.getAttribute('aria-pressed')).toBe('false');
-    expect(secondColQuery.getAttribute('aria-sort')).toBe('ascending');
-    expect(secondColQuery.getAttribute('aria-pressed')).toBe('true');
+    expect(firstColQuery?.getAttribute('aria-sort')).toBeNull();
+    expect(firstColQuery?.getAttribute('aria-pressed')).toBe('false');
+    expect(secondColQuery?.getAttribute('aria-sort')).toBe('ascending');
+    expect(secondColQuery?.getAttribute('aria-pressed')).toBe('true');
```
haxxmaxx, line 150: Do need the ? here? I think getBy always returns an element, otherwise throws an error

src/types.ts
```diff
@@ -105,14 +105,16 @@ export interface Row {
   [key: string]: Cell | string;
 }
 
+type ColAlign = 'inherit' | 'left' | 'center' | 'right' | 'justify';
```
haxxmaxx, line 108: We never set this to inherit or justify. only `'left' | 'center' | 'right'` works. Then use it in `TextAlign` as well

src/table/components/__tests__/TableHeadWrapper.spec.tsx
haxxmaxx: Can it? I think the test is checking if aria-sort is null. This should fail btw since aria-sort is undefined if it is not sorted on that column

Donya-qlik: uh.. sorry forget my comment, doesn't matter using any of them... `firstColQuery` might be null anyway

haxxmaxx: ah I get it, you are doing `closest` that can return null. I would say it is better to cast it, otherwise you don't know if it is the attribute or the element that is null. Do something like this:
`const firstColQuery = getByText(tableData.columns[0].label).closest('th') as HTMLTableCellElement;
    const secondColQuery = getByText(tableData.columns[1].label).closest('th') as HTMLTableCellElement;`

src/table/components/TableHeadWrapper.tsx
```diff
@@ -39,7 +39,11 @@ function TableHeadWrapper({
           // when nebula does not handle keyboard navigation
           const tabIndex = columnIndex === 0 && !keyboard.enabled ? 0 : -1;
           const isCurrentColumnActive = layout.qHyperCube.qEffectiveInterColumnSortOrder[0] === column.dataColIdx;
-          const handleKeyDown = (evt) => {
+          const ariaSort = isCurrentColumnActive
+            ? (`${column.sortDirection}ending` as 'ascending' | 'descending')
+            : undefined;
+
+          const handleKeyDown = (evt: React.KeyboardEvent) => {
```
LiKang6688, line 46:
```suggestion
          const handleKeyDown = (evt: React.KeyboardEvent) => {
           const ariaSort = isCurrentColumnActive
            ? (`${column.sortDirection}ending` as 'ascending' | 'descending')
            : undefined;
```
ariaSort is only used inside handleKeyDown right?

src/types.ts
```diff
@@ -343,13 +345,18 @@ export interface RootProps {
   announce: Announce;
 }
 
-export interface BodyWrapperProps
-  extends Omit<RootProps, 'changeSortOrder' | 'rect' | 'pageInfo' | 'setPageInfo' | 'footerContainer' | 'translator'> {
+export interface TableBodyWrapperProps
```
LiKang6688, line 348: I am not a fan of this, is there no other way?

haxxmaxx: The name? or the omit? I will look into the best way to extend the prop types in a new pr

src/table/components/TableHeadWrapper.tsx
haxxmaxx: Nope, it is prop on TableCell

src/types.ts
LiKang6688: the omit.

src/table/components/__tests__/TableHeadWrapper.spec.tsx
Donya-qlik: Ok, I'll go for it

src/types.ts
Donya-qlik: I'll fix it with the next commit

src/types.ts
haxxmaxx: yeah it is a bit tricky. I might just define them independently. But that is a later problem

src/table/components/__tests__/TableHeadWrapper.spec.tsx
haxxmaxx: now you should be able to remove the ?

haxxmaxx (approved):
Remove the ? in the test and it is fine :)

PR #667, opened 2022-08-23
refactor: move types wip

haxxmaxx:
- dividing types into two files: one on a nebula/data level and one on a "table" level
- about the component props: there are no perfect way of doing this. Technically those components that spread the props from TableWrapper will recieve all those props. But they only use a subset of them, and it is different for all components. We are currently using omit, which this is not really the correct use-case for (it is more for omitting and overriding props). I decided to create a common interface for all components. Turns out that it is only 3 props, but better then nothing. You could potentially extend the type gradually, but I don't think that is very future proof. And they are kind of independent, so the chaining would only be for not repeating code
- Also note that I am keeping some types in the selection-utils. I think they fit nicely there since they are very much related to the dispatch


src/table/types.ts
```diff
@@ -0,0 +1,218 @@
...
+  cursor?: string;
+  borderWidth?: string;
+  backgroundColor?: string;
+  sortLabelColor?: string;
+  hoverBackgroundColor?: string;
+  hoverFontColor?: string;
+}
+
+export interface CellStyle {
+  backgroundColor: string | undefined; // This is always set but could be undefined in the theme
```
mhs30, line 57: Does it makes sense of making it optional?

src/table/types.ts
```diff
@@ -0,0 +1,218 @@
...
+  app?: EngineAPI.IApp;
+}
+
+export interface CommonTableProps {
+  tableData: TableData;
+  theme: ExtendedTheme;
+  keyboard: stardust.Keyboard;
+}
+
+export interface TableWrapperProps extends CommonTableProps {
```
mhs30, line 153: Should this one and the next two extend from `TableTotalProps` aswell since they all use rootElement and layout?

mhs30 (approved):
Minor comment, code lgtm

src/table/types.ts
haxxmaxx: well I don't think so. we always set it and sometimes it undefined

src/table/types.ts
haxxmaxx: I thought about it, but let's say we add a prop to totals that the other three don't need, then we would have to revert it to the way it is now. We will be adding a lot features soon so these props are likely to change

mhs30: Makes sense, then leave it as it is until you know more.

PR #669, opened 2022-08-26
fix: press space bar when totals is in focused causes table scrolling

Donya-qlik:


haxxmaxx:
This will solve this perticular verifail but my fix for the other ones will most likely change this, just so you know

haxxmaxx (approved):

PR #670, opened 2022-08-26
fix: seperately check if we should prevent default/stop prpogation

haxxmaxx:
Changing the way we prevent default and stop propagation for key handlers.
Now we first check if the event should bubble. If it should, we return early. If not, we always do `preventDefaultBehaviour`. Then comes the logic of figuring out what to do. Some of that logic can now be removed since it is done in the `shouldBubble` function. 

This makes sure that you can't do any interactions on totals and head when in selection mode, or on excluded cells


src/table/utils/handle-key-press.ts
```diff
@@ -114,6 +101,51 @@ export const moveFocus = (
...
+ * 3. Check if the event should bubble
+ * 4a. If it should bubble, early return and let handleWrapperKeyDown catch the event.
+ * If the key pressed is not relevant to the wrapper, it will keep bubbling (e.g. tab)
+ * 4b. If it shouldn't bubble, run preventDefaultBehavior and run the logic for that specific component
+ */
+
+/**
+ * handles keydown events for the tableWrapper element (move focus, change page)
+ */
+export const handleWrapperKeyDown = ({
```
haxxmaxx, line 118: this is just moved to structure the file better

src/table/utils/styling-utils.ts
```diff
@@ -259,6 +258,7 @@ export function getSelectionStyle(styling: CellStyle, cellSelectionState: Select
     case SelectionStates.EXCLUDED:
       selectionStyling = {
         background: `${STYLING_DEFAULTS.EXCLUDED_BACKGROUND}, ${styling.backgroundColor}`,
+        selectedCellClass: SelectionStates.EXCLUDED,
```
haxxmaxx, line 261: I don't think this should live on the styling. it simplifies thing but it is a bit misleading. Might refactor in a different PR

src/table/components/__tests__/TableTotals.spec.tsx
```diff
@@ -16,6 +16,7 @@ describe('<TableTotals />', () => {
   let layout: TableLayout;
   let selectionsAPI: ExtendedSelectionAPI;
   let keyboard: stardust.Keyboard;
+  let isSelectionMode: boolean;
```
niekvanstaveren, line 19: Seems like you are not changing this

src/table/utils/handle-key-press.ts
```diff
@@ -104,8 +92,7 @@ export const moveFocus = (
   setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>,
   topAllowedRow?: number
 ) => {
-  preventDefaultBehavior(evt);
-  (evt.target as HTMLElement).setAttribute('tabIndex', '-1');
+  updateFocus({ focusType: 'removeTab', cell: evt.target as HTMLTableCellElement });
   const nextCellCoord = getNextCellCoord(evt, rootElement, cellCoord, topAllowedRow);
   const nextCell = getCellElement(rootElement, nextCellCoord);
   updateFocus({ focusType: 'focus', cell: nextCell });
```
niekvanstaveren, line 98: Does it make sense to call updateFocus twice?

niekvanstaveren (approved):
Code looks okay to me. Not completely sure on everything that would be impacted though

src/table/components/__tests__/TableTotals.spec.tsx
haxxmaxx: or just `isSelectionMode={false}` below?

niekvanstaveren: Yes either of them 😄

PR #673, opened 2022-08-27
feat: add TotalsAggrFuncMessage to sn-table property panel

Donya-qlik:


src/handle-data.ts
```diff
@@ -85,6 +85,16 @@ export function getTotalPosition(layout: TableLayout) {
...
+export function getProperties(reference: string, data: any, defaultValue?: undefined) {
+  const steps: string[] = reference.split('.');
+  if (typeof data === 'undefined') return defaultValue;
+  let value = data;
+  for (let i = 0; i < steps.length; ++i) {
+    value[steps[i]] == null ? () => defaultValue : (value = value[steps[i]]);
+  }
+  return value;
+}
+
```
haxxmaxx, line 97: This looks like optional chaining to me, especially since the current use-case does not pass a default value. Have you tried that? Also not sure this should be in handle-data

Donya-qlik: Yes it is true, the default value doesn't paly any role here...
It is actually what is used in sense-client, I was not sure about the conditions, I just kept it because I thought we might reuse the function somewhere else. I'll remove it.

Donya-qlik: BTW, data type is `HypercubeHandler` I don't think we have defined such a type anywhere.

haxxmaxx: hmm no, I don't recognize it. But it is only in the ext file right?

Donya-qlik: Yes

haxxmaxx: Then don't mind it. If we want to refactor and convert ext file that becomes relevant

src/handle-data.ts
```diff
@@ -85,6 +85,15 @@ export function getTotalPosition(layout: TableLayout) {
...
 }
 
+export function getProperties(reference: string, data: any) {
+  const steps: string[] = reference.split('.');
+  let value = data;
+  for (let i = 0; i < steps.length; ++i) {
+    if (value[steps[i]] !== null) value = value[steps[i]];
+  }
+  return value;
+}
```
haxxmaxx, line 95: I meant that you don't need this function at all. It should do the same thing as optional chaining. It was not a part of the language when this function was written, not needed anymore

Donya-qlik: uh... I don't know why I was thinking about having this func all the time... hadn't even thought about that. 
Removed it.

src/ext.js
```diff
@@ -403,6 +403,17 @@ const getDefinition = (env) => {
...
                         defaultValue: env?.anything?.sense?.isUnsupportedFeature?.('totals') ? 'None' : 'Expr',
                       },
+                      totalsAggrFuncMessage: {
+                        component: 'text',
+                        translation: 'Object.Table.AggrFuncMessage',
+                        show(data, handler) {
+                          return (
+                            handler &&
+                            !handler.properties.totals.show &&
+                            handler.properties.totals.position === 'noTotals'
```
Donya-qlik, line 413: Is it working. Is it all that should be kept?

haxxmaxx: also if you switch the order you don't need to check handler first
```suggestion
                            handler.properties.totals.position === 'noTotals' &&
                            !handler.properties.totals.show

```

Donya-qlik: Data does not include properties but handler does. 
As far as I tested it always exists, it just might be shown or not regarding the conditions.
`totals.show` is in higher priority here since with `show = true` it doesn't need to go for checking position at all,  position can stay noTotals in both cases.

haxxmaxx: ok then handler works fine.
Well if `position` is not 'noTotals' there is no need to check `show`. My point is that you can remove the first condition on line 411

Donya-qlik: ok, right. I'll switch them.

src/ext.js
```diff
@@ -403,6 +403,17 @@ const getDefinition = (env) => {
                         },
                         defaultValue: env?.anything?.sense?.isUnsupportedFeature?.('totals') ? 'None' : 'Expr',
                       },
+                      totalsAggrFuncMessage: {
+                        component: 'text',
+                        translation: 'Object.Table.AggrFuncMessage',
+                        show(data, handler) {
+                          return (
+                            handler &&
+                            handler.properties.totals.position === 'noTotals' &&
```
haxxmaxx, line 412:
```suggestion
                            handler?.properties.totals.position === 'noTotals' &&
```

PR #675, opened 2022-08-29
test: ensure the running container is stopped even when a test breaks

LiKang6688:
Use exit tap to ensure the running container `sn-table-playwright` is stopped even when a test breaks.


haxxmaxx:
Good idea but this means even more duplicate code. The only thing that differs is the update flag. it should be passed as a flag/argument instead.

LiKang6688:
> Good idea but this means even more duplicate code. The only thing that differs is the update flag. it should be passed as a flag/argument instead.

Right, sure, I will do the refactor in the next PR.

LiKang6688:
In that case, there will be only one bash script. What file name do you suggest?

haxxmaxx:
> In that case, there will be only one bash script. What file name do you suggest?

run-rendering-tests works, you are still running them when you update the snapshots

haxxmaxx:
> Right, sure, I will do the refactor in the next PR.

I think you should do it the other way around though. Make it into one file before adding code to both files

haxxmaxx (approved):

PR #677, opened 2022-08-29
fix: move total to reduce rerenders

haxxmaxx:
Previously the totals component was passed as a child to the body. This was to reduce props drilling. Totals is memoed using React.memo and it used to work works. This is what the profiler looks like, the totals don't rerender when the body does. But it is the children property that makes the body update (I have a snippet to test this). Feels really backwards to me, but maybe that is how is should work. Anyway, adding totals inside the body component instead, works!


src/table/components/TableBodyWrapper.tsx
```diff
@@ -50,9 +50,13 @@ function TableBodyWrapper({
     });
   }, []);
 
+  const totals = (
+    <TableTotals rootElement={rootElement} tableData={tableData} theme={theme} layout={layout} keyboard={keyboard} />
+  );
```
haxxmaxx, line 55: I don't know if this is considered less props drilling? still the same properties I pass through, just not as `{...props}`

Donya-qlik (approved):

PR #678, opened 2022-08-30
chore!: remove module section from package.json

LiKang6688:
head:
Now the core/esm/index.js does not have react and react-dom code.

BREAK CHANGE:
for the mashup, people have to install react and react-dom if they want to use sn-table using es module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "node": ">=16"
   },
   "main": "dist/sn-table.js",
-  "module": "core/esm/index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


Caele (approved):

LiKang6688:
One weird thing:
when I test locally in our mashup example - qlik-sense-desktop/configure.js using
`import table from '../../../dist/sn-table';`
Got an error message:
Uncaught SyntaxError: The requested module '/@fs/Users/gqq/Qlik/sn-table/dist/sn-table.js' does not provide an export named 'default' 

But it works correctly using `import table from '@nebula.js/sn-table';`

PR #679, opened 2022-08-30
fix: possibility of focus on totals cell when cell is in selection mode

Donya-qlik:


src/table/utils/handle-key-press.ts
```diff
@@ -187,16 +191,20 @@ export const handleBodyKeyDown = ({
...
-  const cellCoord: [number, number] = [cell.rawRowIdx + firstBodyRowIdx, cell.rawColIdx];
+  const rowIdxDiff = totalsPosition === 'top' ? 2 : 1;
+
+  const cellCoord: [number, number] = [cell.rawRowIdx + rowIdxDiff, cell.rawColIdx];
   // Make sure you can't navigate to header (and totals) in selection mode
-  const topAllowedRow = isSelectionMode ? firstBodyRowIdx : 0;
+  const allowedRows = {
+    cellIdx: isSelectionMode ? rowIdxDiff : 0,
+    totalsPosition,
+  };
```
haxxmaxx, line 201: I like the idea of putting this on an object. But I think makes more sense to calculate both values in here and not add logic in `getNextCellCoord`. So if you do something like this here:
```suggestion
  const allowedRows = {
    cellIdx: isSelectionMode ? rowIdxDiff : 0,
    bottom: isSelectionMode && totalsPosition === 'bottom' ? 1 : 0,
  };
```

you don't need the if statement in `getNextCellCoord`

src/table/utils/handle-key-press.ts
```diff
@@ -187,16 +191,20 @@ export const handleBodyKeyDown = ({
 }: HandleBodyKeyDownProps) => {
   const isSelectionMode = selectionsAPI.isModal();
   // Adjust the cellCoord depending on the totals position
-  const firstBodyRowIdx = totalsPosition === 'top' ? 2 : 1;
-  const cellCoord: [number, number] = [cell.rawRowIdx + firstBodyRowIdx, cell.rawColIdx];
+  const rowIdxDiff = totalsPosition === 'top' ? 2 : 1;
```
haxxmaxx, line 194: I'm not sure this name is better, is the old one not clear?

src/table/utils/handle-key-press.ts
```diff
@@ -54,22 +54,26 @@ export const getNextCellCoord = (
...
+      nextRow > allowedRows.cellIdx && nextRow--;
       break;
     case 'ArrowRight':
-      // topAllowedRow greater than 0 means we are in selection mode
-      if (topAllowedRow > 0) break;
+      // allowedRows.cellIdx greater than 0 means we are in selection mode
+      if (allowedRows.cellIdx > 0) break;
       if (nextCol < columnCount - 1) {
         nextCol++;
       } else if (nextRow < rowCount - 1) {
```
haxxmaxx, line 79: you probably need to add something here since you can press right arrow to go to next column

haxxmaxx: the next row I mean of course

src/table/utils/handle-key-press.ts
Donya-qlik: yes, sure. I had another idea first then changed it but forgot to undo the name.

src/table/utils/handle-key-press.ts
Donya-qlik: Yes ofcourse! good idea!

src/table/utils/handle-key-press.ts
Donya-qlik: Ah, you mean it should not move to Total row then, I thought it's how it works.. I compared it with the old table so I'll fix it.

haxxmaxx: exactly. Yes this is a new feature, makes it a bit more accessible I think

src/table/utils/handle-key-press.ts
```diff
@@ -54,35 +54,42 @@ export const getNextCellCoord = (
...
   cellCoord: [number, number],
-  topAllowedRow = 0
+  allowedRows: {
+    selected: {
+      cellIdx: number;
+      bottom: number;
+    };
+    cellIdx: number;
+    bottom: number;
+  } = { selected: { cellIdx: 0, bottom: 0 }, cellIdx: 2, bottom: 1 }
```
Donya-qlik, line 64: It went confusing so I decided to set the object with both selected and without selection cell index. 😣

src/table/utils/handle-key-press.ts
```diff
@@ -188,15 +202,23 @@ export const handleBodyKeyDown = ({
...
   // Make sure you can't navigate to header (and totals) in selection mode
-  const topAllowedRow = isSelectionMode ? firstBodyRowIdx : 0;
+  const allowedRows = {
+    selected: {
+      top: isSelectionMode ? firstBodyRowIdx : 0,
+      bottom: isSelectionMode && totalsPosition === 'bottom' ? 1 : 0,
+    },
+    top: firstBodyRowIdx,
+    bottom: totalsPosition === 'bottom' ? 2 : 1,
+  };
```
haxxmaxx, line 215: This should not be needed, on value for top and bottom should be enough. Is there some case that it isn't working for? I tried to change it to this
```suggestion
  const allowedRows = {
    top: isSelectionMode ? firstBodyRowIdx : 0,
    bottom: isSelectionMode && totalsPosition === 'bottom' ? 1 : 0,
  };
```
and it seemed to work for me

src/table/utils/handle-key-press.ts
```diff
@@ -54,35 +54,42 @@ export const getNextCellCoord = (
...
       break;
     case 'ArrowRight':
-      // topAllowedRow greater than 0 means we are in selection mode
-      if (topAllowedRow > 0) break;
+      // allowedRows.top greater than 0 means we are in selection mode
+      if (allowedRows.selected.top > 0) break;
       if (nextCol < columnCount - 1) {
         nextCol++;
-      } else if (nextRow < rowCount - 1) {
+      } else if (nextRow < rowCount - 1 - allowedRows.bottom) {
```
haxxmaxx, line 82: I was wrong here. In selection mode you can not go left or right at all, so no need to change anything here :)

src/table/utils/handle-key-press.ts
```diff
@@ -54,35 +54,42 @@ export const getNextCellCoord = (
...
       break;
     case 'ArrowLeft':
-      // topAllowedRow greater than 0 means we are in selection mode
-      if (topAllowedRow > 0) break;
+      // allowedRows.top greater than 0 means we are in selection mode
+      if (allowedRows.selected.top > 0) break;
       if (nextCol > 0) {
         nextCol--;
-      } else if (nextRow > 0) {
+      } else if (nextRow > allowedRows.top) {
```
haxxmaxx, line 92: same here, you left arrow is blocked in selection mode

src/table/utils/handle-key-press.ts
Donya-qlik: well...not totally. Out of selection mode arrow used to move the cell to Totals row, it is not how the old table works. Now with the new condition it works fine and  Totals row is excluded. I fixed it for Top case as well.

haxxmaxx: Sorry I don't understand. On row 79 we make sure that ArrowRight does not do anything when in selection mode. When not in selection mode, you should be able to reach totals, both with up and down, and left and right

haxxmaxx: The left and right arrow was working correctly before this fix, so no need to change anything in the right and left case

Donya-qlik: Umm... ok. I was not aware of it. Just for being sure, in old table Totals row is not reachable by left/right arrow in both selection and normal mode, just up/down works, in the current table all works as moving to another cell except in selection mode, correct?

haxxmaxx: yes. if you only press right it goes through the cells like a book: left to right, then to to the start of the row below. The only thing you need to fix in this PR is the condition for the down arrow

Donya-qlik: Ok, I'll revert the change then.

src/table/utils/handle-key-press.ts
haxxmaxx: this line should still be reverted

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -274,6 +274,17 @@ describe('handle-key-press', () => {
       expect(nextCol).toBe(0);
     });
 
+    it('should go one cell forward when the last cell is not selected and the key is pressed down to bottom Totals row', () => {
```
haxxmaxx, line 277: This description does not make sense. And I don't see how this test is needed. What is new in this PR is that you don't go to the last row if allowedRows.bottom === 1. That could use a test

src/table/utils/handle-key-press.ts
Donya-qlik: Sorry for these mistakes, things went mess when I reverted the last changes, I see my selection mode testing scenario is gone and this guy is left here. Thanks for considering them. I'll fix it.

haxxmaxx (approved):
great work!

PR #681, opened 2022-08-31
fix: ctrl and shift key focus issue on Totals

Donya-qlik:


src/table/utils/handle-accessibility.ts
```diff
@@ -94,7 +97,7 @@ export const handleResetFocus = ({
       const hasSelectedClassName = cell?.classList?.contains('selected');
       announce({
         keys: [
-          `${cell.textContent},`,
+          `${cell?.textContent},`,
```
Donya-qlik, line 100: not sure if it is correct... the added test only works if I set this optional

haxxmaxx: good question. In getCellElement we cast the result to HTMLCellElement, but it seems like it might be undefined. I'll look into it

src/table/utils/handle-accessibility.ts
```diff
@@ -79,10 +79,13 @@ export const handleResetFocus = ({
...
   keyboard,
   announce,
+  totalsPosition,
 }: HandleResetFocusProps) => {
   updateFocus({ focusType: 'removeTab', cell: findCellWithTabStop(rootElement) });
   // If you have selections ongoing, you want to stay on the same column
-  const cellCoord: [number, number] = isSelectionMode ? [1, focusedCellCoord[1]] : [0, 0];
+  const cellCoord: [number, number] = isSelectionMode
+    ? [totalsPosition === 'top' ? 2 : 1, focusedCellCoord[1]]
+    : [0, 0];
```
haxxmaxx, line 88: I would split this up. Nested ternaries are a bit hard to read. Something like this
```suggestion
    const selectionCellCoord = [totalsPosition === 'top' ? 2 : 1, focusedCellCoord[1]];
    const cellCoord: [number, number] = isSelectionMode ? selectionCellCoord : [0, 0];
```

src/table/utils/handle-accessibility.ts
haxxmaxx: We always assume that `getCellElement` gets a valid `cellCoord`, so it will return a cell. What you need to do in your tests is to mock `rootElement` as in the test above it

Donya-qlik: Yes, I had tried it before and didn't work, also added a mocked cell textContent and didn't pass either. the problem seems is  `totalsPosition`, without that it works fine. I've no idea what change it makes that ends up to an undefined cell?

haxxmaxx: hmm ok let me have a look

haxxmaxx: the reason is that there is now third row in the mock (index 2).
``` 
      rootElement = {
        getElementsByClassName: () => [row, row],
        querySelector: () => cell,
      } as unknown as HTMLElement;
```
As you see here we add only two rows in the return of getElementsByClassName, so it will be undefined for index 2. Add an extra row there and it works

haxxmaxx (approved):
looks good. Fix the last comment if you want, I'm fine with both

PR #684, opened 2022-09-06
feat: To be able to copy table cell via keyboard

Donya-qlik:


src/table/utils/handle-accessibility.ts
```diff
@@ -145,3 +145,16 @@ export const announceSelectionState = (
...
+  textElement.style.position = 'absolute';
+  textElement.style.left = '-9999px';
+  textElement.setAttribute('readonly', '');
+
+  textElement.innerText = cell.qText as string;
+  document.body.appendChild(textElement);
+  textElement.select();
+  !document.execCommand('copy') && alert('The content is not copied.');
+  document.body.removeChild(textElement);
+};
```
Donya-qlik, line 151: This is the old version by using execCommand but then I replaced it with ClipboardAPI in the next commit.

src/table/utils/handle-accessibility.ts
```diff
@@ -145,3 +145,7 @@ export const announceSelectionState = (
       : announce({ keys: ['SNTable.SelectionLabel.NotSelectedValue'] });
   }
 };
+
+export const handleKeyToCopyCell = (cell: Cell) => {
+  !navigator.clipboard.writeText(cell.qText as string) && alert('The content is not copied!');
```
Donya-qlik, line 150: Not sure alert is what we want to have here but thought is good to be for failed cases.

src/table/utils/handle-key-press.ts
```diff
@@ -301,6 +307,9 @@ export const handleBodyKeyDown = ({
     case 'Tab':
       focusSelectionToolbar(evt.target as HTMLElement, keyboard, evt.shiftKey);
       break;
+    case 'c':
+      evt.ctrlKey && evt.key === 'c' && handleKeyToCopyCell(cell);
```
haxxmaxx, line 311: should only need to check for c in the case, `evt.key === 'c'` shouldn't be needed.

src/table/utils/handle-accessibility.ts
haxxmaxx: writeText is asynchronous, needs to be an await before it and if we want to catch the error we need a try catch block. Should look something like this

```suggestion
  try {
    await navigator.clipboard.writeText(cell.qText as string);
  } catch (error) {
    alert(error);
  }
```

you also need to add `async` to the function definition.

It is a good idea to have some kind of indication of what is going on and that we could potentially get an error back. But we don't to do it with alert, it is not that big of a deal. Not sure what to do here. potentially just log the error in the consol

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -417,4 +417,34 @@ describe('handle-accessibility', () => {
...
+      cellValue = {
+        qText: '251',
+      } as Cell;
+    });
+    it('should copying value to clipboard successfully', () => {
+      Object.assign(navigator, {
+        clipboard: {
+          writeText: jest.fn((value) => value),
+        },
+      });
```
haxxmaxx, line 433: this is not a proper mock of `writeText`. have you tried using the real function? If that works you can test it by running the method`readText` to see if the text is on the clipboard

src/table/utils/handle-key-press.ts
haxxmaxx: also we need meta key here as well, that is cmd for mac

src/table/utils/handle-accessibility.ts
haxxmaxx: or the other way around, indicate that we have successfully copied to clipboard. But that is not in the old implementation, so no need to do that now

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -805,6 +806,13 @@ describe('handle-key-press', () => {
       expect(announce).not.toHaveBeenCalled();
     });
 
+    it('should call handleKeyToCopyCell only when the pressed keys are Ctrl and C keys', () => {
```
cbt1, line 809: Also add test for the metakey?

Donya-qlik: Sure, It is added now.

src/table/utils/__tests__/handle-accessiblity.spec.ts
Donya-qlik: I couldn't make it work without mocking, properties will be undefined.

src/table/utils/handle-accessibility.ts
```diff
@@ -145,3 +145,11 @@ export const announceSelectionState = (
       : announce({ keys: ['SNTable.SelectionLabel.NotSelectedValue'] });
   }
 };
+
+export const handleKeyToCopyCell = async (cell: Cell) => {
```
haxxmaxx, line 149: This is not really a handler. Functions that handles events. This is just a generic function  copying the value of the cell. Rename to `copyCellValue`

haxxmaxx: also, it has nothing to do with a11y. But let's keep it here for now, not sure where it should be

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -417,4 +417,34 @@ describe('handle-accessibility', () => {
...
+      handleAccessibility.handleKeyToCopyCell(cellValue);
+      expect(console.log).not.toHaveBeenCalled();
+    });
+
+    it('should fail copying value to clipboard', () => {
+      Object.defineProperty(navigator, 'clipboard', {
+        value: {
+          writeText: !jest.fn().mockReturnValueOnce(Promise.resolve()),
```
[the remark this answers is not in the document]
haxxmaxx: Hmm actually it seems a bit trickier that that. Feel free to comment this test out, since we don't know what we want to do for the error handling anyway

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -623,6 +623,7 @@ describe('handle-key-press', () => {
       totalsPosition = 'noTotals';
       jest.spyOn(handleAccessibility, 'focusSelectionToolbar').mockImplementation(() => jest.fn());
       jest.spyOn(handleAccessibility, 'announceSelectionState').mockImplementation(() => jest.fn());
+      jest.spyOn(handleAccessibility, 'handleKeyToCopyCell').mockImplementation(() => Promise.resolve());
```
haxxmaxx, line 626: Do you need to mock the implementation or is spyOn enough?

src/table/utils/handle-key-press.ts
```diff
@@ -301,6 +307,9 @@ export const handleBodyKeyDown = ({
     case 'Tab':
       focusSelectionToolbar(evt.target as HTMLElement, keyboard, evt.shiftKey);
       break;
+    case 'c':
+      (evt.ctrlKey || evt.metaKey) && handleKeyToCopyCell(cell);
```
haxxmaxx, line 311: We need a feature flag here before this is merged. I will find out what it should be

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -417,4 +417,34 @@ describe('handle-accessibility', () => {
...
+        clipboard: {
+          writeText: jest.fn().mockReturnValueOnce(Promise.resolve()),
+        },
+      });
+      handleAccessibility.handleKeyToCopyCell(cellValue);
+      expect(console.log).not.toHaveBeenCalled();
+    });
+
+    it('should fail copying value to clipboard', () => {
+      Object.defineProperty(navigator, 'clipboard', {
```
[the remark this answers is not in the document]
Donya-qlik: Yes it's actually possible

src/table/utils/handle-accessibility.ts
Donya-qlik: what is a11y? 😅

haxxmaxx: accessibility

src/table/utils/__tests__/handle-accessiblity.spec.ts
```diff
@@ -417,4 +417,34 @@ describe('handle-accessibility', () => {
       expect(announce).toHaveBeenCalledWith({ keys: ['SNTable.SelectionLabel.NotSelectedValue'] });
     });
   });
+
+  describe('handleKeyToCopyCell: ', () => {
+    let cellValue: Cell;
+    beforeEach(() => {
```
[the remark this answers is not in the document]
haxxmaxx: ah look at that, nice

src/table/components/TableBodyWrapper.tsx
```diff
@@ -76,7 +78,7 @@ function TableBodyWrapper({
             const { id, align } = column;
             const cell = row[id] as Cell;
             const CellRenderer = columnRenderers[columnIndex];
-            const handleKeyDown = (evt: React.KeyboardEvent) => {
+            const handleKeyDown = (evt: React.KeyboardEvent, flag: stardust.Flags = flags) => {
```
haxxmaxx, line 81: This shouldn't be needed at all. flags is available on the scope, no need to pass it as an argument to this function

src/types.ts
```diff
@@ -198,4 +199,5 @@ export interface RenderWithCarbonArguments {
   rect: stardust.Rect;
   layout: TableLayout;
   changeSortOrder?: ChangeSortOrder;
+  flags: stardust.Flags;
```
haxxmaxx, line 202: remember to remove it here as well

src/index.ts
```diff
@@ -50,13 +50,15 @@ const renderWithCarbon = ({
...
+  flags,
 }: RenderWithCarbonArguments) => {
   if (env.carbon && changeSortOrder && theme) {
-    render({ rootElement, layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect });
+    render({ rootElement, layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect, flags });
   }
 };
 
 export default function supernova(env: Galaxy) {
+  const { flags } = env;
```
haxxmaxx, line 61: We only need isEnabled, so you can destruct that and only pass that function along

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -623,10 +627,16 @@ describe('handle-key-press', () => {
...
+      jest.spyOn(handleAccessibility, 'copyCellValue');
       jest.spyOn(handleScroll, 'handleNavigateTop').mockImplementation(() => jest.fn());
     });
 
-    afterEach(() => jest.clearAllMocks());
+    afterEach(() => {
+      jest.clearAllMocks();
+      flags = {
+        isEnabled: () => true,
+      };
```
haxxmaxx, line 638: doesn't really make sense to do this in the afterEach, should be in the beforeEach

src/table/types.ts
```diff
@@ -143,6 +144,7 @@ export interface RenderProps {
     setPageInfo: SetPageInfo
   ): Promise<TableData | null>;
   app?: EngineAPI.IApp;
+  flags: stardust.Flags;
```
haxxmaxx, line 147: should probably be optional since it is not in the native render function

src/index.ts
```diff
@@ -50,13 +50,15 @@ const renderWithCarbon = ({
   rect,
   layout,
   changeSortOrder,
+  flags,
 }: RenderWithCarbonArguments) => {
   if (env.carbon && changeSortOrder && theme) {
-    render({ rootElement, layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect });
+    render({ rootElement, layout, model, manageData, theme, selectionsAPI, changeSortOrder, app, rect, flags });
```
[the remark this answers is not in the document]
Donya-qlik: The reason is I added the flag as a prop so it will be part of RenderProps, wherever I use render() I should add it otherwise I'll get error.

src/index.ts
Donya-qlik: I thought it is more readable if I pass it as `flags` than `isEnable` but can change it.

haxxmaxx: you can just rename it to `isFlagEnabled` if you want

src/table/components/TableBodyWrapper.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect, useMemo, memo } from 'react';
+import { stardust } from '@nebula.js/stardust';
```
haxxmaxx, line 2: remove this

src/table/types.ts
```diff
@@ -94,6 +94,7 @@ export interface HandleBodyKeyDownProps {
   totalsPosition: TotalsPosition;
   paginationNeeded: boolean;
   selectionsAPI: ExtendedSelectionAPI;
+  isFlagEnabled?(flag: string): boolean;
```
haxxmaxx, line 97: This one shouldn't be optional. We cast to this and then we know that these props exist

Donya-qlik:
> Sorry if I was unclear. You could merge this PR before mine, but no without approval... Looks good though, just one small comment that would be nice if you clean up

You had approved it this morning but after fixing the later comments I merge it. Sorry was my fault, I should have sent you another request for the final approval. I created a new PR here: https://github.com/qlik-oss/sn-table/pull/690

haxxmaxx:
Yes sorry I did, my bad

PR #688, opened 2022-09-12
chore: enhance airbnb eslint config with typescript support

LiKang6688:


.eslintrc.json
```diff
@@ -1,16 +1,12 @@
...
   },
   "parserOptions": {
-    "sourceType": "module",
-    "babelOptions": {
-      "presets": ["@babel/preset-react"]
-    }
+    "project": ["./tsconfig.json"]
   },
-  "extends": ["airbnb", "prettier", "plugin:react/recommended"],
+  "extends": ["airbnb", "airbnb-typescript", "prettier"],
```
LiKang6688, line 9: airbnb includes eslint, eslint-plugin-import, eslint-plugin-react, eslint-plugin-react-hooks, and eslint-plugin-jsx-a11y https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb#eslint-config-airbnb-1

.eslintrc.json
```diff
@@ -1,16 +1,12 @@
...
-  "parser": "@babel/eslint-parser",
   "env": {
     "browser": true
   },
   "parserOptions": {
-    "sourceType": "module",
-    "babelOptions": {
-      "presets": ["@babel/preset-react"]
-    }
+    "project": ["./tsconfig.json"]
```
LiKang6688, line 7: it is needed for airbnb-typescript

.eslintrc.json
```diff
@@ -47,15 +43,10 @@
...
+    "@typescript-eslint/no-shadow": ["error"],
+    "@typescript-eslint/no-unused-expressions": "warn"
   },
   "overrides": [
-    {
-      "files": ["**/*.{ts,tsx}"],
-      "parser": "@typescript-eslint/parser",
-      "extends": ["plugin:@typescript-eslint/recommended"],
-      "plugins": ["jest", "react", "@typescript-eslint"]
-    },
```
LiKang6688, line 58: now it is a default one, airbnb-typescript applyes @typescript-eslint/eslint-plugin and @typescript-eslint/parser

.eslintrc.json
```diff
@@ -1,16 +1,12 @@
 {
   "root": true,
-  "parser": "@babel/eslint-parser",
```
LiKang6688, line 3: it is replaced with @typescript-eslint/parser

src/table/utils/selections-utils.ts
```diff
@@ -2,6 +2,7 @@
 /* eslint-disable @typescript-eslint/no-empty-interface */
 import { stardust } from '@nebula.js/stardust';
 import { Cell, ExtendedSelectionAPI, Announce } from '../../types';
+// eslint-disable-next-line import/no-cycle
```
LiKang6688, line 5: Please fix it when you have time

src/table/types.ts
```diff
@@ -14,6 +14,7 @@ import {
   TableLayout,
   TotalsPosition,
 } from '../types';
+/* eslint-disable-next-line import/no-cycle */
 import { TSelectionActions } from './utils/selections-utils';
```
LiKang6688, line 18: Please fix it when you have time

.eslintrc.json
```diff
@@ -47,15 +43,10 @@
     "react/jsx-filename-extension": [2, { "extensions": [".js", ".jsx", ".ts", ".tsx"] }],
     "@typescript-eslint/ban-ts-comment": ["error", { "ts-ignore": "allow-with-description" }],
     "no-shadow": "off",
-    "@typescript-eslint/no-shadow": ["error"]
+    "@typescript-eslint/no-shadow": ["error"],
+    "@typescript-eslint/no-unused-expressions": "warn"
```
LiKang6688, line 47: you can decide if this should be an error or not. I change the pattern from `a && b` to `if(a) b` because of it

Donya-qlik:
Thanks Li for this improvement! Sorry that I have zero knowledge about airbnb eslint and the following requirements so I prefer to pass reviewing duty to experts.

haxxmaxx:
we have done some refactoring, so this needs some conflict fixing

src/table/types.ts
LiKang6688: no, there is no problem when compiling and after compiling. It is only an eslint issue.

.eslintrc.json
LiKang6688: Good.
allowShortCircuit allows  a && b.
allowTernary allows a ? b : c.
I will make both be true

.eslintrc.json
```diff
@@ -47,15 +43,10 @@
     "react/jsx-filename-extension": [2, { "extensions": [".js", ".jsx", ".ts", ".tsx"] }],
     "@typescript-eslint/ban-ts-comment": ["error", { "ts-ignore": "allow-with-description" }],
     "no-shadow": "off",
-    "@typescript-eslint/no-shadow": ["error"]
+    "@typescript-eslint/no-shadow": ["error"],
+    "@typescript-eslint/no-unused-expressions": ["warn", { "allowShortCircuit": true, "allowTernary": true }]
```
haxxmaxx, line 47: But you still have changed the code. We should either not have the exception to this rule or change the code, not both

LiKang6688: I see. I gonna revert the change from a &&b to if(a) b

haxxmaxx:
Sorry if I was unclear. I think there is no need to change any of these uses, both if() and && are fine with me at least. It depends. If there is a really good reason to pick one of them, we should consider it. Otherwise no need to make more changes than necessary IMO

LiKang6688:
> Sorry if I was unclear. I think there is no need to change any of these uses, both if() and && are fine with me at least. If there is a really good reason to pick one of them, we should consider it. Otherwise no need to make more changes than necessary IMO

Sure. We can do those changes in another PR if needed.

haxxmaxx (approved):

PR #689, opened 2022-09-13
feat: Introduce drag with mouse to select multiple cells

haxxmaxx:
Adding functionality to to select multiple cells with mouse.

- creating/renaming selection actions: `SELECT_MULTI_START/ADD/END`
  - `START` runs on mousedown and changes the `isSelectMultiValues` to true, nothing added to selection state at this point
  - `ADD` runs when you hold mouse button (or hold shift on keyboard) and move to the next cell. this adds both cells to the state but no back-end call
    - Not the same as old table. It actually adds the first cell if you drag without leaving the first element. Could be a later improment, not even sure user will tell the difference. I've been playing with `mousemove` here, but can't find a clean solution for it atm.
  - `END` runs when you release mouse button (or release shift on keyboard), sending the selections to the back-end and sets `isSelectMultiValues` to false, i.e. same as before
- Refactor so select multiple has a separate function, used both mouse and keyboard
- Currently keeping focus logic on mousedown. This means that after you drag, the focus is not where mouseup is. Natively focus happens on mouse down so I think this most sense. This would need to be tested if it feels right.
- Adding `allRows` to the selection state. Since if you are moving the mouse fast, there wont be an  event for each cell. That means we need to add all cells between where the dragging started and the current mouse position. This means that we can't rely on values being stored on each cell component. So I decided to add all of them to the selection state for easy access. This means that we can remove next/prev cell currently stored on a cell. Wont be done in this PR
  -  This means that we need to keep them updated, so there is a specific action for that: `UPDATE_ALL_ROWS`. Since we never use the property in any component, updating this only triggers a rerender on the context "component" 👍 

This is not the full feature, but an initial commit behind feature flag. Not tested on touch yet. Also should run  `SELECT_MULTI_END` on mouseup outside the table, not sure how to do that yet, will be done in a later PR. Despite this, PR got pretty big. Sorry about that, the majority of lines are tests though


src/table/utils/selections-utils.ts
```diff
@@ -144,35 +161,91 @@ export const getSelectedRows = (
...
+  }
+
+  if (!firstCell) return selectedRows;
+  const lowestCellIdx = Math.min(firstCell.rawRowIdx, cell.rawRowIdx);
+  const highestCellIdx = Math.max(firstCell.rawRowIdx, cell.rawRowIdx);
+
+  for (let idx = lowestCellIdx; idx <= highestCellIdx; idx++) {
+    const selectedCell = allRows[idx][`col-${firstCell.rawColIdx}`] as Cell;
+    newSelectedRows[selectedCell.qElemNumber] = selectedCell.rowIdx;
+  }
```
haxxmaxx, line 191: suggestions of smarter ways of doing this are appreciated. If we have to loop all cells from `firstCell` to `cell`, I think this is the fastest way

src/table/utils/handle-click.ts
```diff
@@ -28,3 +30,40 @@ export const handleMouseDownLabelToFocusHeadCell = (evt: MouseEvent, rootElement
...
+
+  const handleMouseOver = (evt: React.MouseEvent) => {
+    if (selectMultiEnabled && evt.button === 0 && evt.buttons === 1)
+      selectionDispatch({ type: SelectionActions.SELECT_MULTI_ADD, payload: { cell, evt, announce } });
+  };
+
+  const handleMouseUp = (evt: React.MouseEvent) => {
+    // update focusCellCoord and manually focus the element
+    onClick?.(evt as React.MouseEvent<HTMLTableCellElement>);
+    (evt.target as HTMLElement).focus();
```
MTjody, line 59: What's up with the forced casting?

src/table/types.ts
```diff
@@ -29,13 +32,14 @@ export interface ActionPayload {
   evt: React.KeyboardEvent | React.MouseEvent;
 }
 
+export type SelectionDispatch = React.Dispatch<TSelectionActions> | jest.Mock<any, any>;
```
MTjody, line 35: IMO test code should not leak into production code. I'd rather put a union type in a test/types file that adds the mock typing to `SelectionDispatch`. But I also saw that you just moved it, so never mind I guess. End rant :)

MTjody (commented):
Looks good Max! Two very minor comments, I struggle to understand when this PR or feature is considered done also, so you could perhaps clarify that 👏

src/table/utils/selections-utils.ts
MTjody: LGTM honestly, trusty old for-loop does the trick

src/table/types.ts
haxxmaxx: haha yeah I see your point. I looked at that recently, not even sure we need to pass that mock to be able to test it. I will double-check. Not going to do anything about it in this PR though

src/table/utils/handle-click.ts
haxxmaxx: I think we need to re-visit the typing for the events. I think the target here is not always a proper element, unless you specify that. In this case we know what the target is and can assume that `focus` is on there

src/table/types.ts
```diff
@@ -218,6 +223,7 @@ export interface CellHOCProps extends TableCellProps {
   cell: Cell;
   column: Column;
   announce: Announce;
+  isFlagEnabled: (flag: string) => boolean;
```
mhs30, line 226: Line 152 and 192 has this typed in this shape, might be worth keep the same way.

```suggestion
  isFlagEnabled(flag: string): boolean;
```

src/table/utils/handle-click.ts
```diff
@@ -28,3 +30,40 @@ export const handleMouseDownLabelToFocusHeadCell = (evt: MouseEvent, rootElement
   evt.preventDefault();
   updateFocus({ focusType: 'focus', cell: getCellElement(rootElement, [0, columnIndex]) });
 };
+
+export const getMouseHandlers = (
+  cell: Cell,
+  announce: Announce,
+  onClick: React.MouseEventHandler<HTMLTableCellElement> | undefined,
+  selectionDispatch: SelectionDispatch,
+  isFlagEnabled: (flag: string) => boolean
```
mhs30, line 39:
```suggestion
  isFlagEnabled(flag: string): boolean
```

src/table/utils/handle-key-press.ts
```diff
@@ -180,22 +180,25 @@ export const handleBodyKeyDown = ({
     bottom: isSelectionMode && totalsPosition === 'bottom' ? 1 : 0,
   };
 
+  const basicFeaturesEnabled = isFlagEnabled('PS_15585_SN_TABLE_BASIC_FEATURES');
```
mhs30, line 183: Does it make sense to propagate a feature flag to qlik-oss?

src/table/utils/handle-key-press.ts
```diff
@@ -180,22 +180,25 @@ export const handleBodyKeyDown = ({
...
+
   switch (evt.key) {
     case 'ArrowUp':
```
mhs30, line 186: Seems we are using this strings (arrow keys) in many places, might be worth creating an enum and have them only in one place.

src/table/utils/handle-key-press.ts
```diff
@@ -241,8 +244,14 @@ export const handleBodyKeyDown = ({
 /**
  * confirms selections when making multiple selections with shift + arrows and shit is released
  */
-export const handleBodyKeyUp = (evt: React.KeyboardEvent, selectionDispatch: React.Dispatch<TSelectionActions>) => {
-  evt.key === 'Shift' && selectionDispatch({ type: SelectionActions.SELECT_MULTI_VALUES });
+export const handleBodyKeyUp = (
+  evt: React.KeyboardEvent,
+  selectionDispatch: React.Dispatch<TSelectionActions>,
+  isFlagEnabled: (flag: string) => boolean
```
mhs30, line 250:
```suggestion
  isFlagEnabled(flag: string): boolean
```

src/table/utils/handle-key-press.ts
```diff
@@ -241,8 +244,14 @@ export const handleBodyKeyDown = ({
...
+) => {
+  isFlagEnabled('PS_15585_SN_TABLE_BASIC_FEATURES') &&
+    evt.key === 'Shift' &&
```
mhs30, line 253: Same here, maybe create an enum with the keys we are using.

src/table/utils/selections-utils.ts
```diff
@@ -144,35 +161,91 @@ export const getSelectedRows = (
...
+  evt: React.KeyboardEvent | React.MouseEvent,
+  cell: Cell,
+  firstCell?: Cell
+): Record<string, number> => {
+  const newSelectedRows = { ...selectedRows };
+
+  if ('key' in evt && evt.shiftKey && evt.key.includes('Arrow')) {
+    newSelectedRows[cell.qElemNumber] = cell.rowIdx;
+    // add the next or previous cell to selectedRows, based on which arrow is pressed
+    if (evt.key === 'ArrowDown') {
```
mhs30, line 176: Another suggestion might be of having an array with arrow key codes;

`const arrowKeyCodes = [KeyCodes.ARROW_DOWN,KeyCodes.ARROW_UP...etc]`

then in line 173 `arrowKeyCodes.includes(evt.key)` instead of `evt.key.includes('Arrow')`

```suggestion
    if (evt.key === KeyCodes.ARROW_DOWN) {
```

src/table/types.ts
haxxmaxx: agree, will change

src/table/utils/selections-utils.ts
haxxmaxx: ahh, yes so one enum with all used keycodes and then another array with arrow keys. I like it. Will do this in another PR right away

src/table/utils/handle-click.ts
```diff
@@ -28,3 +30,40 @@ export const handleMouseDownLabelToFocusHeadCell = (evt: MouseEvent, rootElement
...
+  const handleMouseDown = (evt: React.MouseEvent) => {
+    // Prevent focus to be set to the cell, done on mouse up instead
+    evt.preventDefault();
+    evt.stopPropagation();
+    if (selectMultiEnabled && cell.isSelectable)
+      selectionDispatch({ type: SelectionActions.SELECT_MULTI_START, payload: { cell } });
+  };
+
+  const handleMouseOver = (evt: React.MouseEvent) => {
+    if (selectMultiEnabled && evt.button === 0 && evt.buttons === 1)
```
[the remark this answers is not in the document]
haxxmaxx: ahh I thought that was the number of buttons pressed. I'll revisit this

src/table/utils/handle-click.ts
haxxmaxx: Is this really possible here?

src/table/Root.tsx
```diff
@@ -8,15 +8,16 @@ import TableWrapper from './components/TableWrapper';
...
 export function render(props: RenderProps, reactRoot?: ReactDom.Root) {
-  const { direction, selectionsAPI } = props;
+  const { direction, selectionsAPI, tableData } = props;
   const muiTheme = muiSetup(direction);
 
   reactRoot?.render(
     <StyleSheetManager stylisPlugins={direction === 'rtl' ? [rtlPluginSc] : undefined}>
       <ThemeProvider theme={muiTheme}>
-        <TableContextProvider selectionsAPI={selectionsAPI}>
+        <TableContextProvider selectionsAPI={selectionsAPI} tableRows={tableData?.rows as Row[]}>
```
haxxmaxx, line 20: We know that rows exist here, it just has to be optional on the type, to work with the native implementation

src/table/components/__tests__/withSelections.spec.tsx
```diff
@@ -46,33 +45,4 @@ describe('withSelections', () => {
 
     expect(queryByText(value)).toBeVisible();
   });
-
-  it('should call selectCell on mouseUp', () => {
```
haxxmaxx, line 50: these test are moved to handle-click

src/table/utils/handle-click.ts
haxxmaxx: pretty sure it is not

src/table/utils/selections-utils.ts
a-m-dev: that is fine, you **have to** iterate through selected ones and you are already using O(n) solution with accessing cell data with `allRows[idx][...]`
do you have access to first + last selected cell idx here? what is cell? 

I was thinking if you have start and end indexes, then you should not loop over all rows right? even the selection comes in different chunks (skip 10 rows, select 5 row, skip 15 rows, select 6 rows, ...)

a-m-dev (approved):
Great job, just minor comments and change the code if you see that comment makes sense, looks good otherwise!

src/__test__/generate-test-data.ts
```diff
@@ -55,3 +55,23 @@ export function generateLayout(
...
 }
+
+export const createCell = (rowIdx: number, colIdx = 0) =>
+  ({
+    qElemNumber: rowIdx,
+    rowIdx,
+    colIdx,
+    rawRowIdx: rowIdx,
+    rawColIdx: colIdx,
+  } as unknown as Cell);
```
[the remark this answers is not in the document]
haxxmaxx: you're right, will change

src/table/Root.tsx
haxxmaxx: Hmm, not sure what is best practice here. The thing is that we have two root files that, one for web implementation, one for react native (mobile app). Nebula has constraints on what some functions must be called etc in the index files. So we have to override the root file with root.native. They have to share types, so tableData has to be optional. BUT, it is always defined when THIS code runs. I think casting here indicates that we actually have a non-empty array, where `|| []` might be confusing to someone not knowing about this, since tableRows will never be `[]`

src/table/types.ts
```diff
@@ -29,17 +30,37 @@ export interface ActionPayload {
...
+  | SelectAction
+  | SelectMultiStartAction
+  | SelectMultiAddAction
+  | SelectMultiEndAction
+  | ResetAction
+  | ClearAction
+  | UpdateAllRowsAction;
 
-export type SelectionActionTypes = SelectAction | ResetAction | ClearAction | SelectMultiValuesAction;
+export type SelectionDispatch = React.Dispatch<SelectionActionTypes> | jest.Mock<any, any>;
```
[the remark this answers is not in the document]
haxxmaxx: what do you mean problem? I am just creating a `SelectionDispatch` type so I don't have to have `React.Dispatch<SelectionActionTypes> | jest.Mock<any, any>`everywhere?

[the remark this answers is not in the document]
haxxmaxx: or do you mean I don't need the `| jest.Mock<any, any>` part?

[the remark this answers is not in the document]
a-m-dev: yes, i was meant that is `| jest.Mock<any, any>` part causing some issues in types maybe?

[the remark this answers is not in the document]
haxxmaxx: ah ok, I guess I thought that was needed. I can try to remove it

src/table/components/__tests__/withSelections.spec.tsx
```diff
@@ -13,16 +12,16 @@ describe('withSelections', () => {
...
   let announce: Announce;
   let column: Column;
   let selectionDispatchMock: jest.Mock<any, any>;
+  let isFlagEnabled: (flag: string) => boolean;
 
   const renderWithSelections = () =>
     render(
-      <TableContextProvider selectionsAPI={selectionsAPI} selectionDispatchMock={selectionDispatchMock}>
-        <HOC cell={cell} styling={styling} announce={announce} column={column}>
+      <TableContextProvider selectionsAPI={selectionsAPI} tableRows={[]} selectionDispatchMock={selectionDispatchMock}>
```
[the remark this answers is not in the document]
haxxmaxx: ahh that is nice. that removes the need to cast in the root file as well :)

PR #690, opened 2022-09-14
refactor: copy cell value via keyboard

Donya-qlik:


src/table/types.ts
```diff
@@ -94,7 +94,7 @@ export interface HandleBodyKeyDownProps {
   totalsPosition: TotalsPosition;
   paginationNeeded: boolean;
   selectionsAPI: ExtendedSelectionAPI;
-  isFlagEnabled?(flag: string): boolean;
+  isFlagEnabled(flag: string): boolean;
```
Donya-qlik, line 97: Removing the option(?), causes other upper level props complain about the type, like `TableBodyWrapperProps` since it has been set optional from the very beginning. How can I cast it if it is possible ?

haxxmaxx: The only place it should be optional is `RenderProps`. So you need to change tableWrapper and tableBodyWrapper as well

haxxmaxx (approved):

PR #692, opened 2022-09-18
feat: add copyCellValue as an item to context menu 

Donya-qlik:


src/index.ts
```diff
@@ -82,6 +86,20 @@ export default function supernova(env: Galaxy) {
...
+          event.target &&
+          menu.addItem({
+            translation: 'contextMenu.copyCellValue',
+            icon: 'lui-icon lui-icon--copy',
+            tid: 'copy-cell-context-item',
+            select: async () => {
+              copyCellValue(event.target.firstChild.textContent);
+            },
+          });
+        return Promise.resolve(menu);
```
haxxmaxx, line 100: doesn't this make the function async

src/index.ts
```diff
@@ -82,6 +86,20 @@ export default function supernova(env: Galaxy) {
       const announce = useAnnounceAndTranslations(rootElement, translator);
       const changeSortOrder = useSorting(model);
 
+      onContextMenu((menu: Menu, event: any) => {
```
haxxmaxx, line 89: We should scope out the content of this function to a separate file

src/index.ts
Donya-qlik: Well, I think(if I'm not wrong) it is not async here because I'm not calling another function after returning the Promise, but then the main function in sense-client that I'm calling `onContextMenu ` in it is an async function `async function getGridCellContextMenu(...)`, does it make sense ?

src/index.ts
```diff
@@ -58,8 +59,9 @@ const renderWithCarbon = ({
 
 export default function supernova(env: Galaxy) {
   const {
-    flags: { isEnabled: isFlagEnabled },
+    flags: { isEnabled },
   } = env;
+  const areBasicFeaturesEnabled = isEnabled('PS_15585_SN_TABLE_BASIC_FEATURES');
```
haxxmaxx, line 64: No real point of desctructuring this since is only used once. I would move this to inside `component` function.
```suggestion
const areBasicFeaturesEnabled = env.flags.isEnabled('PS_15585_SN_TABLE_BASIC_FEATURES');
```

src/nebula-hooks/use-context-menu.ts
```diff
@@ -0,0 +1,18 @@
+import { onContextMenu } from '@nebula.js/stardust';
+import { copyCellValue } from '../table/utils/accessibility-utils';
+import { Menu } from '../types';
+
+export default function useContextMenu(areBasicFeaturesEnabled: boolean) {
```
haxxmaxx, line 5: very nice idea to put this in a hook

src/table/utils/handle-key-press.ts
```diff
@@ -231,7 +231,9 @@ export const handleBodyKeyDown = ({
       focusSelectionToolbar(evt.target as HTMLElement, keyboard, evt.shiftKey);
       break;
     case 'c':
-      isFlagEnabled('PS_15585_SN_TABLE_BASIC_FEATURES') && (evt.ctrlKey || evt.metaKey) && copyCellValue(cell);
+      areBasicFeaturesEnabled &&
+        (evt.ctrlKey || evt.metaKey) &&
+        copyCellValue((evt.target as HTMLElement).textContent as string);
```
haxxmaxx, line 236: we might just want handle all the this casting and sanity checking inside `copyCellValue`. Let's keep it like this for now

haxxmaxx (approved):

src/nebula-hooks/use-context-menu.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore  TODO: remove tag when onContextMenu is declared in nebula/stardust
```
Donya-qlik, line 1: Are you agree with this ? We will remove it soon after fixing the sn-table version

src/table/utils/handle-click.ts
```diff
@@ -39,9 +39,9 @@ export const getSelectionMouseHandlers = (
   announce: Announce,
   onMouseDown: React.MouseEventHandler<HTMLTableCellElement> | undefined,
   selectionDispatch: SelectionDispatch,
-  isFlagEnabled: (flag: string) => boolean
+  areBasicFeaturesEnabled: boolean
 ) => {
-  const selectMultiEnabled = isFlagEnabled('PS_15585_SN_TABLE_BASIC_FEATURES');
+  const selectMultiEnabled = areBasicFeaturesEnabled;
```
haxxmaxx, line 44: might as well use the `areBasicFeaturesEnabled` in the code

Donya-qlik: Ah, so silly! yes off course, I'll fix it.

src/nebula-hooks/use-context-menu.ts
haxxmaxx: It looks fine. What do you mean by fixing the sn-table version? I think it will be like this until onContext is add `onContextMenu` tot the nebula api

PR #693, opened 2022-09-19
feat: extend copyCellValue via keyDown to Heads and Total cells

Donya-qlik:
CopyCellValue via keyDown ( `ctrl + c ` and `meta + c` ) is functioning now on Head and Total cells as well as body cells.

### DIFF
--- a/src/nebula-hooks/use-context-menu.ts
+++ b/src/nebula-hooks/use-context-menu.ts
@@ -1,10 +1,9 @@
 // @ts-ignore  TODO: remove tag when onContextMenu is declared in nebula/stardust
 import { onContextMenu } from '@nebula.js/stardust';
 import { copyCellValue } from '../table/utils/accessibility-utils';
-import { Menu } from '../types';
 
 export default function useContextMenu(areBasicFeaturesEnabled: boolean) {
-  onContextMenu?.((menu: Menu, event: any) => {
+  onContextMenu((menu: any, event: any) => {
     areBasicFeaturesEnabled &&
       event.target &&
       menu.addItem({
--- a/src/nebula-hooks/use-context-menu.ts
+++ b/src/nebula-hooks/use-context-menu.ts
@@ -11,7 +11,7 @@ export default function useContextMenu(areBasicFeaturesEnabled: boolean) {
         icon: 'lui-icon lui-icon--copy',
         tid: 'copy-cell-context-item',
         select: async () => {
-          copyCellValue(event.target.firstChild.textContent);
+          copyCellValue(event);
         },
       });
   });
--- a/src/table/utils/__tests__/handle-key-press.spec.ts
+++ b/src/table/utils/__tests__/handle-key-press.spec.ts
@@ -23,6 +23,10 @@ describe('handle-key-press', () => {
     areBasicFeaturesEnabled = true;
   });
 
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
   describe('shouldBubble', () => {
     let evt: React.KeyboardEvent;
     let isSelectionMode: boolean;
@@ -41,6 +45,8 @@ describe('handle-key-press', () => {
       isSelectionMode = false;
       keyboardEnabled = false;
       paginationNeeded = true;
+      areBasicFeaturesEnabled = true;
+      jest.spyOn(handleAccessibility, 'copyCellValue');
     });
 
     it('should return true when esc is pressed and isSelectionMode is false', () => {
@@ -274,6 +280,7 @@ describe('handle-key-press', () => {
         layout,
         isInteractionEnabled,
         setFocusedCellCoord,
+        areBasicFeaturesEnabled,
       });
 
     beforeEach(() => {
@@ -287,6 +294,13 @@ describe('handle-key-press', () => {
         target: {
           blur: jest.fn(),
           setAttribute: jest.fn(),
+          children: [
+            {
+              firstChild: {
+                textContent: '',
+              },
+            },
+          ],
         } as unknown as HTMLElement,
       } as unknown as React.KeyboardEvent;
       rootElement = {
@@ -299,6 +313,7 @@ describe('handle-key-press', () => {
       changeSortOrder = jest.fn();
       isInteractionEnabled = true;
       setFocusedCellCoord = jest.fn();
+      jest.spyOn(handleAccessibility, 'copyCellValue');
     });
 
     it('when press arrow down key on head cell, should prevent default behavior, remove current focus and set focus and attribute to the next cell', () => {
@@ -345,6 +360,28 @@ describe('handle-key-press', () => {
       expect(changeSortOrder).not.toHaveBeenCalled();
       expect(setFocusedCellCoord).not.toHaveBeenCalled();
     });
+
+    it('should call copyCellValue on Head cell when the pressed keys are Ctrl and C keys', () => {
+      evt.key = KeyCodes.C;
+      evt.ctrlKey = true;
+      callHandleHeadKeyDown();
+      expect(handleAccessibility.copyCellValue).toHaveBeenCalled();
+    });
+
+    it('should call copyCellValue on Head cell when the pressed keys are Meta and C keys', () => {
+      evt.key = KeyCodes.C;
+      evt.metaKey = true;
+      callHandleHeadKeyDown();
+      expect(handleAccessibility.copyCellValue).toHaveBeenCalled();
+    });
+
+    it('should not call copyCellValue on Head cell when the flag is disabled', () => {
+      evt.key = KeyCodes.C;
+      evt.metaKey = true;
+      areBasicFeaturesEnabled = false;
+      callHandleHeadKeyDown();
+      expect(handleAccessibility.copyCellValue).not.toHaveBeenCalled();
+    });
   });
 
   describe('handleTotalKeyDown', () => {
@@ -372,10 +409,11 @@ describe('handle-key-press', () => {
       } as unknown as HTMLElement;
       setFocusedCellCoord = jest.fn();
       isSelectionMode = false;
+      jest.spyOn(handleAccessibility, 'copyCellValue');
     });
 
     it('should move the focus from the current cell to the next when arrow key down is pressed on a total cell', () => {
-      handleTotalKeyDown(evt, rootElement, cellCoord, setFocusedCellCoord, isSelectionMode);
+      handleTotalKeyDown(evt, rootElement, cellCoord, setFocusedCellCoord, isSelectionMode, areBasicFeaturesEnabled);
       expect(evt.preventDefault).toHaveBeenCalledTimes(1);
       expect(evt.stopPropagation).toHaveBeenCalledTimes(1);
       expect((evt.target as HTMLElement).setAttribute).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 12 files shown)


src/table/utils/accessibility-utils.ts
```diff
@@ -211,9 +211,9 @@ export const announceSelectionState = (
   }
 };
 
-export const copyCellValue = async (cell: Cell) => {
+export const copyCellValue = async (value?: string) => {
```
Donya-qlik, line 214: I found it more practical to use `cell value` instead of cell with cell properties. It makes it much more easier when the function is used for Totals and Head.

src/index.ts
```diff
@@ -58,8 +58,9 @@ const renderWithCarbon = ({
 
 export default function supernova(env: Galaxy) {
   const {
-    flags: { isEnabled: isFlagEnabled },
+    flags: { isEnabled },
   } = env;
+  const isFlagEnabled = isEnabled('PS_15585_SN_TABLE_BASIC_FEATURES');
```
haxxmaxx, line 63: This name is too generic. It doesn't say what flag is enabled. Either keep it as before or rename to `areBasicFeaturesEnabled`

src/table/utils/handle-key-press.ts
```diff
@@ -100,6 +102,7 @@ export const handleHeadKeyDown = ({
   if (shouldBubble(evt)) return;
   preventDefaultBehavior(evt);
 
+  const cellValue = column.label;
```
haxxmaxx, line 105: this is only used in one case, no need to define this outside the switch

src/table/utils/handle-key-press.ts
```diff
@@ -180,6 +192,7 @@ export const handleBodyKeyDown = ({
     bottom: isSelectionMode && totalsPosition === 'bottom' ? 1 : 0,
   };
 
+  const cellValue = cell.qText;
```
haxxmaxx, line 195: same here

src/table/utils/handle-key-press.ts
```diff
@@ -124,7 +130,9 @@ export const handleTotalKeyDown = (
   rootElement: HTMLElement,
   cellCoord: [number, number],
   setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>,
-  isSelectionMode: boolean
+  isSelectionMode: boolean,
+  column: Column,
```
haxxmaxx, line 134: We should try to reduce arguments here. So let's use the text on event.target instead. Same for header

src/table/utils/handle-key-press.ts
```diff
@@ -133,6 +141,7 @@ export const handleTotalKeyDown = (
   if (shouldBubble(evt)) return;
   preventDefaultBehavior(evt);
 
+  const cellValue = column.totalInfo;
```
haxxmaxx, line 144: same here, no need for it to be outside the switch

src/table/utils/handle-key-press.ts
```diff
@@ -111,6 +114,9 @@ export const handleHeadKeyDown = ({
     case 'Enter':
       changeSortOrder(layout, column);
       break;
+    case 'c':
+      isFlagEnabled && copyCellValue(cellValue);
```
haxxmaxx, line 118: check for ctrl/meta here?

src/table/utils/accessibility-utils.ts
haxxmaxx, line 214: why is value optional? When defined like this, the call `copyCellValue()` would be valid, but then here is no value to be copied. If you change to use the text from the event it might always be there?

src/table/utils/handle-key-press.ts
haxxmaxx: Your tests are still passing, so wouldn't hurt to add a test to see that it is not called when ctrl/meta is not pressed.

src/index.ts
haxxmaxx: the flag name is not perfect to begin with, since we are introducing multiple unrelated features. But just `isFlagEnabled` wouldn't say anything to someone seeing the code for the first time

src/table/utils/handle-key-press.ts
```diff
@@ -231,7 +245,10 @@ export const handleBodyKeyDown = ({
       focusSelectionToolbar(evt.target as HTMLElement, keyboard, evt.shiftKey);
       break;
     case 'c':
-      isFlagEnabled('PS_15585_SN_TABLE_BASIC_FEATURES') && (evt.ctrlKey || evt.metaKey) && copyCellValue(cell);
+      {
+        const cellValue = cell.qText;
+        areBasicFeaturesEnabled && (evt.ctrlKey || evt.metaKey) && copyCellValue(cellValue);
```
haxxmaxx, line 250: Sorry I meant this
```suggestion
        areBasicFeaturesEnabled && (evt.ctrlKey || evt.metaKey) && copyCellValue(cell.qText;);
```
cellValue is only used once, no need to create a seperate variable.

src/nebula-hooks/use-context-menu.ts
```diff
@@ -1,10 +1,9 @@
 // @ts-ignore  TODO: remove tag when onContextMenu is declared in nebula/stardust
 import { onContextMenu } from '@nebula.js/stardust';
 import { copyCellValue } from '../table/utils/accessibility-utils';
-import { Menu } from '../types';
 
 export default function useContextMenu(areBasicFeaturesEnabled: boolean) {
-  onContextMenu((menu: Menu, event: any) => {
+  onContextMenu((menu: any, event: any) => {
```
Donya-qlik, line 6: As we decided to set menu type to `any` I should change this which was missed in the original PR, It's a small change so I fixed through this PR.

src/table/utils/handle-key-press.ts
```diff
@@ -128,6 +129,12 @@ export const handleHeadKeyDown = ({
       // Space bar / Enter: update the sorting
       changeSortOrder(layout, column);
       break;
+    case KeyCodes.C: {
+      areBasicFeaturesEnabled &&
+        (evt.ctrlKey || evt.metaKey) &&
+        copyCellValue((evt.target as HTMLElement).children[0].firstChild?.textContent as string);
```
Donya-qlik, line 135: FirstChild is optional here but still the best option we have to get the textContent for Head cells without being disturbed by SortLabel span content.

src/nebula-hooks/use-context-menu.ts
haxxmaxx: sure :)

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -23,6 +23,10 @@ describe('handle-key-press', () => {
     areBasicFeaturesEnabled = true;
   });
 
+  afterEach(() => {
```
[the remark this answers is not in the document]
Donya-qlik: Yes, It is already on top level, I've removed the others :)

src/table/utils/accessibility-utils.ts
```diff
@@ -216,9 +216,12 @@ export const announceSelectionState = (
   }
 };
 
-export const copyCellValue = async (value: string) => {
+export const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
+  const value = isHeadCell
```
cbt1, line 220: This looks like a potential bugg. As isHeadCell can be `undefined`, when `String(value)` is called. A `"undefined"` string will be copied.

I think should should just not attempt to copy the `value` if `value` is not a string.

src/table/utils/accessibility-utils.ts
```diff
@@ -216,9 +216,14 @@ export const announceSelectionState = (
   }
 };
 
-export const copyCellValue = async (value: string) => {
+export const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
+  const value =
+    isHeadCell && (evt.target as HTMLElement).children[0].firstChild
```
cbt1, line 221: It do seem a bit clunky and hard ta read when you do the type casting 4 times for the same property. Why not just store the casted value in a variable? Ex: `const target = evt.target as HTMLElement`. If linting complains I'd probably just disable it for this line (not sure if you can to type casting when you do object destructing..)

cbt1 (approved):
Some minor comment which you can opt-in to fix for readablity. Otherwise LTGM.

src/nebula-hooks/use-context-menu.ts
```diff
@@ -1,18 +1,17 @@
 // @ts-ignore  TODO: remove tag when onContextMenu is declared in nebula/stardust
 import { onContextMenu } from '@nebula.js/stardust';
 import { copyCellValue } from '../table/utils/accessibility-utils';
-import { Menu } from '../types';
 
 export default function useContextMenu(areBasicFeaturesEnabled: boolean) {
-  onContextMenu?.((menu: Menu, event: any) => {
+  onContextMenu((menu: any, event: any) => {
```
haxxmaxx, line 6: put the question mark back

src/table/utils/accessibility-utils.ts
```diff
@@ -216,9 +216,14 @@ export const announceSelectionState = (
   }
 };
 
-export const copyCellValue = async (value: string) => {
+export const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
+  const target = evt.target as HTMLElement;
+  const value =
+    isHeadCell && target.children[0].firstChild
```
haxxmaxx, line 222: this bool will be false when using the context menu to copy a head cell. Do you really need it? I would guess that target has no children for totals and body cells, maybe that is enough to know what `firstChild` to copy

src/table/utils/accessibility-utils.ts
```diff
@@ -216,9 +216,14 @@ export const announceSelectionState = (
...
+      ? target.children[0].firstChild?.textContent
```
haxxmaxx, line 223: question mark not needed here, you check that in the line above

src/table/utils/accessibility-utils.ts
```diff
@@ -216,9 +216,14 @@ export const announceSelectionState = (
...
+      : target.firstChild && target.firstChild?.textContent;
```
haxxmaxx, line 224: doing the same thing twice here, `target.firstChild?.textContent` is enough

src/table/utils/accessibility-utils.ts
```diff
@@ -216,9 +216,14 @@ export const announceSelectionState = (
   }
 };
 
-export const copyCellValue = async (value: string) => {
+export const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
```
haxxmaxx, line 219: I think we can give this its own file. It doesn't have anything to do with accessibility and accessibility-utils is already pretty big

haxxmaxx (commented):
I think this needs some cleaning up. Perfect to do in a refactor PR next

Donya-qlik:
> I think this needs some cleaning up. Perfect to do in a refactor PR next

Sure! I'll create a new one

src/nebula-hooks/use-context-menu.ts
haxxmaxx: As a safeguard if someone does not have the latest nebula version. The table crashes and they get a cryptic error. Better to just skip trying to call `onContextMenu` and the table still runs.

PR #700, opened 2022-09-24
docs: add git hooks

LiKang6688:


package.json
```diff
@@ -56,7 +56,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
-      "yarn lint"
+      "yarn lint && yarn types:check"
```
haxxmaxx, line 59: I thought about this and I am not sure we want to do this. For example, you can changes some types for something and want to commit it. But then the types for the tests could be wrong, and you would have to fix that first

LiKang6688: Yes, you can commit when types are wrong as I document:
    git commit -m "commit message" --no-verify
You need to run `yarn types:check` anyway before committing, so you know that types are wrong earlier and not from CI.

haxxmaxx: ahh yes of course, that works

haxxmaxx (approved):

PR #701, opened 2022-09-26
fix: properly focus cell on click

haxxmaxx:
When clicking a cell to focus it, you remove the tab-index for the one the cell that has a tab stop already. When keyboard.focus does not run, the tab-index is still -1 for the new focused cell. Need to make sure it is added manually.


Donya-qlik (approved):
lgtm : )

PR #702, opened 2022-09-27
refactor: keep constants in seperate file

haxxmaxx:
- move enums and other constant to constants.ts
- move selection types to types file to avoid circular deps
- create `KeyCodes` enum for key codes
- add more values to `StylingDefaults`

### DIFF
--- a/src/table/utils/accessibility-utils.ts
+++ b/src/table/utils/accessibility-utils.ts
@@ -174,7 +174,7 @@ export const handleFocusoutEvent = (
   keyboard: stardust.Keyboard
 ) => {
   const targetElement = evt.currentTarget as HTMLDivElement;
-  if (keyboard?.enabled && !targetElement.contains(evt.relatedTarget as Node) && !shouldRefocus.current) {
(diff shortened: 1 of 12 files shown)


src/table/utils/accessibility-utils.ts
```diff
@@ -174,7 +174,7 @@ export const handleFocusoutEvent = (
   keyboard: stardust.Keyboard
 ) => {
   const targetElement = evt.currentTarget as HTMLDivElement;
-  if (keyboard?.enabled && !targetElement.contains(evt.relatedTarget as Node) && !shouldRefocus.current) {
+  if (keyboard.enabled && !targetElement.contains(evt.relatedTarget as Node) && !shouldRefocus.current) {
```
haxxmaxx, line 177: keyboard always exists

src/table/utils/selections-utils.ts
```diff
@@ -129,7 +100,7 @@ export const getSelectedRows = (
   if (evt.shiftKey && key.includes('Arrow')) {
```
haxxmaxx, line 100: Will most likely keep the Arrow string here. Creating a sperate array to test this in one place seems overkill to me

mhs30 (approved):
LGTM

PR #707, opened 2022-09-30
Removed subtitle from sn-table definition

rajani-varpe:
Removed subtitle from sn-table definition


haxxmaxx:
where is the added translation? If it is added in sense-client, no need to add it to the title of this PR

rajani-varpe:
> 

Yes translations are added in sense, here the subtitle field is removed.

haxxmaxx (approved):

PR #708, opened 2022-09-30
refactor: remove prev/next qElemNumber

haxxmaxx:
Since we now store `allRows` on the selection state, no need to store qElemNumber for any neighboring cells on the cell. Replacing with a last row bool, to keep track of when it is bossible to select by shift + arrow down. also:
- use KeyCodes in test file
- breaking out `shouldSelectMultiValues` for readability in `handleBodyKeyPress`


cbt1 (approved):
LGTM

PR #710, opened 2022-10-03
fix: call api.select on mouseup outside table body

haxxmaxx:
So when dragging to select, you are suppose to be able to let go (mouseup) outside the table. The only way I can figure out how to do this is to add a listener temporarily to the document. When using add/removeEventListener, the callback must be the same, in order for it to be removed correctly. And the callback itself needs the latest state. The easiest way do that, and possibly only way, is do do a dispatch.

That means on mousedown we pass callback that runs dispatch to the reducer, then store that callback on the state, then run remove it on mouseup.

Also changed so all selects are done on mouse up. If you haven't moved the mouse outside the first cell, we do a normal select. Otherwise we add the cells you dragged over. Solves the bug @cbt1 mentions below.


src/index.ts
```diff
@@ -57,9 +57,10 @@ const renderWithCarbon = ({
 };
 
 export default function supernova(env: Galaxy) {
-  const {
-    flags: { isEnabled: isFlagEnabled },
-  } = env;
+  // const {
```
haxxmaxx, line 60: for dev reason, will remove

cbt1:
Did some testing on this, seem to work fine to me. Only found one "quirk" and that is when you do mouse down on a cell, move to select more cells and then move back to the starting cell to do mouse up. The starting cell will be de-selected.

haxxmaxx:
> Did some testing on this, seem to work fine to me. Only found one "quirk" and that is when you do mouse down on a cell, move to select more cells and then move back to the starting cell to do mouse up. The starting cell will be de-selected.

That one we found earlier as well, it is there before this PR. I'll do that in the next PR. might look at switching to mousemove as well

haxxmaxx:
@cbt1 actually solved that in here as well

haxxmaxx:
> > Did some testing on this, seem to work fine to me. Only found one "quirk" and that is when you do mouse down on a cell, move to select more cells and then move back to the starting cell to do mouse up. The starting cell will be de-selected.
> 
> That one we found earlier as well, it is there before this PR. I'll do that in the next PR. might look at switching to mousemove as well

Decided to fix that bug in here as well

src/table/utils/__tests__/selections-utils.spec.ts
```diff
@@ -200,27 +201,42 @@ describe('selections-utils', () => {
 
       afterEach(() => jest.clearAllMocks());
 
+      it('should return state unchanged, when type is selectMultiStart and mouseupOutsideCallback is undefined', () => {
+        const action = { type: SelectionActions.SELECT_MOUSE_DOWN, payload: { cell } } as SelectMouseDownAction;
+        const newState = reducer(state, action);
+        expect(newState).toEqual(state);
```
cbt1, line 207: Are you expected the same object reference? Such that `newState === state`? Then you should be using `.toBe()` method instead.

haxxmaxx: good question, I don't remember how `useReducer` does that. I will try with toBe, that will give us the answer :)

haxxmaxx:
need to finish up some tests since I added some more code

cbt1 (approved):
LGTM

PR #711, opened 2022-10-03
feat: add react-native example

vcellu:


src/nebula-hooks/use-extended-theme.native.js
```diff
@@ -0,0 +1,29 @@
+import { useTheme, useState, useEffect } from '@nebula.js/stardust';
+
+const useExtendedTheme = () => {
+  const nebulaTheme = useTheme();
+  const [theme, setTheme] = useState(nebulaTheme);
+
+  useEffect(() => {
+    if (nebulaTheme) {
+      const themeProxy = new Proxy(nebulaTheme, {
```
haxxmaxx, line 9: @Caele So for the native implementation the index is shift for `getColorPickerColor`. I remember we saw something like this in the client when it was picking the color palette from the default nebula theme and not the sense theme or something?

tools/build.js
```diff
@@ -51,9 +51,15 @@ const configureReactNative = () => {
   // cleanup old build
   const reactNativeFolder = './react-native';
   fs.removeSync(path.resolve(process.cwd(), `${reactNativeFolder}/dist`));
-  rnSnTablePackage.version = snTablePackage.version;
+  const nextVersion = rnSnTablePackage.version.split('.');
+  const minor = parseInt(nextVersion[nextVersion.length - 1], 10) + 1;
+  nextVersion[nextVersion.length - 1] = minor;
+  rnSnTablePackage.version = nextVersion.join('.');
```
haxxmaxx, line 57: So this bumps the minor version every time you build it for react native? How would that work when you want to make a release? it is always a minor bump?

src/table/native/Table.jsx
```diff
@@ -105,6 +118,9 @@ const Table = ({ layout, model, manageData, selectionsAPI, changeSortOrder, app,
         onColumnsResized={onColumnsResized}
         clearSelections={clearSelections}
         onHeaderPressed={onHeaderPressed}
+        freezeFirstColumn={freezeFirstColumn(layout)}
```
haxxmaxx, line 121: Easier to just do
```suggestion
        freezeFirstColumn={freezeFirstColumn(layout.scrolling?.keepFirstColumnInView)}
```

than to create a file that you have to type the props for?

tools/build.js
vcellu: We're still working on a better build pipeline.  I would like to use semantic release.

src/nebula-hooks/use-extended-theme.native.js
vcellu: Let's resolve this in another PR.

haxxmaxx: sure

tools/build.js
haxxmaxx: ok, if this works for you now it is fine with me

react-native/package.json
```diff
@@ -24,8 +24,8 @@
   "main": "dist/sn-table.js",
   "peerDependencies": {
     "react": "18.2.0",
-    "react-native": "0.70.1",
-    "@nebula.js/stardust": "2.x"
+    "@nebula.js/stardust": "3.x",
```
haxxmaxx, line 27: Sorry I was unclear, this needs to be `>=3.0.3` since we added `onContextMenu` to nebula in this version

react-native/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nebula.js/react-native-sn-table",
-  "version": "1.6.0",
+  "version": "1.14.4",
```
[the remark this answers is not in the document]
vcellu: We haven't built the package in a while.  It should synch up on next build

README.md
```diff
@@ -70,6 +70,30 @@ Compress the generated 'sn-table-ext' folder into the 'application/zip' file for
...
+2. A QCS Account with an API Key
+3. An app with a table.
+
+### Running
+
+** Please Note this can take a while to build if it's the first time **
+
+1. `yarn`
+2. `yarn setup:rn:dev`
+3. `yarn ios`
```
haxxmaxx, line 95: trying to follow these instructions. There is no setup:rn:dev command, is it called build:rn now?

README.md
```diff
@@ -70,6 +70,30 @@ Compress the generated 'sn-table-ext' folder into the 'application/zip' file for
...
+
+- [Helium](https://github.com/qlik-oss/react-native-helium) -> A high performnce rendering engine that uses a GPU backed Skia implementation
+- [Carbon](https://github.com/qlik-oss/react-native-carbon) -> The business logic behind rendering Supernova's in React Native
+- [react-native-simple-grid](https://github.com/qlik-oss/react-native-simple-grid) -> The native sn-table
+
+To run the example app in React-Native;
+
+### Pre-reqs
+
+1. Install [react-native dev environment](https://reactnative.dev/docs/0.68/environment-setup). This includes Xcode and Android Studio.
```
haxxmaxx, line 85: what exactly should you install here, there are two different instructions: Expo Go and Native CLI. I am assuming the latter?

haxxmaxx (commented):
Code looks good, now the instructions just need to be correct :)

PR #713, opened 2022-10-04
chore: peer dependencies

haxxmaxx:


package.json
```diff
@@ -125,7 +125,7 @@
     "yargs": "17.5.1"
   },
   "peerDependencies": {
-    "@nebula.js/stardust": "3.x"
+    "@nebula.js/stardust": ">=3.0.3"
```
Donya-qlik, line 128: so it will not support 3.0.2 anymore? seems the previous one could do.

Donya-qlik: Ah, ok, because the mashup issue, you said before.

haxxmaxx: yes :)

Caele (approved):

PR #715, opened 2022-10-05
fix: block arrow keys from propagating

haxxmaxx:
To reduce risk of arrow key events being picked up at higher levels. This is sometimes used to navigate between charts on a "sheet". Without stopping the propagation, this could potentially make you lose focus when you intended to stay inside the table. pagination specifically since it has no listener for arrows


cbt1 (approved):
LGTM

PR #716, opened 2022-10-11
refactor: move copyCellValue to a new file

Donya-qlik:


src/table/utils/copy-utils.ts
```diff
@@ -0,0 +1,14 @@
+const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
+  const target = evt.target as HTMLElement;
+  const value =
+    isHeadCell && target.children[0].firstChild
+      ? target.children[0].firstChild.textContent
+      : target.firstChild && target.firstChild.textContent;
```
haxxmaxx, line 6: you could use optional chaining here

src/table/utils/copy-utils.ts
```diff
@@ -0,0 +1,14 @@
+const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
+  const target = evt.target as HTMLElement;
+  const value =
+    isHeadCell && target.children[0].firstChild
```
haxxmaxx, line 4: I tested this a bit. Since `isHeadCell` is false when you copy using the context menu, you can get the wrong value. If it is possible, better to rely on if `target.children[0].firstChild` exists or not

src/table/utils/copy-utils.ts
```diff
@@ -0,0 +1,14 @@
+const copyCellValue = async (evt: any, isHeadCell: boolean = false) => {
```
haxxmaxx, line 1: I just realized we could follow the same pattern as other nebula hooks, so you can put this function in the use-context-menu file

src/table/utils/copy-utils.ts
Donya-qlik: What do you mean.., what you see is the checkup expression. As I tested it last time before merging everything looked work properly, what error type you get? 
It basically takes the value depends on the cell position, if it is head cell and `target.children[0].firstChild` exists it take `target.children[0].firstChild.textContent` otherwise it checks `target.firstChild` availability and takes the `target.firstChild.textContent` otherwise it will not copying it since I have value check at the beginning of the try/catch.

src/table/utils/copy-utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  const target = evt.target as HTMLElement;
+  const value =
+    isHeadCell && target.children[0].firstChild
+      ? target.children[0].firstChild.textContent
+      : target.firstChild && target.firstChild.textContent;
+  try {
+    value && (await navigator.clipboard.writeText(String(value)));
```
Donya-qlik, line 8: I'm checking the falsy values here

src/table/utils/copy-utils.ts
Donya-qlik: ok I'll move it

src/table/utils/copy-utils.ts
haxxmaxx: `target.firstChild && target.firstChild.textContent` will always return the same thing as `target.firstChild?.textContent`, it is just a shorter way of writing it

haxxmaxx (approved):

PR #721, opened 2022-10-13
fix: correct index usage for select multiple

haxxmaxx:
Was using the wrong column index. Was really confused by the which one is which, so I renamed them
- rawCol/RowIdx -> pageCol/RowIdx. raw never made sense to me since raw could refer to the raw data, meaning the entire data set. page refers dataPage, so row 0 is the first row that we are actually showing atm
- allRows -> pageRows. Same terminology, it is all rows on the dataPage

### DIFF
--- a/src/table/components/TableBodyWrapper.tsx
+++ b/src/table/components/TableBodyWrapper.tsx
@@ -62,7 +62,7 @@ function TableBodyWrapper({
       areBasicFeaturesEnabled={areBasicFeaturesEnabled}
     />
   );
-
+  console.log('body rows', rows);
   return (
     <StyledTableBody paginationNeeded={paginationNeeded} bodyCellStyle={bodyCellStyle}>
       {totalsPosition === 'top' ? totals : undefined}
--- a/src/table/components/TableBodyWrapper.tsx
+++ b/src/table/components/TableBodyWrapper.tsx
@@ -62,7 +62,7 @@ function TableBodyWrapper({
       areBasicFeaturesEnabled={areBasicFeaturesEnabled}
     />
   );
-  console.log('body rows', rows);
+
   return (
     <StyledTableBody paginationNeeded={paginationNeeded} bodyCellStyle={bodyCellStyle}>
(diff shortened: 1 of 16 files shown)


src/table/utils/selections-utils.ts
```diff
@@ -172,17 +172,17 @@ export const getMultiSelectedRows = (
...
   if (!firstCell) return selectedRows;
-  const lowestCellIdx = Math.min(firstCell.rawRowIdx, cell.rawRowIdx);
-  const highestCellIdx = Math.max(firstCell.rawRowIdx, cell.rawRowIdx);
+  const lowestRowIdx = Math.min(firstCell.pageRowIdx, cell.pageRowIdx);
+  const highestRowIdx = Math.max(firstCell.pageRowIdx, cell.pageRowIdx);
 
-  for (let idx = lowestCellIdx; idx <= highestCellIdx; idx++) {
-    const selectedCell = allRows[idx][`col-${firstCell.rawColIdx}`] as Cell;
+  for (let idx = lowestRowIdx; idx <= highestRowIdx; idx++) {
+    const selectedCell = pageRows[idx][`col-${firstCell.colIdx}`] as Cell;
```
haxxmaxx, line 185: this is the only change to fix the bug

src/table/components/TableBodyWrapper.tsx
```diff
@@ -76,6 +76,8 @@ function TableBodyWrapper({
         >
           {columns.map((column, columnIndex) => {
             const { id, align } = column;
+            // Note that rows are not necessarily ordered in qColumnOrder.
+            // So for each row, the cells are mapped according to column.id
```
haxxmaxx, line 80: Not sure we want it this way. I am considering changing to the `rows` array being ordered in the way we show things. And then just keep track of the way they are stored in the back-end with the index stored on the cell. I think this is left from your prototype @Caele. If so, do you remember if there is any particular reason to keep columns in the current order?

Donya-qlik (approved):
lgtm

PR #722, opened 2022-10-14
fix: copyCellValue on head cell when context menu is used

Donya-qlik:
copyCell value can be copied properly on head cells via contextMenu

### DIFF
--- a/src/table/utils/__tests__/copy-utils.spec.ts
+++ b/src/table/utils/__tests__/copy-utils.spec.ts
@@ -4,26 +4,90 @@ describe('copyCellValue: ', () => {
   let evt: {
     target: HTMLElement;
   };
-  beforeEach(() => {
-    console.log = jest.fn();
-  });
+  let writeMock: (arg: string) => void;
+  let elementClass: string;
+
+  const classList = {
+    contains: (className: string) => className === elementClass,
+  };
+
+  const closest = () => {
+    return {
+      querySelector: () => ({
+        textContent: 'textFromClosest',
+      }),
+    };
+  };
 
   beforeEach(() => {
-    jest.clearAllMocks();
-  });
-
-  it('should not send console error when copying cell values', () => {
+    elementClass = '';
     evt = {
       target: {
-        children: [],
+        closest,
+        querySelector: () => ({
+          textContent: 'text',
+        }),
+        classList,
       } as unknown as HTMLElement,
     };
+    jest.spyOn(console, 'log');
+    writeMock = jest.fn();
     Object.assign(navigator, {
-      clipboard: {
-        writeText: jest.fn().mockReturnValueOnce(Promise.resolve()),
-      },
+      clipboard: { writeText: writeMock },
     });
-    copyCellValue(evt);
-    expect(console.log).not.toHaveBeenCalled();
+  });
+
+  afterEach(() => jest.clearAllMocks());
+
+  it('should call clipboard.writeText but not console log for a table cell', async () => {
+    elementClass = 'sn-table-cell';
+    await copyCellValue(evt);
+    expect(global.console.log).toHaveBeenCalledTimes(0);
+    expect(writeMock).toHaveBeenCalledTimes(1);
+    expect(writeMock).toHaveBeenCalledWith('text');
+  });
+
+  it('should call clipboard.writeText but not console log for a non-table cell', async () => {
+    elementClass = '';
+    await copyCellValue(evt);
+    expect(global.console.log).toHaveBeenCalledTimes(0);
+    expect(writeMock).toHaveBeenCalledTimes(1);
+    expect(writeMock).toHaveBeenCalledWith('textFromClosest');
+  });
+
+  it('should not call clipboard.writeText nor console log when value is not defined', async () => {
+    evt = {
+      target: {
+        querySelector: () => undefined,
+        classList,
+      } as unknown as HTMLElement,
+    };
+    elementClass = 'sn-table-cell';
+    await copyCellValue(evt);
+
+    expect(global.console.log).toHaveBeenCalledTimes(0);
+    expect(writeMock).toHaveBeenCalledTimes(0);
+  });
+
+  it('should call console.log when writeText throws error', async () => {
+    writeMock = jest.fn().mockImplementation(() => {
+      throw new Error();
+    });
+    Object.assign(navigator, {
+      clipboard: { writeText: writeMock },
+    });
+    evt = {
+      target: {
+        querySelector: () => ({
+          textContent: 'text',
+        }),
+        classList,
+      } as unknown as HTMLElement,
+    };
+    elementClass = 'sn-table-cell';
+
+    await copyCellValue(evt);
+    expect(writeMock).toHaveBeenCalledTimes(1);
+    expect(global.console.log).toHaveBeenCalledTimes(1);
   });
 });
--- a/src/table/utils/__tests__/handle-key-press.spec.ts
+++ b/src/table/utils/__tests__/handle-key-press.spec.ts
@@ -22,7 +22,7 @@ describe('handle-key-press', () => {
 
   beforeEach(() => {
     areBasicFeaturesEnabled = true;
-    jest.spyOn(handleCopy, 'default');
+    jest.spyOn(handleCopy, 'default').mockImplementation(() => new Promise(() => {}));
   });
 
   afterEach(() => jest.clearAllMocks());
@@ -304,7 +304,6 @@ describe('handle-key-press', () => {
         target: {
           blur: jest.fn(),
           setAttribute: jest.fn(),
-          children: [],
         } as unknown as HTMLElement,
       } as unknown as React.KeyboardEvent;
       rootElement = {
@@ -402,7 +401,6 @@ describe('handle-key-press', () => {
         target: {
           blur: jest.fn(),
           setAttribute: jest.fn(),
-          children: [],
         } as unknown as HTMLElement,
       } as unknown as React.KeyboardEvent;
       cellCoord = [1, 1];
@@ -506,7 +504,6 @@ describe('handle-key-press', () => {
           classList: {
             contains: () => isExcluded,
           },
-          children: [],
         } as unknown as HTMLElement,
       } as unknown as React.KeyboardEvent;
       rootElement = {
--- a/src/table/utils/copy-utils.ts
+++ b/src/table/utils/copy-utils.ts
@@ -1,9 +1,11 @@
 const copyCellValue = async (evt: any) => {
-  const target = evt.target as HTMLElement;
+  let target = evt.target as HTMLElement;
 
-  const value = target.children[0]?.firstChild
-    ? target.children[0].firstChild.textContent
-    : target.firstChild?.textContent;
+  if (!target.classList.contains('sn-table-cell')) {
+    target = target.closest('.sn-table-cell') as HTMLElement;
+  }
+  const value = target.querySelector('.sn-table-cell-text')?.textContent;
+
   try {
     value && (await navigator.clipboard.writeText(String(value)));
   } catch (error) {
--- a/src/table/components/CellText.tsx
+++ b/src/table/components/CellText.tsx
@@ -1,0 +1,10 @@
+import React from 'react';
+import Box from '@mui/material/Box';
+
+export default function CellText({ children }: { children: React.ReactNode }) {
+  return (
+    <Box component="span" className="sn-table-cell-text" sx={{ borderLeft: '0 !important' }}>
+      {children}
+    </Box>
+  );
+}
--- a/src/table/components/TableBodyWrapper.tsx
+++ b/src/table/components/TableBodyWrapper.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useMemo, memo } from 'react';
-import Box from '@mui/material/Box';
 
 import getCellRenderer from '../utils/get-cell-renderer';
 import { useContextSelector, TableContext } from '../context';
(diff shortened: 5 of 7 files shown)


src/table/utils/copy-utils.ts
```diff
@@ -1,9 +1,20 @@
...
+  if (!target.classList.contains('sn-table-head-cell') && target.classList.contains('sn-table-cell')) {
+    value = target.firstChild?.textContent;
+  }
+  if (target.classList.contains('MuiTableSortLabel-root')) {
+    value = target.firstChild && target.firstChild.textContent;
+  }
+  if (target.classList.contains('MuiTableSortLabel-icon')) {
+    value = target.parentElement && target.parentElement.firstChild?.textContent;
+  }
+
```
Donya-qlik, line 17: This solution might not be a nice one but the only way I could cover all possible click-area.

src/table/utils/__tests__/handle-key-press.spec.ts
```diff
@@ -710,21 +734,21 @@ describe('handle-key-press', () => {
...
-    it('should call copyCellValue when the pressed keys are Meta and C keys', () => {
+    it.skip('should call copyCellValue when the pressed keys are Meta and C keys', () => {
       evt.key = KeyCodes.C;
       evt.metaKey = true;
       runHandleBodyKeyDown();
       expect(handleCopy.default).toHaveBeenCalled();
     });
 
-    it('should not call copyCellValue when the flag is disabled', () => {
+    it.skip('should not call copyCellValue when the flag is disabled', () => {
```
Donya-qlik, line 751: Unfortunately these tests can't get pass, there is a mock on `contains: () => isExcluded` that is required for the first statement in the body handleKey function, later it needs to return true for when we are checking the copy function calling inside handleKey, because of this conflict none of these test cases work :(

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -15,7 +15,21 @@ describe('copyCellValue: ', () => {
...
+        children: [
+          {
+            firstChild: {
+              textContent: 'Test',
+            },
+          },
+        ],
+        classList: {
+          contains: () => true,
+        },
```
Donya-qlik, line 32: I spent some times on adding more testing scenarios here but didn't go well. The way clipboard is tested in jest is very limited and only through mocks. clipboard doesn't really copy any value if the context is not secure. I couldn't find a way to force it to do. I think integration tests are more suitable for copy to clipboard cases

src/table/utils/copy-utils.ts
```diff
@@ -1,9 +1,20 @@
...
+  }
+  if (!target.classList.contains('sn-table-head-cell') && target.classList.contains('sn-table-cell')) {
+    value = target.firstChild?.textContent;
+  }
+  if (target.classList.contains('MuiTableSortLabel-root')) {
+    value = target.firstChild && target.firstChild.textContent;
+  }
+  if (target.classList.contains('MuiTableSortLabel-icon')) {
+    value = target.parentElement && target.parentElement.firstChild?.textContent;
+  }
```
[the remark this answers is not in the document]
haxxmaxx: not sure relying on mui classes is the best idea. we could use tags instead, the sort label is the only span you can click and the icon is the only svg

src/table/utils/__tests__/handle-key-press.spec.ts
haxxmaxx: I mean you should be able to mock the implementation of `copyCellValue`? All we want to do in these tests is to check if the `copyCellValue` function is called, not that it does what it originally should do

src/table/utils/__tests__/copy-utils.spec.ts
haxxmaxx: you can use `.className` instead of `.classList` and you get a string like this `'class1 class2 class3...'` and then you can use the string method `includes` instead

Donya-qlik: ... well, replacing it with className doesn't make a big change. The issue is clipboard itself, it is only available in secure context like https or localhost so in a unit test even without mocking it will not do copying so I cannot verify the string is copied ( to test that the cell value is copied as expected ).

haxxmaxx: ah ok, I misread the comment. Something like this works:
```
  let writeMock: (arg: string) => void;
  beforeEach(() => {
    console.log = jest.fn();
    writeMock = jest.fn();
    Object.assign(navigator, {
      clipboard: {
        writeText: writeMock,
      },
    });
  });
```
Also, you have one of the `beforeEach` should be a `afterEach`

Donya-qlik: no problem, yes we already have the similar mock, I actually meant that I would like to add more tests on what the clipboard result could be. It's only because we have few conditions to catch the right cell value so wanted to verify that the copied values are the expected ones, something like writeText and then later check the result by readText(), but because of clipboard limitations it is not really possible.

haxxmaxx: I mean whatever happens after we call writeText shoulldn't really be tested here. We want to make sure that writeText is called with the correct value. That is tested yet. Also we should test all three cases of setting the value.
We could possibly do a test in sense-client later for this, but not sure we should do integration tests on all the cases here

haxxmaxx:
Code looks good. I think we can extend the tests for copy-cell-values a bit since there now are three cases. Right now it looks like you have mocked the `writeText` method, but never check if it is called. We should be able to check that it was called and what it was called with

Donya-qlik:
> Code looks good. I think we can extend the tests for copy-cell-values a bit since there now are three cases. Right now it looks like you have mocked the `writeText` method, but never check if it is called. We should be able to check that it was called and what it was called with

Yes this is what I tried to say on the previous comments 😅 that clipboard has a limitation in general and in our case here too. We have covered those that copy function is called, but about writeText is what I was worry about because it should stay mocked, maybe if we bring it out of `copyCellValue` and just call it inside of it then spyOn works... ? Do we wanna have a new function `copyToClipboard` ? 🤔

src/table/utils/copy-utils.ts
```diff
@@ -1,9 +1,15 @@
 const copyCellValue = async (evt: any) => {
+  let value: string | undefined | null = '';
   const target = evt.target as HTMLElement;
 
-  const value = target.children[0]?.firstChild
-    ? target.children[0].firstChild.textContent
-    : target.firstChild?.textContent;
+  if (target.classList.contains('sn-table-head-cell')) {
+    value = target.children[0].firstChild?.textContent;
```
cbt1, line 6: This seem to be dependent on a very specific dom structure. Which adds the risk of making it break when that dom structure changes.

Possible risk mitigation for that could be to wrap the text in it's own span element that can then more easily be identifed?

src/table/utils/__tests__/handle-key-press.spec.ts
Donya-qlik: As we are using tagName instead it works now, so no need of `contains` anymore.

src/table/utils/copy-utils.ts
cbt1: Just some spontaneous thinking here but something like this:
```js
if (target.tagName === 'svg') {
 target = target.parentNode;
}

value = target.querySelector('.span-with-text').textContent;
```

haxxmaxx: yupp very good point. I was expecting this to break when we made changes, which we will soon.

That is a good suggestion. We would still need some logic to identify the parent then search for that span, but would be more general and robust. Not sure how much work it is to add an extra span though

haxxmaxx: yeah it would be something like
```
if (!target.className.includes('sn-table-cell')) {
 target = target.parentNode; // or maybe grandparent, or search up don't remember the 
}

value = target.querySelector('.span-with-text').textContent;
```

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -4,26 +4,128 @@ describe('copyCellValue: ', () => {
...
+          },
+        },
+        tagName: 'svg',
+        classList,
+      } as unknown as HTMLElement,
+    };
+    await copyCellValue(evt);
+
+    expect(global.console.log).toHaveBeenCalledTimes(0);
+    expect(writeMock).toHaveBeenCalledTimes(0);
```
Donya-qlik, line 129: @haxxmaxx ... but the test explanation says console.log should be called

src/table/utils/copy-utils.ts
cbt1: Maybe you can use [closest()](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) when you need to traverse up the tree?

src/table/utils/__tests__/copy-utils.spec.ts
haxxmaxx: ah yes sloppy. They both should be called. I need to add it to clipboard again. I just pushed the fix

src/table/utils/copy-utils.ts
haxxmaxx: should work

haxxmaxx: I tried wrapping all type of cells in spans, seems to work out of the box. Let's go for this approach, much more robust. Would be good to do some sanity checking to see that the screen reader behaves  the same

cbt1: Will you also simplify the if logic as talked about?

haxxmaxx: yes donya is working on that

src/table/utils/copy-utils.ts
```diff
@@ -1,9 +1,18 @@
 const copyCellValue = async (evt: any) => {
-  const target = evt.target as HTMLElement;
+  let value: string | undefined | null = '';
+  let target = evt.target as HTMLElement;
+
+  if (target.classList.contains('sn-table-cell-text')) {
```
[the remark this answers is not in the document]
cbt1: Actually it wont work as `querySelector` doesn't match itself. Would need an extra check there.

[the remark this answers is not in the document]
haxxmaxx: You are right, it includes itself in the search so that should work

cbt1 (approved):
I guess you could wrap that new Box component with the text in a custom component to limit the duplicated className and component props. But not really an issue so LGTM.

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -4,26 +4,94 @@ describe('copyCellValue: ', () => {
...
+
+  const classList = {
+    contains: (className: string) => className === elementClass,
+  };
+
+  const closest = (className: string) => {
+    className === closestElementClass;
+    return {
+      querySelector: jest.fn(() => ({
+        textContent: 'svg text',
```
haxxmaxx, line 19: svg text and head text doesn't really make sense, but it doesn't really matter.

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -4,26 +4,94 @@ describe('copyCellValue: ', () => {
...
+  let closestElementClass: string;
+
+  const classList = {
+    contains: (className: string) => className === elementClass,
+  };
+
+  const closest = (className: string) => {
+    className === closestElementClass;
+    return {
+      querySelector: jest.fn(() => ({
```
haxxmaxx, line 18: no need to use jest.fn here, you are not checking if it is called

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -4,26 +4,94 @@ describe('copyCellValue: ', () => {
...
+    elementClass = 'sn-table-cell';
+    await copyCellValue(evt);
+    expect(global.console.log).toHaveBeenCalledTimes(0);
+    expect(writeMock).toHaveBeenCalledTimes(1);
+    expect(writeMock).toHaveBeenCalledWith('head text');
+  });
+
+  it('should call clipboard.writeText but not console log for a non-table cell', async () => {
+    elementClass = 'sn-table-head-cell';
+    closestElementClass = 'sn-table-cell';
```
haxxmaxx, line 55: this is not a real case. If it has sn-table-head-cell class, it will have sn-table-cell class as well. You can just set elementClass to '' or 'someClass' if you want

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -4,26 +4,94 @@ describe('copyCellValue: ', () => {
...
+  };
+
   beforeEach(() => {
-    console.log = jest.fn();
+    elementClass = '';
+    closestElementClass = '';
+    evt = {
+      target: {
+        closest,
+        querySelector: jest.fn(() => ({
```
haxxmaxx, line 30: same here, no need for jest.fn

haxxmaxx (commented):
some clean up of the test file would be nice

haxxmaxx:
hmm rendering tests seem to be failing due to adding the span to the text. I will take a look

src/table/utils/__tests__/copy-utils.spec.ts
```diff
@@ -6,30 +6,27 @@ describe('copyCellValue: ', () => {
...
     elementClass = '';
-    closestElementClass = '';
     evt = {
       target: {
         closest,
-        querySelector: jest.fn(() => ({
-          textContent: 'head text',
-        })),
+        querySelector: () => ({
+          textContent: 'text',
```
haxxmaxx, line 28: very minor thing, but now you return the same value from `querySelector` and `closest().querySelector`, so you can't really tell what happened in the function. could change the other one to `'textFromClosest'` or whatever

haxxmaxx (approved):
Great work! Minor comment, feel free to merge without it if you want

src/table/utils/__tests__/copy-utils.spec.ts
Donya-qlik: Ahh, true! I removed them blindly. 
It is fixed now.

PR #727, opened 2022-10-25
fix: only make selections on left click

haxxmaxx:
with flag off it was possible to make selections with right click

### DIFF
--- a/src/table/utils/handle-click.ts
+++ b/src/table/utils/handle-click.ts
@@ -57,10 +57,12 @@ export const getSelectionMouseHandlers = (
   };
 
   const handleMouseUp = (evt: React.MouseEvent) => {
-    if (areBasicFeaturesEnabled && evt.button === 0) {
-      selectionDispatch({ type: SelectionActions.SELECT_MOUSE_UP, payload: { cell, evt, announce } });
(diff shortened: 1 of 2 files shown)


src/table/utils/handle-click.ts
```diff
@@ -57,10 +57,12 @@ export const getSelectionMouseHandlers = (
   };
 
   const handleMouseUp = (evt: React.MouseEvent) => {
-    if (areBasicFeaturesEnabled && evt.button === 0) {
-      selectionDispatch({ type: SelectionActions.SELECT_MOUSE_UP, payload: { cell, evt, announce } });
-    } else if (!areBasicFeaturesEnabled && cell.isSelectable) {
-      selectionDispatch({ type: SelectionActions.SELECT, payload: { cell, evt, announce } });
+    if (evt.button === 0) {
```
haxxmaxx, line 60: complexity goes up a bit for not early returning here, but I think it is no problem here +  readability is better

Donya-qlik:
Thanks for fixing it! It could be problem for downloading table because the unwanted selections used to change the table data size before exporting them.

Donya-qlik (approved):

PR #729, opened 2022-10-31
feat: custom column widths

haxxmaxx:
Known limitations

- [ ] hug calculations and min width do not adjust to font size
- [ ] the browser still enforces a min width, since the text does not use ellipsis yet. Bit tricky to override
- [ ] sticky header is broken
- [ ] only the column being dragged is updated when dragging, the fill columns adjust on mouse up

### DIFF
--- a/src/table/components/TableHeadWrapper.tsx
+++ b/src/table/components/TableHeadWrapper.tsx
@@ -35,7 +35,7 @@ function TableHeadWrapper({
 
   useEffect(() => {
     headRowRef.current && setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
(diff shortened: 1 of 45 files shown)


src/table/components/TableHeadWrapper.tsx
```diff
@@ -61,7 +73,7 @@ function TableHeadWrapper({
 
           return (
             <TableCell
-              sx={headerStyle}
+              style={style}
```
haxxmaxx, line 76: if you are dragging the column, the width updates very frequently. If we would use `sx`, mui would create a new class for every width value. That throws warnings and would blow  up the head in the DOM

PR #730, opened 2022-11-01
feat: initial infinite scroll

cbt1:
WIP for adding infinite scroll support. I'll add some tests if we find that this is the way to go. Also will have to make sure that this feature cannot be toggle on before I merge this PR.

Feature support in this PR:
- Scroll infinite* number of rows and/or columns
- If data contains more than 250k rows, show pagination in the footer.

Limitions:
- Only fetching data for cells that are visible to the user. Should consider using some kind of buffer to have cell data ready when scrolling in a resonable pace.
- None of the features in the "pagination table" support are supported yet. They will have to be added incrementally. Ex. selections, accessibility, right click..
- Is a stand-alone table, it should if possible be merged into the "pagination table" at some point. Possible when they are near feature parity.


cbt1:
Lets create a new PR on the main repo instead of my fork to get the build steps to work.

PR #731, opened 2022-11-01
refactor: align hooks

haxxmaxx:
Noticed some things we could clean up.
- wrap the selection reducer in a hook. Will ad more stuff to TableContext so want to keep it clean
- not passing layout to sorting. When it changes we rerender from root, so we can add the hypercube in the factory


Donya-qlik (approved):
lgtm

cbt1 (approved):

PR #733, opened 2022-11-07
feat: initial infinite scroll

cbt1:
Feature support in this PR:
- Scroll infinite* number of rows and/or columns (*if data contains more than 250k rows, show pagination in the footer)

Limitions:
- Only fetching data for cells that are visible to the user. Should consider using some kind of buffer to have cell data ready when scrolling in a resonable pace.
- None of the features in the "pagination table" support are supported yet. They will have to be added incrementally. Ex. selections, accessibility, right click..
- Is a stand-alone table, it should if possible be merged into the "pagination table" at some point. Possible when they are near feature parity.


src/table/Root.tsx
```diff
@@ -24,6 +26,17 @@ export function render(props: RenderProps, reactRoot?: ReactDom.Root) {
   );
 }
 
+export function renderVirtualizedTable(props: WrapperProps, reactRoot?: ReactDom.Root) {
+  const muiTheme = muiSetup('ltr');
+
+  reactRoot?.render(
+    <React.StrictMode>
```
haxxmaxx, line 33: FYI we scrapped strict mode since we got douple rerenders from useEffects I think it was

haxxmaxx: https://medium.com/geekculture/the-tricky-behavior-of-useeffect-hook-in-react-18-282ef4fb570a Might be ways around it though

src/index.ts
```diff
@@ -110,6 +129,7 @@ export default function supernova(env: Galaxy) {
               rect,
               footerContainer,
               announce,
+              model,
```
haxxmaxx, line 132: why are you adding model here?

src/table/components/virtualized-table/constants/index.ts
```diff
@@ -0,0 +1,9 @@
+const DEFAULT_ROW_HEIGHT = 32;
+
+const HEADER_HEIGHT = 32;
+
+const PAGINATION_HEIGHT = 49;
+
+const MAX_PAGE_SIZE = 250000;
```
haxxmaxx, line 7: we probably want to make this dependent on font size later

haxxmaxx (commented):
Pretty hard to review this much code. All of it looks good, but have not tried to understand the scrolling part hehe. We need to sync on how we should do styling. Maybe doesn't matter here, but it will down the line

src/table/components/virtualized-table/HeaderCell.tsx
```diff
@@ -0,0 +1,64 @@
...
+  style: React.CSSProperties;
+  data: {
+    columns: Column[];
+    headerStyle: GeneratedStyling;
+  };
+}
+
+const HeaderCell = ({ columnIndex, style, data }: HeaderCellProps) => {
+  const { columns, headerStyle } = data;
+  const datum = columns[columnIndex];
```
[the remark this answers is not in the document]
cbt1: Datum in this context means a single data point

src/index.ts
cbt1: That is likely some left over from the prototypes I made. Will remove it.

cbt1:
> Pretty hard to review this much code. All of it looks good, but have not tried to understand the scrolling part hehe. We need to sync on how we should do styling. Maybe doesn't matter here, but it will down the line

Agree, it just ended up being pretty hard to have less code with an actual working scroll solution

src/table/Root.tsx
cbt1: I don't really have any issue with it being run in strict mode. It works as intended in dev env. So IMO I might as well keep it to potentially catch issues down the road.

src/index.ts
```diff
@@ -82,14 +82,33 @@ export default function supernova(env: Galaxy) {
       const changeSortOrder = useSorting(model, layout.qHyperCube);
 
       const [pageInfo, setPageInfo] = useState(initialPageInfo);
+      const shouldRenderVirtualizedTable = layout.scrollType === 'virtualized';
       const [tableData] = usePromise(
         async () =>
-          env.carbon && !model?.getHyperCubeData
+          env.carbon || shouldRenderVirtualizedTable
```
haxxmaxx, line 88: not sure what the consequenses of removing the sanity check for `getHyperCubeData` is here

src/table/Root.tsx
haxxmaxx: okok, seems to be in dev mode as well

src/index.ts
```diff
@@ -82,14 +82,33 @@ export default function supernova(env: Galaxy) {
       const changeSortOrder = useSorting(model, layout.qHyperCube);
 
       const [pageInfo, setPageInfo] = useState(initialPageInfo);
+      const shouldRenderVirtualizedTable = layout.scrollType === 'virtualized';
```
haxxmaxx, line 85: how is this set?

haxxmaxx (commented):
Looks good in general, some more qustions

src/index.ts
cbt1: It should not have been remove. Will fix.

src/index.ts
```diff
@@ -82,14 +82,33 @@ export default function supernova(env: Galaxy) {
       const changeSortOrder = useSorting(model, layout.qHyperCube);
 
       const [pageInfo, setPageInfo] = useState(initialPageInfo);
+      const shouldRenderVirtualizedTable = false; // layout.scrollMode === 1;
```
cbt1, line 85: @haxxmaxx I ended up disabling this feature for now

haxxmaxx (approved):

PR #735, opened 2022-11-08
Add a drop down menu to snTable header

Donya-qlik:
The sn-table dropdown menu contains a menu button that opens the context menu when it is clicked. The menu currently has two options to sort the column cell values ascending/descending:
- Sort ascending
- Sort descending

The menu button is hidden by default, but will be visible as soon as the mouse hovers the head cell or the head cell is on focus or clicked. Clicking anywhere out side the head cell results the context menu to be closed and the menu button returns to the hidden mode.
To see the changes, table feature flag should be toggled: `PS_15585_SN_TABLE_BASIC_FEATURES`


src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export default function HeadCellMenu({ headerStyle }: { headerStyle: GeneratedStyling }) {
+  const [open, setOpen] = React.useState(false);
+  const anchorRef = React.useRef<HTMLButtonElement>(null);
+  // It returns focus to the button when we transitioned from !open -> open
+  const isOpen = React.useRef(open);
+
+  const handleClickOpen = (event: any) => {
+    event.stopPropagation();
```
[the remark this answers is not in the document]
Donya-qlik: I tried two couple of options but because of the on the way issues couldn't make them complete:
- Not using stopPtopagation but instead making a callback that confirms the menu is open so disabling sorting function, but because of some reasons it couldn't update the setState value at the right time.
- Using stopPropagation  but moving the ClickAwayEventListener to very higher level component to TableCell but then one of useRef hooks didn't get the right value and a console error came up that I couldn't resolve it.
Which do you think might make more sense to continue with?

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+import ArrowUpwardIcon from '@mui/icons-material/ArrowUpward';
+import ArrowDownwardIcon from '@mui/icons-material/ArrowDownward';
+import MoreHoriz from '@mui/icons-material/MoreHoriz';
+import { StyledMenuIconButton, StyledCellMenu } from '../styles';
+import { GeneratedStyling } from '../types';
+
+const ListItems = ({ children, itemTitle }: { children: any; itemTitle: string }) => {
+  return (
+    <ListItem disablePadding>
+      <ListItemButton>
```
haxxmaxx, line 21: do you really need the button component here?

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </ListItemButton>
+    </ListItem>
+  );
+};
+
+export default function HeadCellMenu({ headerStyle }: { headerStyle: GeneratedStyling }) {
+  const [open, setOpen] = useState(false);
+  const anchorRef = useRef<HTMLButtonElement>(null);
+  // It returns focus to the button when we transitioned from !open -> open
+  const prevOpen = useRef(open);
```
haxxmaxx, line 33: I think this can mean trouble. We want a single source of truth whether it is open or not

src/table/components/HeadCellMenu.tsx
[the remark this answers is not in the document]
haxxmaxx: Well, the main problem is that we have a button (open menu) inside a different button (sort). This is bad design. Not sure how we solve that, because it depends on the a11y design for the header... What I would do for now is to try to treat the sort "button" like it has a hole in it, where the menu button is. In the click handler for the sorting, check if the event comes from the menu button, and then early return.

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const handleClickOpen = (event: any) => {
+    event.stopPropagation();
+    setOpen((value) => !value);
+  };
+
+  const handleClickClose = (event: Event | React.SyntheticEvent) => {
+    !(anchorRef.current && anchorRef.current.contains(event.target as HTMLElement)) ? setOpen(false) : undefined;
```
haxxmaxx, line 41:
```suggestion
    !(anchorRef.current?.contains(event.target as HTMLElement)) && setOpen(false);
```

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+      anchorRef.current!.focus();
+    }
+    prevOpen.current = open;
+  }, [open]);
+
+  return (
+    <StyledCellMenu headerStyle={headerStyle}>
+      <StyledMenuIconButton
+        ref={anchorRef}
+        id="head-cell-button"
```
haxxmaxx, line 55: use the sn-table prefix for consistency `sn-table-head-menu-button`

src/table/hooks/use-head-cell-menu.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable */
+import React, { useEffect, useRef } from 'react';
+
+export default function useHeadCellMenu() {
```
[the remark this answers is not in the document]
haxxmaxx: hmm no not really adding any value I would say, just makes it harder to follow. the open states are very much tied to the component

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+        {({ TransitionProps, placement }) => (
+          <Grow
+            {...TransitionProps}
+            style={{
+              transformOrigin: placement === 'bottom-start' ? 'left top' : 'left bottom',
+            }}
+          >
+            <Paper sx={{ boxShadow: 15 }}>
+              <ClickAwayListener onClickAway={handleClickClose}>
+                <MenuList autoFocusItem={open} className="head-cell-menu" aria-labelledby="head-cell-button">
```
haxxmaxx, line 80: `sn-table-head-menu`

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  <nav aria-label="Sorting orders">
```
haxxmaxx, line 81: This is not a navigation menu, should not need the nav tag

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    <List>
+                      <ListItems itemTitle="Sort ascending">
```
haxxmaxx, line 83: will need to add proper strings for this eventually

src/table/components/TableHeadWrapper.tsx
```diff
@@ -70,27 +72,34 @@ function TableHeadWrapper({
...
-              >
-                <CellText>{column.label}</CellText>
-                {isFocusInHead && (
-                  <VisuallyHidden data-testid={`VHL-for-col-${columnIndex}`}>
-                    {translator.get('SNTable.SortLabel.PressSpaceToSort')}
-                  </VisuallyHidden>
-                )}
-              </StyledSortLabel>
-            </TableCell>
+              <StyledHeadMenu>
```
haxxmaxx, line 80: this is not what this component is, more like HeadCellContent

haxxmaxx (commented):
I will checkout the branch and see if I can get it to work

src/table/components/HeadCellMenu.tsx
Donya-qlik: Can't say surely it is. I mostly followed the standard context menu I found in different examples, for me it is more like a button in the context menu that is pressed and the event fires.

src/table/components/HeadCellMenu.tsx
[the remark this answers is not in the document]
Donya-qlik: ok I'll remove it but what should I put instead? just the text with the bigger size padding ?

src/table/components/HeadCellMenu.tsx
Donya-qlik: True, it is not really needed.

src/table/components/HeadCellMenu.tsx
Donya-qlik: Do you mean the itemTitle text? It is what design team has chosen.

src/table/hooks/use-head-cell-menu.ts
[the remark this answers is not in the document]
Donya-qlik: Sure, I'll delete it.

src/table/components/HeadCellMenu.tsx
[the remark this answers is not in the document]
haxxmaxx: no you wont actually change the styling. You just treat it as a "hole" in the click handler.
if (event.target === button) return

src/table/components/HeadCellMenu.tsx
haxxmaxx: yes but we need to add it to translation, so we call translator.get('somStringName') so it works for all languages

src/table/components/HeadCellMenu.tsx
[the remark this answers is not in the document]
Donya-qlik: Aha you mean 'sort' button, sorry I misunderstood first, got it now.

src/table/components/HeadCellMenu.tsx
Donya-qlik: Should it be added in sense-client or there is a place in sn-table repo for it ?

haxxmaxx (commented):
Also this needs to be behind flag

src/locale/locales/en.json
```diff
@@ -51,6 +51,14 @@
     "value": "Sorted in descending order.",
     "comment": "direction of sorting for a column, announced by screen readers. (tew 210929)"
   },
+  "SNTable.MenuItem.SortAscending": {
+    "value": "Sort ascending.",
```
haxxmaxx, line 55: remove the dot

src/table/utils/handle-click.ts
```diff
@@ -31,6 +31,18 @@ export const handleMouseDownLabelToFocusHeadCell = (evt: MouseEvent, rootElement
   updateFocus({ focusType: 'focus', cell: getCellElement(rootElement, [0, columnIndex]) });
 };
 
+export const handleClickToSort = (
```
haxxmaxx, line 34: no need to put it here. My point was to pit it ouside the elements, like we do for `handleKeyDown`

src/table/styles.ts
```diff
@@ -122,6 +123,46 @@ export const VisuallyHidden = styled('span')({
...
+  },
+  '.head-cell-menu': {
+    color: headerStyle.sortLabelColor,
+  },
+}));
+
+export const StyledTableCell = styled(TableCell, {
+  shouldForwardProp: (prop: string) => prop !== 'headerStyle',
+})(({ headerStyle }) => ({
+  ...headerStyle,
```
haxxmaxx, line 160: the headerStyle prop does not exist here, so this one should be static. There are a couple of examples of the syntax for this, e.g. `VisuallyHidden`

src/table/components/__tests__/HeadCellMenu.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+    fireEvent.click(getByRole('button'));
+    expect(getByRole('menu')).toBeVisible();
+  });
+
+  it('should show the menu items by opening the menu', () => {
+    const { getByRole, getByText } = renderTableHeadCellMenu();
+
+    fireEvent.click(getByRole('button'));
+    expect(getByText('SNTable.SortItem.SortAscending')).toBeVisible();
+    expect(getByText('SNTable.SortItem.SortDescending')).toBeVisible();
```
Donya-qlik, line 51: It's a weird situation, I've already modified the SortItem to MenuItem, but it only passes the test with these strings! 
As I've shown you, the translator doesn't work properly for me.

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+    setOpen(true);
+  };
+
+  const handleClickMenuClose = (event: Event | React.SyntheticEvent) => {
+    !anchorRef.current?.contains(event.target as HTMLElement) && setOpen(false);
+  };
+
+  return (
+    <StyledCellMenu headerStyle={headerStyle}>
+      <StyledMenuIconButton
```
haxxmaxx, line 49: set tabindex to -1 for now, so it is not possible to tab here

src/table/components/TableHeadWrapper.tsx
```diff
@@ -59,19 +59,9 @@ function TableHeadWrapper({
...
-              className="sn-table-head-cell sn-table-cell"
-              tabIndex={tabIndex}
-              aria-sort={ariaSort}
-              aria-pressed={isCurrentColumnActive}
-              onKeyDown={handleKeyDown}
-              onMouseDown={() => handleClickToFocusHead(columnIndex, rootElement, setFocusedCellCoord, keyboard)}
-              onClick={() => isInteractionEnabled && changeSortOrder(column)}
-            >
+          /* eslint-disable */
+          const HeadCell = () => {
```
haxxmaxx, line 63: this does not need to be a function

src/table/components/__tests__/HeadCellMenu.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
```
haxxmaxx, line 52: this can be the same test

src/table/styles.ts
```diff
@@ -122,6 +123,43 @@ export const VisuallyHidden = styled('span')({
...
+
+export const StyledCellMenu = styled('div', {
+  shouldForwardProp: (prop: string) => prop !== 'headerStyle',
+})(({ headerStyle }) => ({
+  '.MuiPaper-root': {
+    width: '220px',
+  },
+  '.MuiListItemText-primary, .MuiSvgIcon-root': {
+    fontSize: '16px',
+    color: headerStyle.sortLabelColor,
```
haxxmaxx, line 150: I think this and the one for the menu can be inherited

src/table/components/HeadCellMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+        ref={anchorRef}
+        className="sn-table-head-menu-button"
```
haxxmaxx, line 51: I think you want this to be id, otherwise you are breaking the aria-labeledby for the menu

src/table/components/TableHeadWrapper.tsx
```diff
@@ -89,7 +79,31 @@ function TableHeadWrapper({
...
+              className="sn-table-head-cell sn-table-cell"
+              tabIndex={tabIndex}
+              aria-sort={ariaSort}
+              aria-pressed={isCurrentColumnActive}
+              onKeyDown={handleKeyDown}
+              onMouseDown={() => handleClickToFocusHead(columnIndex, rootElement, setFocusedCellCoord, keyboard)}
+              onClick={(evt: React.MouseEvent) => handleClickToSort(evt, column, changeSortOrder, isInteractionEnabled)}
+            >
+              {areBasicFeaturesEnabled ? (
+                <HeadCellContent>
```
haxxmaxx, line 99: wait, why are you only using `HeadCellContent` when the flag is on? I think it could be there for both cases. Tried it quickly and it looks better when the arrows are next to the text, more like the design. Will save some work when I fix the styling later

Donya-qlik: What do you mean? when the flag is on it shows everything as the design says.  When it is off it looks exactly as same as it is on the old table. shouldn't it be like that  with flag

Donya-qlik: I guess you might forgot to set the flag to `true` when you run it on nebula ?

src/table/components/TableHeadWrapper.tsx
```diff
@@ -88,6 +79,32 @@ function TableHeadWrapper({
...
+                  <HeadCell />
+                  <HeadCellMenu headerStyle={headerStyle} translator={translator} />
+                </HeadCellContent>
+              ) : (
+                <HeadCellContent>
+                  <HeadCell />
+                </HeadCellContent>
+              )}
```
haxxmaxx, line 107:
```suggestion
              <HeadCellContent>
                <StyledSortLabel
                  headerStyle={headerStyle}
                  active={isCurrentColumnActive}
                  title={!constraints.passive ? column.sortDirection : undefined} // passive: turn off tooltips.
                  direction={column.sortDirection}
                  tabIndex={-1}
                  onMouseDown={(evt: React.MouseEvent) =>
                    handleMouseDownLabelToFocusHeadCell(evt, rootElement, columnIndex)
                  }
                >
                  <CellText>{column.label}</CellText>
                  {isFocusInHead && (
                    <VisuallyHidden data-testid={`VHL-for-col-${columnIndex}`}>
                      {translator.get('SNTable.SortLabel.PressSpaceToSort')}
                    </VisuallyHidden>
                  )}
                </StyledSortLabel>
                {areBasicFeaturesEnabled && <HeadCellMenu headerStyle={headerStyle} translator={translator} />}
              </HeadCellContent>
```

src/table/components/TableHeadWrapper.tsx
haxxmaxx: remove this variable, see my suggestion below

src/table/components/TableHeadWrapper.tsx
Donya-qlik: AAh, yes off course! I forgot to refactor it when both are with HeadCellContent

src/table/components/__tests__/TableHeadWrapper.spec.tsx
```diff
@@ -82,7 +82,17 @@ describe('<TableHeadWrapper />', () => {
     expect(queryByText(tableData.columns[1].label)).toBeVisible();
   });
 
-  it('should call changeSortOrder when clicking a header cell', () => {
+  it.skip('should show the menu button when the head cell is on focus', () => {
+    areBasicFeaturesEnabled = true;
```
Donya-qlik, line 86: I'm going to skip this for now. I couldn't verify that the button is visible after focusing the head cell element. The focus is a css focus so I don't see any other option to focus on it than what it is now. It used to work before rebasing the branch. By the way, the current scenario works properly in reality on the table.

src/table/components/TableHeadWrapper.tsx
```diff
@@ -60,7 +61,7 @@ function TableHeadWrapper({
 
           return (
             <StyledHeadCell
-              sx={headerStyle}
+              sx={{ ...headerStyle, zIndex: columns.length - columnIndex }}
```
haxxmaxx, line 64: you can remove this now since the menu is below the header

src/table/components/__tests__/TableHeadWrapper.spec.tsx
haxxmaxx: ok that is fine

haxxmaxx (approved):
looks good for now. All the styling will probably be adjusted when there is a pattern in place and we are using the complete sprout. Wait for string approvals before you merge

tklkan (approved):

PR #736, opened 2022-11-14
refactor: remove mui overrides in mui-setup

haxxmaxx:
Beginning to refactor to remove the overrides in `mui-setup`. We are applying styles in so many places. It's decided to use styled whenever possible. It creates the cleanest component files and is faster than sx. The overrides in mui-setup where really hard to read also. Keeping `MuiTable.styleOverrides.root = {}` since it contains a lot of stuff we don't want

This refactor slightly changes the look of the selects. But those will look different when we actually change the styling, so I am updating the artifacts for that


src/table/styles.ts
```diff
@@ -102,9 +136,21 @@ export const StyledHeadRow = styled(TableRow, {
   },
 }));
 
+export const StyledHeadCell = styled(TableCell)({
+  ...cellCommon,
+  lineHeight: '150%',
+});
```
Donya-qlik, line 142: I've added the similar style in my PR but never mind I'll update it with yours after you merge it.

Donya-qlik (approved):
Totally agree with the refactor you made over styles 👍 
Just a commented code line you missed to remove. 
LGTM!

src/table/styles.ts
haxxmaxx: ok

src/table/styles.ts
```diff
@@ -45,6 +59,7 @@ export const StyledSelect = styled(Select, {
   shouldForwardProp: (prop: string) => prop !== 'paginationTheme',
 })(({ paginationTheme }) => ({
   backgroundColor: 'inherit',
+  // padding: '0px 12px',
```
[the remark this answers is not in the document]
haxxmaxx: I have another PR ready already. i'll do it in the next one

PR #739, opened 2022-11-16
refactor: move compoents to sub-folders

haxxmaxx:
- split up into three sub-folders: head, body, footer
- move tests to those folders
- create a styles file for each folder
- move common cell styling to constants file
- remove usages of `sx`, for `CellText` and `StyledHeadCell`


mhs30 (approved):
Safe PR code lgtm.

PR #740, opened 2022-11-16
feat: change data section definition

linhnguyen-qlik:
Change data section definition.

### DIFF
--- a/src/ext/property-panel/data.js
+++ b/src/ext/property-panel/data.js
@@ -165,12 +165,8 @@ const getData = (env) =>
     ? {
         type: 'items',
         translation: 'Common.Data',
-        items: {
-          dataAssets: {
-            type: 'items',
-            component: 'data-assets-panel',
-          },
-        },
+        component: 'data-assets-panel',
+        items: {},
       }
     : {
         type: 'items',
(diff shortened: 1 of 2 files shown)


src/ext/property-panel/data.js
```diff
@@ -165,12 +165,8 @@ const getData = (env) =>
     ? {
         type: 'items',
         translation: 'Common.Data',
-        items: {
-          dataAssets: {
-            type: 'items',
-            component: 'data-assets-panel',
-          },
-        },
+        component: 'data-assets-panel',
```
mhs30, line 168: Why change the structure here? does this follow the property panel common structure? will this change affect how the data component will render?

linhnguyen-qlik: Yes it follows the property panel structure. No it wont. It will only affect the rendering of DataAssetPanel. This change is necessary to make the DataAssetPanel expand to 100% of the height of parent component.

mhs30: Ok, sounds good to me, i would like to have @veinfors opinion on this one.

veinfors: Looks good to me!

veinfors (approved):

haxxmaxx (approved):

PR #746, opened 2022-11-21
chore: enable system js build

Caele:
Updates the Nebula dependency and enables the systemJS build artifact.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     ]
   },
   "devDependencies": {
-    "@babel/core": "7.19.3",
+    "@babel/core": "7.19.6",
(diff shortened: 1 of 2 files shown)


haxxmaxx (approved):

Caele:
Had to leave nebula/cli-build at 3.0.3 due to an update to @rollup/plugin-node-resolve v14 which somehow breaks the table in particular.
The Nebula PR is this: https://github.com/qlik-oss/nebula.js/pull/946
The change in @rollup/plugin-node-resolve is https://github.com/rollup/plugins/pull/1245
I have yet to determine what actually happens and how to resolve it.

PR #749, opened 2022-11-25
fix: to large data fetch in virtualized table

cbt1:
Fixes an issue where the virt table would call engine with a negative qTop value when fetching data.

### DIFF
--- a/src/table/components/virtualized-table/TableContainer.tsx
+++ b/src/table/components/virtualized-table/TableContainer.tsx
@@ -1,4 +1,4 @@
-import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
+import React, { memo, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { VariableSizeGrid, VariableSizeList } from 'react-window';
 import { getColumns } from '../../../handle-data';
 import useColumnSize from './hooks/use-column-size';
@@ -9,7 +9,7 @@ import Header from './Header';
 import { TableContainerProps } from './types';
 import { getHeaderStyle, getBodyCellStyle } from '../../utils/styling-utils';
 
-export default function TableContainer(props: TableContainerProps) {
+const TableContainer = (props: TableContainerProps) => {
   const { layout, rect, pageInfo, paginationNeeded, model, theme } = props;
   const ref = useRef<HTMLDivElement>(null);
   const headerRef = useRef<VariableSizeList>(null);
@@ -85,4 +85,9 @@ export default function TableContainer(props: TableContainerProps) {
       </FullSizeContainer>
     </div>
   );
-}
+};
+
+// Export non memoized version for testing purpose
+export { TableContainer };
+
+export default memo(TableContainer);
--- a/src/table/components/virtualized-table/Wrapper.tsx
+++ b/src/table/components/virtualized-table/Wrapper.tsx
@@ -1,21 +1,26 @@
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import { WrapperProps } from './types';
 import PaginationContent from '../PaginationContent';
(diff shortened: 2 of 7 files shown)


src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
```diff
@@ -47,8 +48,7 @@ const useInfiniteScrollData = (
   pageInfo: PageInfo
 ): UseInfiniteScrollData => {
   const [rowsInPage, setRowsInPage] = useState<Row[]>([]);
-
-  useEffect(() => setRowsInPage([]), [layout, pageInfo]);
+  useOnPropsChange(() => setRowsInPage([]), [layout, pageInfo]);
```
cbt1, line 51: By not using useEffect here we get 1 less render cycle. Come out of me trying to fix the original issue but ended up being valuable.

src/table/components/virtualized-table/TableContainer.tsx
```diff
@@ -85,4 +85,9 @@ export default function TableContainer(props: TableContainerProps) {
       </FullSizeContainer>
     </div>
   );
-}
+};
+
+// Export non memoized version for testing purpose
+export { TableContainer };
+
+export default memo(TableContainer);
```
cbt1, line 93: On the initial render the `Wrapper` component got an extra render due to the `keyboard` prop being updated. For which the was no reason to re-render so added the memo.

src/table/components/virtualized-table/TableContainer.tsx
```diff
@@ -85,4 +85,9 @@ export default function TableContainer(props: TableContainerProps) {
       </FullSizeContainer>
     </div>
   );
-}
+};
+
+// Export non memoized version for testing purpose
+export { TableContainer };
```
haxxmaxx, line 91: for the record it is possible to test when you export memo(someComponent) but the syntax is weird so I like this better actually

src/table/components/virtualized-table/TableContainer.tsx
haxxmaxx: nice that you are adding memo very intentionally

src/table/components/virtualized-table/Wrapper.tsx
```diff
@@ -1,21 +1,26 @@
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import { WrapperProps } from './types';
 import PaginationContent from '../PaginationContent';
 import { StyledTableWrapper } from '../../styles';
 import { PageInfo, TableData } from '../../../types';
 import FooterWrapper from '../FooterWrapper';
-import TableContainer from './TableContainer';
+import Table from './TableContainer';
```
haxxmaxx, line 7: how come the rename? I think it is good to keep the name of the component and the file the same

src/table/components/virtualized-table/hooks/use-on-props-change.ts
```diff
@@ -0,0 +1,12 @@
+import { useState } from 'react';
+
+// Act as a wrapper for https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes
+export default function useOnPropsChange<T>(callback: () => void, props: T[]) {
```
haxxmaxx, line 4: nice one

haxxmaxx (approved):
lgtm, some minor things

src/table/components/virtualized-table/Wrapper.tsx
cbt1: Good question. Got lint issues when the memoized TableContainer and TableContainer was both exported with the same name, but the memoized version as default. Did not look further into how to best solve it. But I do agree that I would prefer the same name.

PR #751, opened 2022-11-25
feat: selections in virtualized table

cbt1:
Adds support for single selections and multi selections by dragging.

The implementation is using some logic from the ordinary table, but I was not able to use it all, without modifying the shared code. Which would have added a risk to the ordinary table. So I opted for a solution where I used the logic that I could.


src/table/components/virtualized-table/__tests__/TableContainer.spec.tsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
 import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
```
cbt1, line 3: Had to use `@testing-library/user-event` to get the proper chain of events triggerd by a click

src/table/components/virtualized-table/__tests__/TableContainer.spec.tsx
```diff
@@ -69,7 +106,90 @@ describe('<TableContainer />', () => {
...
+
+      const { getByText, user } = renderTableContainer();
+
+      await waitFor(() => expect(getByText(cellText)).toBeVisible());
+
+      await waitFor(() => expect(getByText(cellText).parentNode).not.toHaveClass('selected'));
+
+      user.click(getByText(cellText));
+
+      await waitFor(() => expect(getByText(cellText).parentNode).toHaveClass('selected'));
```
cbt1, line 128: Don't really like the idea of checking a class exists or not. But only other option I could come up with was to add a testid which would add some additional logic just to make the code slight better in terms of testability. Still undecided on this I guess.

src/table/components/virtualized-table/hooks/use-selector.ts
```diff
@@ -0,0 +1,42 @@
...
+import { SelectionStates } from '../../../constants';
+import { TableContext, useContextSelector } from '../../../context';
+import { CellStyle } from '../../../types';
+import { getSelectionMouseHandlers } from '../../../utils/handle-click';
+import { getCellSelectionState } from '../../../utils/selections-utils';
+import { getSelectionStyle } from '../../../utils/styling-utils';
+
+const NOOP_ANNOUNCE = () => {};
+
+const PREVENT_DEFAULT = (e: React.MouseEvent<HTMLTableCellElement>) => e.preventDefault();
```
haxxmaxx, line 11: is it common practice to do these in snake case?

src/table/components/virtualized-table/Cell.tsx
```diff
@@ -19,32 +21,38 @@ const Cell = ({ columnIndex, rowIndex, style, data }: CellProps) => {
...
     return (
       <div
         style={{
           ...style,
+          ...bodyStyle,
+          ...selectionStyling,
           display: 'flex',
           alignItems: 'center',
-          borderColor: bodyStyle.borderColor,
-          borderStyle: 'solid',
```
haxxmaxx, line 30: I have remove borderStyle from the styling-utils. There are no borders now when I am testing it

haxxmaxx (approved):
Looks good in general. How much effort do you think it would be to align the two implementations? what parts would you not need to add here in that case? Reasonable to do it later?

src/table/components/virtualized-table/Cell.tsx
cbt1: Fixed that now

src/table/components/virtualized-table/hooks/use-selector.ts
cbt1: Actually no, will fix

PR #755, opened 2022-11-29
docs(mashup): add a sn-table mashup example in saas with platform sdk

LiKang6688:


a-m-dev (approved):
This is a decent implementation using `webIntegrationId` and seems like nothing wrong with it! 

The only thing is that should we expose tenantUrl, our integration id and app id with it? 

I would say in `main.js` in the `connect({ ... })` you should do something like: 

```js
const app = await connect({
    url: 'https://xxxx.us.qlik.com', //  note: https:// is needed
    webIntegrationId: 'xxx-xxxxxxx-xxxxxxxx',
    appId: 'xxxx-xxx-xxx-xxx-xxxxxxx',
  });
```

because eventually that app and that tenant is not so much reliable to be open sourced as an example

LiKang6688:
> This is a decent implementation using `webIntegrationId` and seems like nothing wrong with it!
> 
> The only thing is that should we expose tenantUrl, our integration id and app id with it?
> 
> I would say in `main.js` in the `connect({ ... })` you should do something like:
> 
> ```js
> const app = await connect({
>     url: 'https://xxxx.us.qlik.com', //  note: https:// is needed
>     webIntegrationId: 'xxx-xxxxxxx-xxxxxxxx',
>     appId: 'xxxx-xxx-xxx-xxx-xxxxxxx',
>   });
> ```
> 
> because eventually that app and that tenant is not so much reliable to be open sourced as an example

Of course not. I forgot to remove them after testing. Thx!

PR #757, opened 2022-11-30
feat: use constraints in vt mode

cbt1:
First part in consuming the constraints from nebula. This PR removes the pagination and disable scrolling given that the `active` constraint is `true`.

### DIFF
--- a/src/table/components/virtualized-table/TableContainer.tsx
+++ b/src/table/components/virtualized-table/TableContainer.tsx
@@ -1,4 +1,4 @@
-import React, { memo, useLayoutEffect, useMemo, useRef, useState } from 'react';
+import React, { memo, useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { VariableSizeGrid, VariableSizeList } from 'react-window';
 import { getColumns } from '../../../handle-data';
 import useColumnSize from './hooks/use-column-size';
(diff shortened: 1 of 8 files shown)


src/table/components/virtualized-table/TableContainer.tsx
```diff
@@ -21,26 +22,7 @@ const TableContainer = (props: TableContainerProps) => {
...
-    }
-
-    if (innerForwardRef.current) {
-      // Keep full size container in sync with the height calculation in react-window is doing
-      if (totalHeight !== innerForwardRef.current.clientHeight) {
-        setTotalHeight(innerForwardRef.current.clientHeight);
-      }
-    }
-  };
+  const scrollHandler = useScrollHandler(headerRef, bodyRef, innerForwardRef, totalHeight, setTotalHeight);
```
cbt1, line 25: Sry about this one, does not really have anything to do with constraints. I ended up doing a change to it as I tought it would be needed, but then I didn't need it. I still want to keep the refactoring part of my change.

There should be no change other than refactoring and added tests for the scroll handler.

src/table/components/virtualized-table/__tests__/Wrapper.spec.tsx
```diff
@@ -49,6 +58,8 @@ describe('<Wrapper />', () => {
         pagination: { borderColor: '' },
       },
     } as unknown as ExtendedTheme;
+
+    constraints = { active: false };
```
haxxmaxx, line 62: Could be good to know: I think it is never false, it is either undefined or true

src/table/components/virtualized-table/hooks/__tests__/use-scroll-hander.test.ts
```diff
@@ -0,0 +1,71 @@
+import { renderHook } from '@testing-library/react';
```
haxxmaxx, line 1: the file name should be use-scroll-handler.spec.ts right?

src/table/components/virtualized-table/TableContainer.tsx
haxxmaxx: np, thanks for explaining

src/table/components/virtualized-table/hooks/__tests__/use-scroll-hander.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  test('should scroll body and header', () => {
+    const headerScrollToSpy = jest.spyOn(listInstance, 'scrollTo');
+    const bodyScrollToSpy = jest.spyOn(gridInstance, 'scrollTo');
+
+    const { result } = renderHook(() =>
```
haxxmaxx, line 38: Didn't know about this, it could be really useful in some other places

haxxmaxx (approved):

src/table/components/virtualized-table/hooks/__tests__/use-scroll-hander.test.ts
cbt1: Indeed. Will fix 👍

PR #759, opened 2022-12-01
feat: copy cell value in vt mode

cbt1:


haxxmaxx (approved):
Nice that something basically worked out of the box

PR #761, opened 2022-12-02
fix: closing menu by clicking head cell causes sorting order

Donya-qlik:
Issues that are fixed:

- Clicking the button does not close the context menu if it is already open.
- Showing menu button when the head cell is hovered or focused
- Making selection on menu-item doesn't close the menu

After fix:

- Clicking the menu button should open the menu as well as closing it.
- Hovering, focusing, clicking the head cell appears the menu button
- Menu will be closed as soon as a menu item is clicked

### DIFF
--- a/src/table/components/head/HeadCellMenu.tsx
+++ b/src/table/components/head/HeadCellMenu.tsx
@@ -20,15 +20,22 @@ const MenuItem = ({
   itemTitle,
   sortOrder,
   sortFromMenu,
+  setOpen,
 }: {
   children: any;
   itemTitle: string;
   sortOrder: string;
   sortFromMenu(evt: React.MouseEvent, sortOrder: string): void;
+  setOpen: any;
 }) => {
   return (
     <ListItem disablePadding>
-      <ListItemButton onClick={(evt) => sortFromMenu(evt, sortOrder)}>
+      <ListItemButton
+        onClick={(evt) => {
+          sortFromMenu(evt, sortOrder);
+          setOpen(false);
+        }}
+      >
         <ListItemIcon sx={{ minWidth: '25px' }}>{children}</ListItemIcon>
         <ListItemText primary={itemTitle} />
       </ListItemButton>
@@ -48,14 +55,6 @@ export default function HeadCellMenu({
   const [open, setOpen] = useState(false);
   const anchorRef = useRef<HTMLButtonElement>(null);
 
-  const handleClickMenuOpen = () => {
-    setOpen(true);
-  };
-
-  const handleClickMenuClose = (event: Event | React.SyntheticEvent) => {
-    !anchorRef.current?.contains(event.target as HTMLElement) && setOpen(false);
-  };
-
   return (
     <StyledCellMenu headerStyle={headerStyle}>
       <StyledMenuIconButton
@@ -65,7 +64,7 @@ export default function HeadCellMenu({
         aria-controls={open ? 'sn-table-head-menu' : undefined}
         aria-expanded={open ? 'true' : undefined}
         aria-haspopup="true"
-        onClick={handleClickMenuOpen}
+        onClick={() => setOpen(!open)}
       >
         <MoreHoriz />
       </StyledMenuIconButton>
@@ -93,7 +92,7 @@ export default function HeadCellMenu({
             }}
           >
             <Paper sx={{ boxShadow: 15 }}>
-              <ClickAwayListener onClickAway={handleClickMenuClose}>
+              <ClickAwayListener onClickAway={() => setOpen(false)}>
                 <MenuList
                   autoFocusItem={open}
                   className="sn-table-head-menu"
@@ -103,6 +102,7 @@ export default function HeadCellMenu({
                     itemTitle={translator.get('SNTable.MenuItem.SortAscending')}
                     sortOrder="A"
                     sortFromMenu={sortFromMenu}
+                    setOpen={setOpen}
                   >
                     <ArrowUpwardIcon />
                   </MenuItem>
@@ -111,6 +111,7 @@ export default function HeadCellMenu({
                     itemTitle={translator.get('SNTable.MenuItem.SortDescending')}
                     sortOrder="D"
                     sortFromMenu={sortFromMenu}
+                    setOpen={setOpen}
                   >
                     <ArrowDownwardIcon />
                   </MenuItem>
--- a/src/table/components/head/__tests__/TableHeadWrapper.spec.tsx
+++ b/src/table/components/head/__tests__/TableHeadWrapper.spec.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { stardust } from '@nebula.js/stardust';
-import { render, fireEvent } from '@testing-library/react';
+import { render, fireEvent, waitFor } from '@testing-library/react';
+import useEvent from '@testing-library/user-event';
 import TableHeadWrapper from '../TableHeadWrapper';
 import { TableContextProvider } from '../../../context';
 import * as handleKeyPress from '../../../utils/handle-key-press';
@@ -73,6 +74,7 @@ describe('<TableHeadWrapper />', () => {
       enabled: false,
     } as stardust.Keyboard;
     translator = { get: (s) => s } as ExtendedTranslator;
+    areBasicFeaturesEnabled = false;
   });
 
   it('should render table head', () => {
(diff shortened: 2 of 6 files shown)


src/table/components/head/TableHeadWrapper.tsx
```diff
@@ -80,7 +87,9 @@ function TableHeadWrapper({
               aria-pressed={isCurrentColumnActive}
               onKeyDown={handleKeyDown}
               onMouseDown={() => handleClickToFocusHead(columnIndex, rootElement, setFocusedCellCoord, keyboard)}
-              onClick={(evt: React.MouseEvent) => handleClickToSort(evt, column, changeSortOrder, isInteractionEnabled)}
+              onClick={(evt: React.MouseEvent) =>
+                !menuOpen && handleClickToSort(evt, column, changeSortOrder, isInteractionEnabled)
```
Donya-qlik, line 91: Couldn't find a way to test it. menuOpen is false by default and there is no way to set it to true in the test file.

src/table/components/head/__tests__/TableHeadWrapper.spec.tsx
```diff
@@ -82,7 +82,7 @@ describe('<TableHeadWrapper />', () => {
     expect(queryByText(tableData.columns[1].label)).toBeVisible();
   });
 
-  it.skip('should show the menu button when the head cell is on focus', () => {
+  it('should show the menu button when the head cell is on focus', () => {
```
Donya-qlik, line 85: I did whatever I could to make it work but it doesn't : ( 
Hovering and focusing works properly on the browser but can't verify it here that it make the button visible through this test. I tried with userEvent.hover or tab and got the same result. It doesn't show that the opacity is changing to '1'.

src/table/components/head/TableHeadWrapper.tsx
```diff
@@ -69,6 +70,12 @@ function TableHeadWrapper({
             changeSortOrder(column, sortOrder);
           };
 
+          // Check context menu state.
+          // If it is open, clicking on the head cell to close the menu will not cause changing sort order.
+          const menuState = (state: boolean) => {
+            setMenuOpen(state);
+          };
```
haxxmaxx, line 77: this is a pointless wrapper, might as well just use `setMenuOpen`

src/table/components/head/__tests__/TableHeadWrapper.spec.tsx
haxxmaxx: This might be a timing issue. Does the first `expect` pass? If so, try a small manual timeout before the second `expect`. If that works we need to figure out a proper way to do the timeout

src/table/components/head/TableHeadWrapper.tsx
haxxmaxx: So here you are mixing where you put the logic. You extracted `handleClickToSort` in a previous PR to make it testable. Now you check `!menuOpen` before you call the function. I think we can add `!menuOpen` to `isInteractionEnabled`

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -40,20 +40,24 @@ export default function HeadCellMenu({
...
   translator,
   sortFromMenu,
+  menuState,
 }: {
   headerStyle: GeneratedStyling;
   translator: ExtendedTranslator;
   sortFromMenu: (evt: React.MouseEvent, sortOrder: string) => void;
+  menuState: (value: boolean) => void;
 }) {
   const [open, setOpen] = useState(false);
```
haxxmaxx, line 50: this state is not needed anymore. There should only be one single source of truth. Pass both the state and the setter. Basically we want to move the state one step up

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -40,20 +40,24 @@ export default function HeadCellMenu({
...
     setOpen(true);
+    menuState(true);
   };
 
-  const handleClickMenuClose = (event: Event | React.SyntheticEvent) => {
-    !anchorRef.current?.contains(event.target as HTMLElement) && setOpen(false);
+  const handleClickMenuClose = () => {
+    setOpen(false);
+    menuState(false);
   };
```
haxxmaxx, line 61: remove both of these

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -65,7 +69,7 @@ export default function HeadCellMenu({
         aria-controls={open ? 'sn-table-head-menu' : undefined}
         aria-expanded={open ? 'true' : undefined}
         aria-haspopup="true"
-        onClick={handleClickMenuOpen}
+        onClick={!open ? handleClickMenuOpen : handleClickMenuClose}
```
haxxmaxx, line 72: This should be the same thing. 

```suggestion
        onClick={() => setMenuOpen(!menuOpen)}
```

src/table/components/head/TableHeadWrapper.tsx
```diff
@@ -35,6 +35,7 @@ function TableHeadWrapper({
   const headRowRef = useRef<HTMLElement>();
   const isInteractionEnabled = !constraints.active && !selectionsAPI.isModal();
 
+  const [menuOpen, setMenuOpen] = useState(false);
```
haxxmaxx, line 38: bit ocd here, but put the state with the other use things, above `isInteractionEnabled`

src/table/components/head/TableHeadWrapper.tsx
Donya-qlik: Yes true, if we keep just menuOpen useState here then it makes no sense

src/table/components/head/TableHeadWrapper.tsx
Donya-qlik: What do you mean 'other use things'? [open, setOpen] ?

src/table/components/head/__tests__/TableHeadWrapper.spec.tsx
cbt1: If it's a timing issue you could try to wrap you `expect` in a `waitFor()` callback

haxxmaxx: nice!

src/table/utils/handle-click.ts
```diff
@@ -37,7 +37,10 @@ export const handleClickToSort = (
   changeSortOrder: ChangeSortOrder,
   isInteractionEnabled: boolean
 ) => {
-  !(evt.target as HTMLElement).closest('#sn-table-head-menu-button') && isInteractionEnabled && changeSortOrder(column);
+  !document.querySelector('.sn-table-head-menu') &&
```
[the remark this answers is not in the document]
haxxmaxx: I see. It is because the header doesn't update when you open and close the menu. This is a good thing. But that means we can't set a variable in the head component, since it will be outdated when we open and close the menu. And we can't use document.

Could you scope this part of the PR out? four verifails in one PR is already a lot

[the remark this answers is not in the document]
haxxmaxx: The thing is that I am not sure we should solve it. The solutions that would work are pretty ugly and it now that I think about it, it is breaking the patterns we have for other context menus. Clicking outside on a button will both close the menu and activate the button

src/table/components/head/HeadCellMenu.tsx
a-m-dev: I think the current implementation is fair. because the direct parent of `<HeadCellMenu />` component is `<TableHeadWrapper />` which iterates through all columns inside of it's return section and does all the logic for rendering there. 

So, the scenario that Donya has mentioned (tried it before -> toggling all dropdown at once) is make sense now, because `<TableHeadWrapper />` [is the holder of all columns in table](https://github.com/qlik-oss/sn-table/blob/eac5ee9375a9a234136ab56254d68231f2153355/src/table/components/head/TableHeadWrapper.tsx#L45), and not one of them (to manage the dropdown of only one cell)

## Solution: 
**The problem that i see here is:** that we don't have a middle component of `<HeadCell />` that would sit in between `<TableHeadWrapper />` and `<HeadCellMenu />` => which would handle the state of toggle dropdown and any other scoped logic for it's own. 

So by that I could propose distinct 2 solution:
1. If we would have the middle component, the logic that Max has mentioned would be kept in there + would also result in reducing testing complexity as well
2. if we still want to move state up, we have to keep track of which component is active in order to tie the dropdown open state into it.

2nd solution is kinda easy fix for this but **1st one the the right way to do it IMO**

haxxmaxx: First of all, see my other comment about scoping the fix for clicking the header when menu is open. I don't think we should fix it at all. blocking a click for something else when the menu is open, we don't do that anywhere else.

If we need a new component, that is fine. I like the principle of developing and structuring code to what you know you need. So I am not a fan of point nr 4. But if we need it for the listbox and the refactor makes sense, we should do it.

Donya-qlik: Thanks @a-m-dev for breaking it through and your solutions. Max had the idea that we can skip passing the menu open/close state to the parent `TableHeadWrapper` to prevent getting all menu open but stopping extra head cell sorting by checking the existence of context menu in the DOM which we faced an issue there.
BTW, if you still think you want to see the whole issue in a clean PR here you have it: https://github.com/qlik-oss/sn-table/pull/765

haxxmaxx (approved):

src/table/components/head/HeadCellMenu.tsx
Donya-qlik: @haxxmaxx ..so for being sure I got it properly..  you believe that we should skip the extra sorting on head cell issue when the click-away(to close the context menu) occurs on the head cell ? and you thinks it's because we might break the same pattern in the other dropdowns ?

haxxmaxx: you should be able to click the header to sort when the menu is open. So the verifail can be closed as works as designed

a-m-dev (approved):

PR #763, opened 2022-12-06
refactor: move non-background styling from extendedTheme to styling-utils

haxxmaxx:
The extended theme hook calculated things that we normally do in the styling-utils. I moved everything that is not related to calculating the background out of use-extended-theme. Still keeping the background stuff there since we want it centralized on the theme
- rename `tableThemeColors` to `getBackgroundColors` and store the result under `background` on the extended theme
  - remove `background` from the varialbe names inside that funciton, since it is implied that we are talking about background
- create `getFooterStyle` to align with body and header
- added some descriptions for the color utils

### DIFF
--- a/src/nebula-hooks/use-extended-theme.ts
+++ b/src/nebula-hooks/use-extended-theme.ts
@@ -1,43 +1,35 @@
 /* eslint-disable react-hooks/exhaustive-deps */
 import { useMemo, useTheme } from '@nebula.js/stardust';
 import { isDarkColor, isTransparentColor } from '../table/utils/color-utils';
-import { ExtendedTheme, TableThemeColors } from '../types';
+import { ExtendedTheme, BackgroundColors } from '../types';
 
-export const tableThemeColors = (theme: ExtendedTheme, rootElement: HTMLElement): TableThemeColors => {
+/**
+ * The colors in the table can depend on background colors set on table, object and sheet level.
+ * Even though not officially supported, it is expected that css settings on object and sheet are respected
+ * The priority order is: table theme > object css > object theme > sheet css
+ * If the result of those is transparent, we use the sheet css color
+ */
+export const getBackgroundColors = (theme: ExtendedTheme, rootElement: HTMLElement): BackgroundColors => {
   const qvInnerObject = rootElement?.closest('.qv-object .qv-inner-object');
-  const objectBackgroundColorFromCSS = qvInnerObject && window.getComputedStyle(qvInnerObject).backgroundColor;
+  const objectColorFromCSS = qvInnerObject && window.getComputedStyle(qvInnerObject).backgroundColor;
 
   const qvPanelSheet = rootElement?.closest('.qv-panel-sheet');
-  const sheetBackgroundColorFromCSS = qvPanelSheet && window.getComputedStyle(qvPanelSheet).backgroundColor;
+  const sheetColorFromCSS = qvPanelSheet && window.getComputedStyle(qvPanelSheet).backgroundColor;
 
-  const tableBackgroundColorFromTheme = theme.getStyle('', '', 'object.straightTable.backgroundColor');
+  // tableColorFromTheme is undefined if nothing is set specifically for object.straightTable.backgroundColor
+  const tableColorFromTheme = theme.getStyle('', '', 'object.straightTable.backgroundColor');
+  // colorFromTheme traverses the theme tree until it finds a value for backgroundColor, is undefined if there is no value
+  const colorFromTheme = theme.getStyle('object', 'straightTable', 'backgroundColor');
 
-  const backgroundColorFromTheme = theme.getStyle('object', 'straightTable', 'backgroundColor');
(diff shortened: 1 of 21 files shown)


src/table/components/styles.ts
```diff
@@ -18,13 +18,14 @@ export const TableAnnouncer = styled('div')({
...
 
 export const StyledTableWrapper = styled(Paper, {
-  shouldForwardProp: (prop: string) => prop !== 'tableTheme' && prop !== 'paginationNeeded',
-})(({ tableTheme, paginationNeeded }) => ({
+  shouldForwardProp: (prop: string) => prop !== 'background' && prop !== 'paginationNeeded',
+})(({ background, paginationNeeded }) => ({
   borderWidth: paginationNeeded ? '0px 1px 0px' : '0px',
   borderStyle: 'solid',
-  borderColor: tableTheme.borderColor,
+  borderColor: background.isDark ? '#F2F2F2' : '#D9D9D9',
```
haxxmaxx, line 25: This is temporary. The wrapper wont have a border when we apply the new styling

src/table/utils/styling-utils.ts
```diff
@@ -166,6 +172,23 @@ export function getBodyCellStyle(layout: TableLayout, theme: ExtendedTheme): Gen
...
+        color: 'rgba(255, 255, 255, 0.9)',
+        iconColor: 'rgba(255, 255, 255, 0.9)',
+        disabledIconColor: 'rgba(255, 255, 255, 0.3)',
+      }
+    : {
+        background: background.color,
+        borderColor: '#D9D9D9',
+        color: '#404040',
+        iconColor: 'rgba(0, 0, 0, 0.54)',
+        disabledIconColor: 'rgba(0, 0, 0, 0.3)',
```
haxxmaxx, line 189: these values will be replaced with sprout values later

src/table/utils/styling-utils.ts
```diff
@@ -166,6 +171,23 @@ export function getBodyCellStyle(layout: TableLayout, theme: ExtendedTheme): Gen
   };
 }
 
+export const getFooterStyle = (background: BackgroundColors) =>
```
a-m-dev, line 174: i'm not sure about the return type of this function but i think it might help inferring the type while passing into styled components

a-m-dev (approved):
just a minor suggestions, looks good otherwise

src/table/components/footer/styles.ts
```diff
@@ -8,8 +8,8 @@ import FormControl from '@mui/material/FormControl';
 // ---------- FooterWrapper ----------
 
 export const StyledFooterWrapper = styled(Paper, {
-  shouldForwardProp: (prop: string) => prop !== 'tableTheme',
-})(({ tableTheme, theme }) => ({
+  shouldForwardProp: (prop: string) => prop !== 'footerStyle',
+})(({ footerStyle, theme }) => ({
```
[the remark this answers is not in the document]
haxxmaxx: yeah I noticed that I get no complaints from ts about not typing this. In qcs-ui-common for example, we do. We probably want to type this, but then we should do it for all places at once IMO

src/nebula-hooks/use-extended-theme.ts
```diff
@@ -1,51 +1,44 @@
...
-    isBackgroundTransparentColor,
-    isBackgroundDarkColor,
-    borderColor,
-    body,
-    pagination,
+    tableColorFromTheme: tableColorFromTheme || 'inherit',
+    color,
+    isDark,
+    isTransparent,
   };
```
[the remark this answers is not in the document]
haxxmaxx: well you need `color` defined when setting isDark and isTransparent. Would still get a mix

src/table/utils/styling-utils.ts
haxxmaxx: ah yes this should have a type, fixing

src/table/components/styles.ts
a-m-dev: I think you can use this function here: https://github.com/qlik-oss/sn-table/blob/d5bb1cf994a1f9f6f69bc96b15f47fd826e002f3/src/table/utils/styling-utils.ts#L51

src/table/utils/styling-utils.ts
```diff
@@ -34,6 +34,7 @@ export function getPadding(styleObj: ContentStyling | undefined): string | undef
  * Gets color from color picker. Defaults to default color if resolved color is invalid
  */
 export function getColor(defaultColor: string, theme: stardust.Theme, color = {}): string {
+  console.log(color);
```
a-m-dev: gotcha 😄

src/table/components/styles.ts
haxxmaxx: yeah, but i will remove it soon. There will be no border at all, that is why I did not bother

src/table/utils/styling-utils.ts
haxxmaxx: This is old, common 😅

PR #764, opened 2022-12-06
chore: add sprout back

haxxmaxx:
Reintroducing the sprout package. This means that you need Qlik access to run `install`. Currently there is no work-around. The rendering test needs to be updtaded. This is an intermitted step though, since there is more styling to come. Better to do that in different PRs and then make sure they are in before we make a release. Also includes:

- Pagination styling. It was completely broken, so updated according to the preliminary design
- Removing the old sprout copy
- Move sanity check for `paginationNeeded` to footer component


haxxmaxx:
the diff is mostly yarn.lock

cbt1:
FYI the sort direction icon is a bit off in its placement when I run this locally

Skärmavbild 2022-12-20 kl  08 23 51

haxxmaxx:
> FYI the sort direction icon is a bit off in its placement when I run this locally
> 
yeah I know. I have the whole head styling almost ready so I figured I fix that in the next PR

package.json
```diff
@@ -72,6 +72,10 @@
     "@babel/preset-typescript": "7.18.6",
     "@commitlint/cli": "17.1.2",
     "@commitlint/config-conventional": "17.1.0",
+    "@material-ui/core": "4.12.4",
+    "@material-ui/icons": "4.11.3",
+    "@material-ui/lab": "4.0.0-alpha.61",
+    "@material-ui/styles": "^4.11.5",
```
cbt1, line 78: Remove `^` and run `yarn`

cbt1 (approved):
Pin the package version as commented and you're good to go

PR #765, opened 2022-12-06
fix: clicking head cell sorts the column when the menu is open

Donya-qlik:
Issue that is fixed:
- Clicking head cell causes changing sort order when the context menu is open

After fix:
- Sorting the column by clicking head cell is not possible as long as the context menu is open


haxxmaxx:
let's close this, we are trying to fix something that works as designed. Sorry about the going full circle on this

Donya-qlik:
> let's close this, we are trying to fix something that works as designed. Sorry about the going full circle on this

Ok, no problem, instead we found one of the solutions we know that will not work! 🙂

PR #767, opened 2022-12-07
fix: sanity check for selections api

haxxmaxx:
`useSelections` can actually initially return null. We need to make sure it is defined when rendering and sanity check for it when checking `isModal`


a-m-dev (approved):
LGTM!

PR #769, opened 2022-12-08
perf: data buffer in virtual table mode

cbt1:


src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
```diff
@@ -43,51 +43,143 @@ const createNewRow = (
   };
 };
 
+const isRowMissingData = (rows: Row[], x: number, y: number, width: number) => {
```
haxxmaxx, line 46: this and the other two functions are pure functions right? Could they be in a util file or something? this file is getting pretty big

src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
```diff
@@ -43,51 +43,143 @@ const createNewRow = (
...
+  // The queue takes a EngineAPI.INxPage object as items and adds them to a queue and
+  // exists to prevent the same page from being fetched more than once.
+  const queue = useGetHyperCubeDataQueue(getDataPages, handleDataPages);
+
+  useOnPropsChange(() => {
+    queue.clear();
+    setRowsInPage([]);
+  }, [layout, pageInfo.page]);
+
+  const loadData: LoadData = useCallback(
```
haxxmaxx, line 122: so this runs only in the last useEffect, right? And it listen to changes on `getDataPages` and  `handleDataPages`. Seems a bit overkill to first wrap this with useCallback, then make that useEffect depend `loadData`, instead of just running the logic inside the useEffect. Or does it have something to do that it is async?

src/table/components/virtualized-table/hooks/use-get-hypercube-data-queue.ts
```diff
@@ -0,0 +1,67 @@
...
+        if (finished.current.has(key)) {
+          return;
+        }
+
+        queued.current.add(page);
+        // Add to finished before it's actually finished. That makes it easier to check if the page should be retrieved or not
+        // as otherwise all three sets (queued, ongoing and finished) have to be queried for presence.
+        finished.current.add(key);
+
+        if (queued.current.size === 1) {
```
haxxmaxx, line 31: ok so trying to understand this. Whenever you run `enqueue,` you also add the passed `page` to `queued`. Then you check if there is only one entry in the queue. Can there be more than one? If so, isn't the rest of the queue skipped? or is it listening to something that makes it run later?

src/table/components/virtualized-table/hooks/use-get-hypercube-data-queue.ts
```diff
@@ -0,0 +1,67 @@
...
+          queueMicrotask(async () => {
+            const qPages = Array.from(queued.current.values());
+
+            queued.current.clear();
+            ongoing.current.add(qPages);
+
+            try {
+              const qDataPages = await getDataPages(qPages);
+
+              if (ongoing.current.has(qPages)) {
```
haxxmaxx, line 41: is this ever false? looks like you add `qPages` then check it right after if it is there?

src/table/components/virtualized-table/hooks/use-get-hypercube-data-queue.ts
```diff
@@ -0,0 +1,67 @@
...
+  getDataPages: (qPages: EngineAPI.INxPage[]) => Promise<EngineAPI.INxDataPage[]>,
+  handleDataPages: (qDataPages: EngineAPI.INxDataPage[]) => void
+) => {
+  const queued = useRef(new Set<EngineAPI.INxPage>()); // Keep track of all unique pages that should be retrieved
+  const ongoing = useRef(new Set<EngineAPI.INxPage[]>()); // Keep track of ongoing request. Should be aborted if page info or layout is changed
+  const finished = useRef(new Set<string>()); // Keep track of all unqiue pages that have added to the queue
+
+  const queue = useMemo(
+    () => ({
+      enqueue: (page: EngineAPI.INxPage) => {
```
haxxmaxx, line 19: so when you run `enqueue` you add to `queued`. Then you pop that into `ongoing`, then eventually you put them in `finished` when the fetching is done?

haxxmaxx (commented):
Code looks good, but it is hard to follow. Maybe need to adjust some things or add comments, depending on what the answers to my questions are 🙃

src/table/components/virtualized-table/hooks/use-get-hypercube-data-queue.ts
cbt1: The general progression of state is `queued` => `ongoing` => `finished`. So 3 different states.

So when you call `enqueue` there are 3 cases that when a page should NOT be retrieved. That is if it's already `queued`, `ongoing` or `finished`.

BUT instead of have to check `queued`, `ongoing` and `finished` if a page is in any of those sets. I added it directly to `finished`, this means `finished` actually contains all 3 states and more importantly, when you check if page should be retrieved, all you have to look at is the `finished` set. There is 1 caveat to this and that is that if the request fails, they are removed from `finished`.

So `finished` is a fast way to check if a page should be retrieved or ignored.

`queued` is needed to keep track of pages that should be retrieved.

`ongoing` is needed to abort a promise, so essentially, when a promise is settled (data pages have been retrieved) it should be handled by the `handleDataPages` callback. But only if it has not been aborted. Example when a "operation" should be aborted is if the user changes page OR a new layout recieved.

src/table/components/virtualized-table/hooks/use-get-hypercube-data-queue.ts
cbt1: It is assumed that the function that is calling `enqueue` is going to call it multiple times. As is the case in `useInfiniteScrollData`where `enqueue` is being called inside a for-loop.

`enqueue` is using the deterministic nature of the javascript event loop to "batch" queued pages.

```js
function task() {
  for (let index = 0; index < 10; index++) {
    // when the first index is passed to enqueue, `queueMicroTask` is called with a function, lets call that function "microTask" 
    enqueue(index);
  }
}

// In the event loop the execution flow would look something like this
task(); // Execute the task function
// and THEN when task() function has exited/finished the queue micro task is executed
microtask(); // Execute the micro task that now have 10 queued indexes as they where all added during the task() execution
```

cbt1: Yes there can and in most cases will be more then 1 queued page.

src/table/components/virtualized-table/hooks/use-get-hypercube-data-queue.ts
cbt1: Yes, `ongoing` is accessable via "closure", but that also means that it can be modified from outside the `enqueue()` function.

Hard to explain as it indeed looks like you commented at first glance. But here we go..

If the call to engine in `getDataPages()` takes some time. I simulated this by replacing it with a promise that resolves after 5 seconds. Making the request reaaaaally slow.. The "call stack" could looke something like this:

```js
loadRows() // finished executing:  enqueued data pages (qTop) 0 to 10
queueMicrotask() //  fetching data pages (qTop) 0 to 10
loadRows() // finished executing:  enqueued data pages (qTop) 11 to 20
queueMicrotask() //  fetching data pages (qTop) 11 to 20
// finished executing the queueMicrotask():  called handleDataPages() for data pages (qTop) 0 to 10
// finished executing the queueMicrotask():  called handleDataPages() for data pages (qTop) 11 to 20
```

But in a scenario when the queue is cleared by calling `queue.clear()`, this happens from `useInfiniteScrollData` when `layout` or `pageInfo` deps changes. The logged "call stack" would look more like this:
```js
loadRows() // finished executing:  enqueued data pages (qTop) 0 to 10
queueMicrotask() //  fetching data pages (qTop) 0 to 10
loadRows() // finished executing:  enqueued data pages (qTop) 11 to 20
queueMicrotask() //  fetching data pages (qTop) 11 to 20
queue.clear(); // was called from "useInfiniteScrollData" because layout or pageInfo changed
// "ongoing.current" set is now empty. So "ongoing.current.has(qPages)" is false.
// handleDataPages() is never called
```

src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
cbt1: Agree, they can be moved

src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
```diff
@@ -43,51 +43,143 @@ const createNewRow = (
...
+
+  if (!targetRow) {
+    return true; // Row is not cached
+  }
+
+  for (let colIndex = x; colIndex < x + width; colIndex++) {
+    if (!targetRow[`col-${colIndex}`]) {
+      return true; // Column is not cached
+    }
+  }
```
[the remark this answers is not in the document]
cbt1: Yes exactly.

A row might look like this `{ col-0, col-1, col-5, col-8 }` in which case it would be awkward at best to find all missing cells and only fetch them.

src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
```diff
@@ -43,51 +43,143 @@ const createNewRow = (
...
+
+  return false;
+};
+
+const isColumnMissingData = (rows: Row[], x: number, y: number, height: number) => {
+  const targetRows = rows.slice(y, height);
```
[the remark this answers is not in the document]
cbt1: I would say the work the same, but this look in the `y` direction while the other looks in the `x` direction.

src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
cbt1: No we can do as you say, will move it. It so little code that there is no longer any reason to have it's own function.

src/table/components/virtualized-table/hooks/use-scroll-direction.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const useScrollDirection = () => {
+  const prevScrollLeft = useRef(0);
+  const prevScrollTop = useRef(0);
+  const scrollDirection = useRef(ScrollDirection.Down);
+
+  const scrollHandler = useCallback(
+    ({ scrollTop, scrollLeft }: OnScroll) => {
+      if (scrollTop > prevScrollTop.current) {
```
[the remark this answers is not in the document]
cbt1: Actually I think you are on to something. I will probably need to keep track both vertical and horizontal scroll direction. So a scroll could potentially trigger both a "loadRows" and a "loadColumns" call at the same time.

Will do some more testing on this.

src/table/components/virtualized-table/hooks/use-table-count.ts
```diff
@@ -0,0 +1,32 @@
+import { stardust } from '@nebula.js/stardust';
```
cbt1, line 1: Refactor. Just move this piece of logic to a custom hook.

src/table/components/virtualized-table/hooks/use-infinite-scroll-data.ts
haxxmaxx: great, helps readability a bit

src/table/components/virtualized-table/hooks/use-scroll-direction.ts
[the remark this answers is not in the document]
cbt1: Fixed and confirmed that it works when simulating a touch device

src/table/components/virtualized-table/Body.tsx
```diff
@@ -32,69 +24,43 @@ const Body = (props: BodyProps) => {
...
+    rowCount,
+    pageInfo,
+  });
 
-  useEffect(() => {
-    // Initial data load
-    const top = pageInfo.page * pageInfo.rowsPerPage;
-    loadData(0, top, visibleColumnCount, visibleRowCount);
-  }, [layout, visibleRowCount, visibleColumnCount, loadData, pageInfo]);
+  const itemData = useMemo(() => ({ rowsInPage, columns, bodyStyle }), [rowsInPage, columns, bodyStyle]);
```
haxxmaxx, line 54: does this make a difference? Something about `VariableSizeGrid` not rerendering if `itemData` is unchanged?

haxxmaxx (approved):

src/table/components/virtualized-table/hooks/use-scroll-direction.ts
```diff
@@ -0,0 +1,55 @@
...
+      } else if (scrollLeft < prevScrollLeft.current) {
+        horizontalScrollDirection.current = ScrollDirection.Left;
+      } else {
+        horizontalScrollDirection.current = ScrollDirection.None;
+      }
+
+      prevScrollTop.current = scrollTop;
+      prevScrollLeft.current = scrollLeft;
+    },
+    [prevScrollTop, prevScrollLeft, verticalScrollDirection, horizontalScrollDirection]
```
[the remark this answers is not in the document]
cbt1: No, a new function is only created when the `refs` are created

src/table/components/virtualized-table/Body.tsx
cbt1: Have a vague memory of it being required to make the react-window `areEqual` function work properly. No I cannot find it any more 😕

src/table/components/virtualized-table/hooks/use-scroll-direction.ts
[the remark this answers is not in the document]
haxxmaxx: ah yes, the object is never changed

PR #772, opened 2022-12-12
feat: disable sortItem when column sort direction is the same

Donya-qlik:
Menu sort option will be disabled if the current column is already ordered in the same sort direction.
In the case that the sorting priority is not on the current column, both sort options stays enable until one of them be chosen.

### DIFF
--- a/src/table/utils/handle-click.ts
+++ b/src/table/utils/handle-click.ts
@@ -37,7 +37,10 @@ export const handleClickToSort = (
   changeSortOrder: ChangeSortOrder,
   isInteractionEnabled: boolean
 ) => {
-  !(evt.target as HTMLElement).closest('#sn-table-head-menu-button') && isInteractionEnabled && changeSortOrder(column);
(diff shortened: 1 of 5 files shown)


src/table/utils/handle-click.ts
```diff
@@ -37,7 +37,10 @@ export const handleClickToSort = (
   changeSortOrder: ChangeSortOrder,
   isInteractionEnabled: boolean
 ) => {
-  !(evt.target as HTMLElement).closest('#sn-table-head-menu-button') && isInteractionEnabled && changeSortOrder(column);
+  !(evt.target as HTMLElement).getElementsByClassName('sn-table-head-menu-item-button')[0]?.ariaDisabled &&
```
haxxmaxx, line 40: Unfortunately we need something like this for now. But the click event listener will move to the the header text only, not the cell, then we can remove these

haxxmaxx (approved):
LGTM :)

PR #773, opened 2022-12-13
Dmh/sn table selection mode no sort

Donya-qlik:


src/table/components/head/styles.ts
```diff
@@ -21,15 +21,16 @@ export const StyledHeadRow = styled(TableRow, {
...
+})(({ headerStyle, isInteractionEnabled }) => ({
   ...COMMON_CELL_STYLING,
   ...headerStyle,
   lineHeight: '150%',
   pointer: 'cursor',
   borderWidth: '1px 1px 1px 0px',
   '&&:hover, &&:focus': {
-    '& svg, & button': { opacity: 1 },
+    '& button': { opacity: 1 },
+    '& svg': { opacity: !isInteractionEnabled ? 0 : 1 },
```
cbt1, line 33: There is no reason to use the logical NOT operator here
```suggestion
    '& svg': { opacity: isInteractionEnabled ? 1 : 0 },
```

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -30,9 +31,13 @@ const MenuItem = ({
   sortDirection: string;
   sortFromMenu(evt: React.MouseEvent, sortOrder: string): void;
   setOpen: any;
+  isInteractionEnabled: boolean;
   isCurrentColumnActive: boolean;
 }) => {
-  const isDisabled = isCurrentColumnActive && sortOrder === (sortDirection === 'asc' ? 'A' : 'D');
+  let isDisabled = false;
+  isDisabled =
+    (!isInteractionEnabled ? !isDisabled : isDisabled) ||
```
haxxmaxx, line 39: So this pattern is never a good idea. There is no reason to set in inital state, then change it right after. The double negation here makes it hard to follow, but I think this should be the same
```suggestion
  const isDisabled = !isInteractionEnabled ||
```

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -30,9 +31,13 @@ const MenuItem = ({
   sortDirection: string;
   sortFromMenu(evt: React.MouseEvent, sortOrder: string): void;
   setOpen: any;
+  isInteractionEnabled: boolean;
   isCurrentColumnActive: boolean;
 }) => {
-  const isDisabled = isCurrentColumnActive && sortOrder === (sortDirection === 'asc' ? 'A' : 'D');
+  let isDisabled = false;
```
cbt1, line 37: This logic seems unnecessary complex. A few remarks:
- Why not use `const isDisabled = ...`?
- Is `!isInteractionEnabled ? !isDisabled : isDisabled` not the same as `!isInteractionEnabled` ?
- Why is `sortOrder` defined as `'A' | 'B'` but sortDirection is `'asc' | 'desc'`?

src/table/components/head/HeadCellMenu.tsx
Donya-qlik: .. but I want to turn disable to true when `isInteractionEnabled` is false

src/table/components/head/TableHeadWrapper.tsx
```diff
@@ -101,12 +102,13 @@ function TableHeadWrapper({
                     </VisuallyHidden>
                   )}
                 </StyledSortLabel>
-                {areBasicFeaturesEnabled && (
+                {true && (
```
Donya-qlik, line 105: I'll fix this, it's just for testing purpose

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: that is exactly what the `!` operator does

src/table/components/head/styles.ts
Donya-qlik: Hiding arrow icon when it is in selection mode, but it hides the button as well which I don't want :(

src/table/components/head/HeadCellMenu.tsx
Donya-qlik: - For your first couple of questions you are right,  sorry I did a silly mistake here... I was thinking of `!isInteractionEnabled` as false but it's a conditional check that toggles the variable.
- column qSortIndicator in hypercube is defined in A, D,.. format. It is used in useSorting hook that also checks the menu item sortOrder somewhere in a condition, so both of them are set in a same format. Here sortDirection is a prop comes from the headCell parent and it is set in `asc`/`desc` format ... so I should find a way to convert it.

haxxmaxx: I think we only need one of them (asc/desc or A/D). I would pass asc/desc then pass `sortDirection.charAt(0).toUpperCase()`

cbt1:
Not really for this PR but both sortOrder and sortDirection appears to be strings that are well defined (A/D and asc/desc). That should be defined like so in typescript definition to, as it would make it obvious what string values are possible.

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: I just looked into this. We need three versions of the sortDirections
- engine stores it using A/D and sort backend call expects it too
- mui `TableSortLabel` wants asc/desc as property to keep track of the icon (I have looked into just doing this manually and stop using that component, could be risky though)
- `aria-sort` needs it to be ascending/descending.
So I guess the typing should be better to reduce the confusion. I think the cleanest solution would be to store A/D on the column data. Then use enums to get asc and ascending, because I think they would only be used in TableHeadWrapper. Not sure how that impacts the virtual scrolling solution

Donya-qlik: I tried your previous solution and it seems is working, have few conflicts need to fix them.  I also used Christofer's idea for setting a type to know which strings are accepted. Do you think you want them to be written with enum instead ?

haxxmaxx: I think we can skip the enums, that would be an improvement for later

src/table/components/head/styles.ts
```diff
@@ -21,15 +21,16 @@ export const StyledHeadRow = styled(TableRow, {
...
+})(({ headerStyle, isInteractionEnabled }) => ({
   ...COMMON_CELL_STYLING,
   ...headerStyle,
   lineHeight: '150%',
   pointer: 'cursor',
   borderWidth: '1px 1px 1px 0px',
   '&&:hover, &&:focus': {
-    '& svg, & button': { opacity: 1 },
+    '& button': { opacity: 1 },
+    '& svg': { opacity: isInteractionEnabled ? 1 : 0 },
```
Donya-qlik, line 33: I'll revert this since it is not fixable in the current PR.

haxxmaxx, line 33: revert this

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -30,16 +31,18 @@ const MenuItem = ({
...
+  isDisabled = !isInteractionEnabled || (isCurrentColumnActive && sortOrder === sortDirection);
   return (
     <ListItem disablePadding>
       <ListItemButton
         disabled={isDisabled}
         className="sn-table-head-menu-item-button"
         onClick={(evt) => {
-          sortFromMenu(evt, sortOrder);
+          sortFromMenu(evt, sortOrder.charAt(0).toUpperCase());
```
haxxmaxx, line 45: you are doing this in two places. If you revert it here it makes sense with the reset

haxxmaxx (approved):
This looks fine, will do it a bit differently in the refactor I have on the way, but this solution should work

src/table/components/head/HeadCellMenu.tsx
Donya-qlik: Fixed

PR #775, opened 2022-12-13
refactor: head cell menu prop types

Donya-qlik:


src/table/types.ts
```diff
@@ -226,6 +227,15 @@ export interface TableHeadWrapperProps extends CommonTableProps {
   areBasicFeaturesEnabled: boolean;
 }
 
+export interface TableHeadCellMenu {
```
haxxmaxx, line 230: name it so it matches the component

haxxmaxx (approved):

PR #776, opened 2022-12-13
refactor: align sort format and remove hyhpercube dependency

haxxmaxx:
- the current sort direction already exists on the column, might as well use that. Still need the number of dimension to get the right path to the field when we call engine
- Store sorting direction as  `A` or `D`. then use enums to get the asc/desc that mui needs + the full word for aria-sort

### DIFF
--- a/src/nebula-hooks/use-sorting.ts
+++ b/src/nebula-hooks/use-sorting.ts
@@ -1,10 +1,10 @@
 import { useMemo } from '@nebula.js/stardust';
-import { Column, HyperCube } from '../types';
+import { Column, HyperCube, SortDirection } from '../types';
 
 export const sortingFactory = (model: EngineAPI.IGenericObject | undefined, hyperCube: HyperCube) => {
   if (!model) return undefined;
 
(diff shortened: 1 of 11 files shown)


src/nebula-hooks/use-sorting.ts
```diff
@@ -46,5 +41,5 @@ export const sortingFactory = (model: EngineAPI.IGenericObject | undefined, hype
 };
 
 const useSorting = (model: EngineAPI.IGenericObject | undefined, hyperCube: HyperCube) =>
-  useMemo(() => sortingFactory(model, hyperCube), [model, hyperCube]);
+  useMemo(() => sortingFactory(model, hyperCube.qDimensionInfo.length), [model, hyperCube]);
```
cbt1, line 44:
```suggestion
  useMemo(() => sortingFactory(model, hyperCube.qDimensionInfo.length), [model, hyperCube.qDimensionInfo.length]);
```

cbt1 (approved):
LGTM

PR #777, opened 2022-12-14
feat: styling for header and body

haxxmaxx:
So this should be all the base styling for the table itself. A lot of stuff going on. It is partly because this PR living for long time, while other changes have been made.
- styling according to designs
- new function `getBorderColors`, border colors depend on totals position now
- new function `getLastRowBottomBorder` to estimate if there should be a border bottom on the last row. Just estimating since we don't know if there there are any rows with multiple lines
- remove padding calculations since they are static now. Only the line-height scales with the font size
- using `CellText` in virtual scroll
- using `background` instead of `backgroundColor` for consistency. Updated a lot of variable names to align with this, especially for the hover color logic
- some smaller refactors and typing stuff. Maybe could have been descoped, but some where more or less necesarry
- icon for head menu

not included, that I will do next
- using color values from sprout
- replace all icons to sprout
- replace head menu with sprout Menu
- make sure bottom border for virtual scroll is correct


src/table/components/body/TableBodyWrapper.tsx
```diff
@@ -40,9 +40,12 @@ function TableBodyWrapper({
       ),
     [columnsStylingIDsJSON, isSelectionsEnabled]
   );
-  const bodyCellStyle = useMemo(() => getBodyCellStyle(layout, theme), [layout, theme]);
+  const { hoverColors, lastRowBottomBorder, ...cellStyle } = useMemo(
```
cbt1, line 43: Are you using spread (...) on `cellStyle` so that `hoverColors` or `lastRowBottomBorder` will be overwritten if they are `undefined`?

haxxmaxx: I want do split them up into three different variables, so that we can use cellStyle for the cell, then the others are applied in other places

cbt1: 👍

src/table/utils/styling-utils.ts
```diff
@@ -42,34 +27,68 @@ export function getColor(defaultColor: string, theme: stardust.Theme, color = {}
...
+  };
+};
+
+/**
+ * Get border widths for body. adds a bottom border if the rendered rows height is estimated to be greater than the container height
+ */
+const getLastRowBottomBorder = (fontSize: string | undefined, rowsLength?: number, rootElement?: HTMLElement) => {
+  if (fontSize && rowsLength && rootElement) {
+    // add line height (133%), padding and borders
+    const rowHeight = (+fontSize.slice(0, -2) * 4) / 3 + 9;
```
cbt1, line 63: Use `parseInt(fontSize, 10)` instead of slice?

haxxmaxx: that removes px?

haxxmaxx: it does, nice

src/table/utils/styling-utils.ts
```diff
@@ -42,34 +27,68 @@ export function getColor(defaultColor: string, theme: stardust.Theme, color = {}
...
-    color: isColorSet(styleObj?.fontColor) ? getColor(StylingDefaults.FONT_COLOR, theme, styleObj?.fontColor) : color,
+    color: isPaletteColorSet(styleObj?.fontColor)
+      ? getColor(StylingDefaults.FONT_COLOR, theme, styleObj?.fontColor)
+      : color,
     fontSize: (styleObj?.fontSize && `${styleObj.fontSize}px`) || fontSize,
-    // When we do not set padding for content or header, but set font size,
-    // we need to calculate the padding based on the font size
-    padding: getPadding(styleObj),
-    borderColor: getBorderColor(theme.background.isDark),
+    padding: styleObj && 'padding' in styleObj ? styleObj?.padding : undefined,
```
cbt1, line 90: Isn't `padding: styleObj?.padding` enough?

src/table/utils/styling-utils.ts
a-m-dev, line 63: 3 + 9 seems to be a magic number, is it about some padding values?

yarn.lock
```diff
@@ -484,11 +484,16 @@
     chalk "^2.0.0"
     js-tokens "^4.0.0"
 
-"@babel/parser@^7.1.0", "@babel/parser@^7.13.16", "@babel/parser@^7.14.0", "@babel/parser@^7.14.7", "@babel/parser@^7.18.10", "@babel/parser@^7.18.11", "@babel/parser@^7.9.4":
+"@babel/parser@^7.1.0", "@babel/parser@^7.13.16", "@babel/parser@^7.14.0", "@babel/parser@^7.14.7", "@babel/parser@^7.18.10", "@babel/parser@^7.9.4":
```
cbt1, line 487: No update in package.json but yarn.lock has changed?

src/table/utils/styling-utils.ts
```diff
@@ -42,34 +27,68 @@ export function getColor(defaultColor: string, theme: stardust.Theme, color = {}
...
+
+/**
+ * Get border widths for body. adds a bottom border if the rendered rows height is estimated to be greater than the container height
+ */
+const getLastRowBottomBorder = (fontSize: string | undefined, rowsLength?: number, rootElement?: HTMLElement) => {
+  if (fontSize && rowsLength && rootElement) {
+    // add line height (133%), padding and borders
+    const rowHeight = (+fontSize.slice(0, -2) * 4) / 3 + 9;
+    // multiply with number of rows plus header and totals. Compare if greater than container
+    const showBottomBorder = rowHeight * (rowsLength + 2) < rootElement.offsetHeight - PAGINATION_HEIGHT;
```
a-m-dev, line 65: since offsetHeight is considering the borders, if we have borderTop on table doe we want to consider it in calculation? 
as Alternate client height could be used maybe if it is the case? 

I'm just thinking out loudly here 😅

a-m-dev (approved):
Some minor comments, otherwise LGTM 🚀

src/table/components/virtualized-table/Cell.tsx
```diff
@@ -55,16 +56,7 @@ const Cell = ({ columnIndex, rowIndex, style, data }: CellProps) => {
...
-          className="sn-table-cell-text"
-          style={{
-            overflow: 'hidden',
-            whiteSpace: 'nowrap',
-            textOverflow: 'ellipsis',
-          }}
-        >
-          {datum.qText}
-        </span>
+        <CellText singleLine>{datum.qText}</CellText>
```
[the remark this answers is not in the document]
cbt1: It's a name for a single data point a set of data. But I've renamed it i a coming PR to something more suitable.

src/table/components/virtualized-table/Cell.tsx
```diff
@@ -45,7 +46,7 @@ const Cell = ({ columnIndex, rowIndex, style, data }: CellProps) => {
           borderRightWidth: datum.isLastColumn ? '0px' : '1px',
           borderStyle: 'solid',
           justifyContent: columns[columnIndex].align,
-          padding: '0px 14px',
+          padding: '4px 12px',
```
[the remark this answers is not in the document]
haxxmaxx: Well not sure what Christoffers plan is here. If we make these into styled components we should defenetly do that. For now I think this is fine

src/table/utils/styling-utils.ts
haxxmaxx: well that's not order of operations work ;) I don't like what the formatter does here, it makes it unclear. This is what happens. pixels * 4/3 + 9. 4/3 is 133% but more exact, and the paddding is 4px top and bottom plus bottom border. I could clarify it in the comment

src/table/utils/styling-utils.ts
haxxmaxx: Well this is an estimation anyway. We don't have access to the header font size, and I don't intend to add it here. Simply because this calculation is for pagination, which wont be used as much. As I understand it, we will be able to do it exactly for the scrolling solution

yarn.lock
haxxmaxx: yeah I had some building issues so I removed yarn.lock. When it rebuilds from scratch you can apparently get a different result

src/table/utils/styling-utils.ts
haxxmaxx: yeah it is a typing thing, because `styleObj: HeaderStyling | ContentStyling | undefined,`
padding only exists on contentStyling. Maybe this can be done in a cleaner way, but it was already in the `getPadding` function, so nothing new

src/table/utils/styling-utils.ts
a-m-dev: Yes, Thanks for clarification, 
I was wondering more about numbers 3 and 9 itself but now it is super clear specially for 9 which is the combination of three things ✅

haxxmaxx: actually I could move some brackets around and it is much clearer

src/table/utils/styling-utils.ts
haxxmaxx: but you are right actually. Client height is better, I'll change

cbt1 (approved):
I like it, a lot

PR #784, opened 2022-12-22
fix: borders in virtual table

cbt1:
Fixes a number of issue with borders and some simplifies some calcuations around width and height.

A simplified example of how the DOM structure is setup to enable scrolling and element width/height can be found here: https://jsfiddle.net/vLsfjmzh/1/

Before:
Skärmavbild 2022-12-23 kl  11 47 46
Skärmavbild 2022-12-23 kl  11 47 58

After:
Skärmavbild 2023-01-10 kl  09 01 39
Skärmavbild 2023-01-10 kl  09 23 28
Skärmavbild 2023-01-10 kl  09 23 41


src/table/components/footer/FooterWrapper.tsx
```diff
@@ -4,12 +4,14 @@ import { StyledFooterWrapper } from './styles';
...
 
-export default function FooterWrapper({ children, theme, footerContainer }: FooterWrapperProps) {
+export default function FooterWrapper({ children, theme, footerContainer, withoutBorders }: FooterWrapperProps) {
   const footerStyle = useMemo(() => getFooterStyle(theme.background), [theme]);
 
   return footerContainer ? (
     ReactDOM.createPortal(children, footerContainer)
   ) : (
-    <StyledFooterWrapper footerStyle={footerStyle}>{children}</StyledFooterWrapper>
+    <StyledFooterWrapper footerStyle={footerStyle} withoutBorders={withoutBorders}>
```
cbt1, line 13: Virtual table does not have borders around the footer so had to make them optional

src/table/components/virtualized-table/ScrollableContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const ScrollableContainer = ({
+  children,
+  width,
+  height,
+  forwardRef,
+  constraints,
+  onScroll,
+}: ScrollableContainerProps): JSX.Element => {
```
a-m-dev, line 20: Shouldn't we use `React.forwardRef(...)` syntax here?

src/table/components/virtualized-table/types/index.ts
```diff
@@ -13,6 +13,15 @@ import { Totals } from '../hooks/use-totals';
...
 export type TotalsPosition = 'bottom' | 'noTotals' | 'top';
 
+interface Rect {
+  width: number;
+  height: number;
+}
+
+export interface TableRect extends Rect {}
+
+export interface StickyContainerRect extends Rect {}
```
a-m-dev, line 23: are these two necessary at this step? because they dont consider other types so cant't we use React instead?

src/table/components/virtualized-table/ScrollableContainer.tsx
cbt1: Agree, will fix

src/table/components/virtualized-table/types/index.ts
cbt1: I had made a comment about this when creating the PR but appearently it ended up in a "pending" state.

Here it is:
"May seem a bit overkill to have two different interfaces with the same props. But I wanted to make a distinction that the actual values are different. Maybe there is a better way to do that?"

But looking at this again now. I think it fair to just use `rect`, as there are two different variables with proper names that describes the difference.

src/table/components/footer/FooterWrapper.tsx
haxxmaxx: but is this still the case? I think it should always have a top border now, for both pagination and vs

src/table/components/virtualized-table/Table.tsx
```diff
@@ -71,48 +75,46 @@ const TableContainer = (props: TableContainerProps) => {
...
+        </StickyContainer>
       </FullSizeContainer>
-    </div>
+    </ScrollableContainer>
   );
 };
 
 // Export non memoized version for testing purpose
-export { TableContainer };
+export { Table as TestableTable };
```
haxxmaxx, line 118: why the casting?

haxxmaxx (approved):
Some questionmarks, but lets get it in and I can finalize  the styling

src/table/components/footer/FooterWrapper.tsx
cbt1: No we (you) should change this as you say

src/table/components/virtualized-table/Table.tsx
cbt1: It is not casted but exported as a non memoized version for testing purpose

PR #788, opened 2023-01-03
feat: Search feature in columns dropdown menu

a-m-dev:
# This PR Includes:
- [x] reverting design changes ([current design's](https://www.figma.com/file/thmnnMr0JUftX8K2xl7uz6/PS-15585-More-basic-properties-for-sn-table?node-id=2068%3A42605&t=BQMwf04itNPTCyxB-0) implementation is in [this closed PR](https://github.com/qlik-oss/sn-table/pull/786)) but we need to hold it for more discussions on weather using existing sprout    component or implementing the custom one like mentioned PR
- [x] listbox integration for search
- [x] filtering by listbox
- [x] everything is behind `PS_18291_SN_TABLE_BASIC_FEATURES` ff 
- [x] unit tests

# Not in this PR:
- [ ] list box actions like select, deselect, ...
- [ ] list box search through dimension


haxxmaxx (commented):
tested it and found a couple of things that didn't work
- should close search popover when confirming 
- should close when canceling
- if you open the ...-menu from inside the listbox popover it is in the wrong place
- you get an empty listbox if you open the listbox for a measuere
I think this is imortant functionality we need before merging. Fix that and scope out the submenu code and I will take another look :)

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,110 +1,142 @@
...
-          <ListItemText primary={itemTitle} />
-        </ListItemButton>
-      </ListItem>
-    );
-  };
+  const getMenuItems = useMemo<HeadCellMenuGroup[]>(
+    () => [
+      {
+        id: 1,
+        menus: [
```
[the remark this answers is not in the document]
a-m-dev: These are group of menus, i had a bigger scale for this use case regarding submenus in multiple level (recursively iterating over this object and rendering `<MenuGroup />` component for each subset) but right now since i removed that part, the only use case for this is for determining when to show the group divider component. 

btw, do you mean changing the `menus` to `options`?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,110 +1,149 @@
...
-            setOpen(false);
-          }}
-        >
-          <ListItemIcon sx={{ minWidth: '25px' }}>{children}</ListItemIcon>
-          <ListItemText primary={itemTitle} />
-        </ListItemButton>
-      </ListItem>
-    );
-  };
+  const getMenuItems = useMemo<HeadCellMenuGroup[]>(
```
cbt1, line 32: Hmm the name `getMenuItems` seem to suggest that it's a function but the actual value is an array. I suggest you changed the name to `menuItems`

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,110 +1,149 @@
...
                   aria-labelledby="sn-table-head-menu-button"
                 >
-                  <MenuItem itemTitle={translator.get('SNTable.MenuItem.SortAscending')} newSortDirection="A">
-                    <ArrowUpwardIcon />
-                  </MenuItem>
-
-                  <MenuItem itemTitle={translator.get('SNTable.MenuItem.SortDescending')} newSortDirection="D">
-                    <ArrowDownwardIcon />
-                  </MenuItem>
+                  {Object.entries(getMenuItems).map(([key, menuGroup]) => (
```
cbt1, line 114: Why use `Object.entries` here?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,110 +1,149 @@
...
+                    <MenuGroup
+                      key={key}
+                      {...menuGroup}
+                      shouldShowDevider={Number(key) !== Object.entries(getMenuItems).length - 1}
```
cbt1, line 118: Why use `Object.entries` here?

src/table/components/head/ListBoxWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (!ref.current || !listboxInstance) return undefined;
+
+    listboxInstance.mount(ref.current, { title: ' ' });
+
+    return () => {
+      listboxInstance.unmount();
+    };
+  }, [ref.current, listboxInstance]);
+
+  return children({ ref, listboxInstance });
```
cbt1, line 41: I don't see `listboxInstance` being used outside of this component? Is there some magic going on or can it be removed?

src/table/components/head/ListBoxWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+    embed.field(layout.qHyperCube.qDimensionInfo[columnIndex]?.qFallbackTitle).then((instance) => {
+      setListboxInstance(instance);
+    });
+  }, []);
+
+  // mounting
+  useEffect(() => {
+    if (!ref.current || !listboxInstance) return undefined;
+
+    listboxInstance.mount(ref.current, { title: ' ' });
```
cbt1, line 34: Why `title` with only a whitespace character?

src/table/components/head/ListBoxWrapper.tsx
a-m-dev: no magic, just removed it

src/table/components/head/ListBoxWrapper.tsx
a-m-dev: apparently, Listbox instance accepts an options while mounting but i've checked the `index.d.ts` files for it and there was an optional title(`title?: string`).

what we need here was that we did not want to show the title of columns there, and seems like either of passing `null` or `undefined` or even not passing title at all will still fallback into showing it anyway. So empty string would work in this case. Also i could not find any flag like `showTitle` or something like that to set it `false` in order to ignore the title.

haxxmaxx: a comment it in the code for that would be nice

cbt1: Seems like a solid workaround to me, but as Max said, a comment for it would be nice

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,110 +1,145 @@
...
+                    <MenuGroup key={menuGroup.id} {...menuGroup} shouldShowDevider={i !== menuItems.length - 1} />
+                  ))}
                 </MenuList>
               </ClickAwayListener>
+            </PrimaryDropdownPaper>
+          </Grow>
+        )}
+      </Popper>
+
+      <Popper
```
haxxmaxx, line 124: maybe move this to the listbox wrapper? Then you wouldn't have to pass the ref like you do here.

a-m-dev: In my opinion no, for a few reasons: 
1. That would make us to pass so many props into ListBox Wrapper which might not be necessary at all
2. if we do so, List box wrapper has to handle the Popper stuff/logic as well
3. This is a common pattern in react which currently reduces a lot's of overhead in implementation phase (the most important one was the replacement of the previous hook and separating the concern of listbox from it's parent dropdown) by only isolating the embed/mount/unmount logic of only listbox

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,110 +1,145 @@
...
-          <Grow
-            {...TransitionProps}
-            style={{
-              transformOrigin: placement === 'bottom-start' ? 'left top' : 'left bottom',
-            }}
-          >
-            <Paper sx={{ boxShadow: 15 }}>
-              <ClickAwayListener onClickAway={() => setOpen(false)}>
+        {({ TransitionProps }) => (
+          <Grow {...TransitionProps} style={{ transformOrigin: 'left top' }}>
```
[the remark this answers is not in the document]
a-m-dev: Because there was no need for it at the first place, since the placement is explicitly been set to `        placement="bottom-start"` in a few lines above in `<Popper />` component and that placement is always going to give us the value of "bottom-start" (as I tested several times in different viewport sizes and different datasets). 

So the value is not going to change and that check is going to be unnecessary.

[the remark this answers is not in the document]
haxxmaxx: ok great :)

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: can't argue with that, let's keep it here

haxxmaxx (approved):
This looks fine for now. We need to get some deciscions on what to do with the menu itself, before it becomes blocking and/or confusing. great work :)

PR #792, opened 2023-01-10
perf: move hoverIndex to context

haxxmaxx:


src/table/components/virtualized-table/utils/get-cell-style.ts
```diff
@@ -2,10 +2,13 @@ import { SelectionStates, SELECTION_STYLING, StylingDefaults } from '../../../co
 import { BodyStyle } from '../types';
 
 const getCellStyle = (
-  isHoveringOnRow: boolean,
+  isRowHovered: boolean,
+  showHoverEffect: boolean,
   cellSelectionState: SelectionStates,
   bodyStyle: BodyStyle
 ): BodyStyle => {
+  const isHoveringOnRow = isRowHovered && showHoverEffect;
```
haxxmaxx, line 10: naming should probably be somethign like
```suggestion
  const showHoverEffetxt = isRowHovered && isHoverEnabled;
```

cbt1 (approved):

a-m-dev:
wouldn't be fine just keeping the hook and moving it's initialisation into context level? 

That way you would prevent spreading the logic into context and provide the same data from hook result into your context tree as usual

PR #795, opened 2023-01-12
refactor: extend table context usage

cbt1:
A prototype where we use React Context to avoid prop-drilling some base properties (nebula props) that are shared by multiple components.

This version has only made changes and virtual scroll mode and is for now just a PoC.

I looked a bit at performance and how many times each component renders in comparision with the solution on `main` and so far there is no obvious difference. The gain seem to be a cleaner "tree" for component properties.


src/table/types.ts
```diff
@@ -78,6 +78,15 @@ export interface ContextValue {
   selectionDispatch: SelectionDispatch;
   hoverIndex: number;
   setHoverIndex: React.Dispatch<React.SetStateAction<number>>;
+  baseProps: {
+    selectionsAPI: ExtendedSelectionAPI;
+    layout: TableLayout;
+    model?: EngineAPI.IGenericObject;
```
haxxmaxx, line 84: I think we should make sure model is defined in the index file

haxxmaxx (commented):
Looks good! Are there any additional renders? Could you try it for the pagination as well? 😇

src/table/types.ts
cbt1: I had to make it optional as the paginiation table does not pass the model

cbt1:
> Looks good! Are there any additional renders? Could you try it for the pagination as well? 😇

Not that I could see from my testing. Performance wise feelt the same but some actual number on that would be nice

src/table/types.ts
haxxmaxx: ah I see, then we shouldn't change anything in the index 🙃

haxxmaxx:
> > Looks good! Are there any additional renders? Could you try it for the pagination as well? 😇
> 
> Not that I could see from my testing. Performance wise feelt the same but some actual number on that would be nice

Is it possible to use the profiler for that?

cbt1:
> Looks good! Are there any additional renders? Could you try it for the pagination as well? 😇

I can, but not in this PR. As that would be to much of a change I think. There are going be a few tests that needs to be wrapped in a provider now, which is a bit of a downside.

src/table/components/virtualized-table/__tests__/Table.spec.tsx
```diff
@@ -14,32 +15,24 @@ describe('<Table />', () => {
   let pageInfo: PageInfo;
   let paginationNeeded: boolean;
   let model: EngineAPI.IGenericObject;
-  let theme: ExtendedTheme;
+  // let theme: ExtendedTheme;
```
a-m-dev, line 18: ops!

jest.config.js
```diff
@@ -4,4 +4,5 @@ module.exports = {
   setupFilesAfterEnv: ['./jest.setup.js'],
   modulePathIgnorePatterns: ['<rootDir>/mashup-example'],
   transformIgnorePatterns: ['<rootDir>/node_modules/(?!(@qlik-trial)/)'],
+  coveragePathIgnorePatterns: ['src/__test__/generate-test-data.ts', 'src/__test__/test-with-providers.tsx']
```
cbt1, line 7: We should move the test utils files to some other location but I'll do that in another PR as I dont want all those changed test files in this PR as well

a-m-dev (approved):
LGTM 🚀

haxxmaxx:
is the description still accurate, that this is considered a PoC/prototype?

cbt1:
> is the description still accurate, that this is considered a PoC/prototype?

No, I've updated it now if someone would ever want to go back and look at this PR.

PR #797, opened 2023-01-13
ci: use workflows to run jobs in parallel

cbt1:
This should improve the run time and make so that you can see in your PR which jobs are failing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,7 @@ jobs:
             - dist
             - package.json
             - core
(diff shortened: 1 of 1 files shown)


haxxmaxx (approved):
nice! I'm surprised it wasn't that much faster, but worth it anyway imo

cbt1:
> nice! I'm surprised it wasn't that much faster, but worth it anyway imo

Yeah the slow part is actually the yarn cache

PR #798, opened 2023-01-13
refactor: use sprout colors

haxxmaxx:
- use colors from sprout for all colors
- make sure there is a `color` and a `font-size` for all cells
- remove `sortLabelColor`, it should just be the same as `color` (will look a bit further into the header including the menu in next pr)
- fixing some dark mode stuff on the pagination

The text color in the inputs where actually wrong, that is why I needed to update the snapshots


cbt1 (approved):
Have not look at the actual output but from a code spective it LGTM

a-m-dev (approved):

PR #800, opened 2023-01-16
fix: listbox inline selection actions

a-m-dev:
This PR is about Selections actions fix,
but wait for this: https://github.com/qlik-oss/nebula.js/pull/1053
👆 should be merged in nebula and synced in this PR


a-m-dev:
because of git history

PR #801, opened 2023-01-16
fix: listbox inline selection actions

a-m-dev:
This PR is about Selections actions fix,
but wait for this: https://github.com/qlik-oss/nebula.js/pull/1062
👆 should be merged in nebula and synced in this PR


src/table/components/head/HeadCellMenu.tsx
```diff
@@ -72,19 +72,31 @@ export default function HeadCellMenu({
...
+          tabIndex={-1}
+          id="sn-table-head-menu-button"
+          aria-controls={openMenuDropdown ? 'sn-table-head-menu' : undefined}
+          aria-expanded={openMenuDropdown ? 'true' : undefined}
+          aria-haspopup="true"
+          onClick={() => setOpenMenuDropdown(!openMenuDropdown)}
+        >
+          <More size="small" />
+        </StyledMenuIconButton>
+        {openListboxDropdown && (
```
haxxmaxx, line 89: how come this is inside a new div next to the button?

a-m-dev: we have a new PR related to the popover version of listbox. 
this one is closed since the solution in the nebula got changed.

PR #802, opened 2023-01-17
fix: avoid rerendering when deleting alternative columns

linhnguyen-qlik:
When deleting alternative columns functions remove in sn-table/ext/data.js is called for both columns and alternative columns. This trigger a re-rendering. For removing alternative columns re-rendering is not necessary.

### DIFF
--- a/src/ext/data.js
+++ b/src/ext/data.js
@@ -64,7 +64,7 @@ export default function getData(env) {
         return dimension;
       },
       remove(dimension, data, hcHandler, idx) {
-        if (dimension && dimension.isAlternative) return;
+        if (dimension?.isAlternative) return;
         const { qColumnOrder, columnWidths } = hcHandler.hcProperties;
         indexRemoved(qColumnOrder, idx);
         columnWidths.splice(qColumnOrder[idx], 1);
--- a/src/ext/data.js
+++ b/src/ext/data.js
@@ -64,9 +64,11 @@ export default function getData(env) {
         return dimension;
       },
       remove(dimension, data, hcHandler, idx) {
-        const { qColumnOrder, columnWidths } = hcHandler.hcProperties;
-        indexRemoved(qColumnOrder, idx);
-        columnWidths.splice(qColumnOrder[idx], 1);
+        if (!dimension.isAlternative) {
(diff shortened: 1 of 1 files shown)


src/ext/data.js
```diff
@@ -64,9 +64,11 @@ export default function getData(env) {
         return dimension;
       },
       remove(dimension, data, hcHandler, idx) {
-        const { qColumnOrder, columnWidths } = hcHandler.hcProperties;
-        indexRemoved(qColumnOrder, idx);
-        columnWidths.splice(qColumnOrder[idx], 1);
+        if (!dimension.isAlternative) {
```
mhs30, line 67: Does this prop always exists? lets add a null check just in case

src/ext/data.js
```diff
@@ -64,9 +64,11 @@ export default function getData(env) {
...
+          const { qColumnOrder, columnWidths } = hcHandler.hcProperties;
+          indexRemoved(qColumnOrder, idx);
+          columnWidths.splice(qColumnOrder[idx], 1);
```
cbt1, line 70: Is the removed dimension not represented in `qColumnOrder` and `columnWidths`?

linhnguyen-qlik: No it is not if it is alternative dimension it is not showed in the table.

linhnguyen-qlik:
> "VIZ-2287" is unknown to external consumers. Remove it from the title and make sure it does end up as part of the commit message when merging.


Yeah, right sorry I forgot.

src/ext/data.js
```diff
@@ -64,6 +64,7 @@ export default function getData(env) {
         return dimension;
       },
       remove(dimension, data, hcHandler, idx) {
+        if (dimension && dimension.isAlternative) return;
```
cbt1, line 67:
```suggestion
        if (dimension?.isAlternative) return;
```

cbt1 (approved):
Not really sure where `isAlternative` is coming from but LGTM

linhnguyen-qlik:
> Not really sure where `isAlternative` is coming from but LGTM

@cbt1 it is coming from sense-client. I added you as reviewer.

mhs30 (approved):

PR #804, opened 2023-01-18
feat: scale row height based on font size

cbt1:
Scale the row height in virtual mode based on the font size


src/table/utils/styling-utils.ts
```diff
@@ -6,6 +6,13 @@ import { GeneratedStyling, CellStyle, FooterStyle } from '../types';
 import { SelectionStates, PAGINATION_HEIGHT } from '../constants';
 import { SELECTION_STYLING, COLORING } from '../styling-defaults';
 
+const LINE_HEIGHT = 4 / 3;
+const CELL_PADDING_HEIGHT = 8;
+const CELL_BORDER_HEIGHT = 1;
```
haxxmaxx, line 11: these would fit nicely in styling-defaults. Possibly that the file should have a different name

src/table/components/virtualized-table/utils/get-height.ts
```diff
@@ -0,0 +1,11 @@
+import { COMMON_CELL_STYLING } from '../../../styling-defaults';
+import { GeneratedStyling } from '../../../types';
+import { fontSizeToRowHeight } from '../../../utils/styling-utils';
+import { MIN_BODY_ROW_HEIGHT, MIN_HEADER_HEIGHT } from '../constants';
+import { BodyStyle } from '../types';
+
+export const getHeaderRowHeight = ({ fontSize = COMMON_CELL_STYLING.fontSize }: GeneratedStyling) =>
+  Math.max(MIN_HEADER_HEIGHT, Math.round(fontSizeToRowHeight(fontSize)));
```
haxxmaxx, line 8: why rounding?

src/table/components/virtualized-table/Table.tsx
```diff
@@ -33,26 +33,29 @@ const Table = (props: TableProps) => {
     }),
     [layout, theme.name()] // eslint-disable-line react-hooks/exhaustive-deps
   );
+  const headerRowHeight = getHeaderRowHeight(headerStyle);
+  const bodyRowHeight = getBodyRowHeight(bodyStyle);
+  const headerAndTotalsHeight = totals.atTop || totals.atBottom ? headerRowHeight + bodyRowHeight : headerRowHeight;
```
haxxmaxx, line 38: I only find the usage of `getHeaderRowHeight` and `getHeaderRowHeight` here. I would hookify them and return these three variables, then the logic is easier to test.

src/table/components/virtualized-table/hooks/use-totals.ts
```diff
@@ -1,24 +1,12 @@
...
-
-  return useMemo(
-    () => ({
-      ...totalsPosition,
-      shrinkBodyHeightBy: totalsPosition.atTop || totalsPosition.atBottom ? HEADER_AND_TOTALS_HEIGHT : HEADER_HEIGHT,
-    }),
-    [totalsPosition]
-  );
-};
+const useTotals = (layout: TableLayout) => useMemo(() => getTotalPosition(layout), [layout]);
```
haxxmaxx, line 10: do you still need this hook? It it would be a one-liner where the hook is used instead

haxxmaxx (commented):
looks good in general, minor questions before I can approve

src/table/components/virtualized-table/Table.tsx
cbt1: I see no value in creating a hook for it as no react hooks are being used. But what I can do is more them to a util function that returns all 3 values? It would possibly reduce the number of lines of code in this file

src/table/components/virtualized-table/hooks/use-totals.ts
cbt1: Agree. Will remove hook 👍

src/table/components/virtualized-table/utils/get-height.ts
cbt1: To fix an issue where the cell borders would sometimes slight overlap, making some borders appear bigger than others. My guess is that it's caused by a floating point precision issue as I went away when I added the `Math.round`

src/table/utils/styling-utils.ts
cbt1: I would think they fitter better in a "constants" file, but since they are only used here I don't think it will add much value to move them

src/table/components/virtualized-table/Table.tsx
haxxmaxx: that sounds good

src/table/components/virtualized-table/utils/get-height.ts
haxxmaxx: I see. I was going to say you should test with really small font sizes, I think we support down to 5px. But now I see that you set a min height. Not sure if we want to add that limit, but also anything smaller than 24 this is tiny

src/table/utils/styling-utils.ts
haxxmaxx: sure, they can be here since until they are needed elsewhere

src/table/components/virtualized-table/utils/get-height.ts
haxxmaxx: let's keep it this way until we know there is a user need for smaller rows

haxxmaxx (approved):

a-m-dev (commented):
I have a suggestion, 

My gut feeling says me that we should not rely on font size of a cell in order to get height of it, because they are two different things and they could be different in several scenarios (theming is one example)

#A dumb idea but,
could we maybe think to have 2 refs (one in header cell, one in normal row) and getting actual heights of them => put it in context for overall use. Then we can add all of the other stuff like border and other spacing things on top of those actual real world values for more precision. 

In that case `fontSizeToRowHeight` function and all the calculation will not be needed because we already have the inner height of our head and normal cells from refs. 

maybe I'm wrong but any ideas?

haxxmaxx:
> I have a suggestion,
> 
> My gut feeling says me that we should not rely on font size of a cell in order to get height of it, because they are two different things and they could be different in several scenarios (theming is one example)
> ## A dumb idea but,
> 
> could we maybe think to have 2 refs (one in header cell, one in normal row) and getting actual heights of them => put it in context for overall use. Then we can add all of the other stuff like border and other spacing things on top of those actual real world values for more precision.
> 
> In that case `fontSizeToRowHeight` function and all the calculation will not be needed because we already have the inner height of our head and normal cells from refs.
> 
> maybe I'm wrong but any ideas?

First off, the font size that we are resizing against is already calculated. What getHeader/BodyStyle returns have taken theme and settings into account. So that is the final font size.

I think react window needs these heights before hand, or @cbt1 ?

cbt1:
> > I have a suggestion,
> > My gut feeling says me that we should not rely on font size of a cell in order to get height of it, because they are two different things and they could be different in several scenarios (theming is one example)
> > ## A dumb idea but,
> > could we maybe think to have 2 refs (one in header cell, one in normal row) and getting actual heights of them => put it in context for overall use. Then we can add all of the other stuff like border and other spacing things on top of those actual real world values for more precision.
> > In that case `fontSizeToRowHeight` function and all the calculation will not be needed because we already have the inner height of our head and normal cells from refs.
> > maybe I'm wrong but any ideas?
> 
> First off, the font size that we are using for height, is already calculated. What getHeader/BodyStyle returns have taken theme and settings into account. So that is the final font size.
> 
> I think react window needs these heights before hand, or @cbt1 ?

Yes we need to figure out the height of a cell BEFORE we render it, otherwise the virtual scroll would not work. So now we just use font-size. But it would be nice to support a case with multiple lines of text, in which case we would need to factor that in as well. But that is much harder to figure before we render and have all the required information. @a-m-dev

PR #805, opened 2023-01-18
feat: check if selections is enabled

cbt1:
Adds a check for constraints in vt mode before enabling selections

### DIFF
--- a/src/table/components/virtualized-table/__tests__/Table.spec.tsx
+++ b/src/table/components/virtualized-table/__tests__/Table.spec.tsx
@@ -18,13 +18,14 @@ describe('<Table />', () => {
   let selectionsAPI: ExtendedSelectionAPI;
   let isModal = false;
   let user: UserEvent;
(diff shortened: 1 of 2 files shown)


src/table/components/virtualized-table/__tests__/Table.spec.tsx
```diff
@@ -177,6 +182,54 @@ describe('<Table />', () => {
       await waitFor(() => expect(screen.getByText('dimension-c1-r0').parentNode).toHaveClass('excluded'));
       await waitFor(() => expect(screen.getByText('measure-c2-r0').parentNode).toHaveClass('excluded'));
     });
+
+    describe('constraints', () => {
+      it('should not be able to select a cell with "active" and "select" constraints', async () => {
```
haxxmaxx, line 187: I would say this is implicitly tested with the other two

haxxmaxx (approved):

PR #806, opened 2023-01-18
feat: color column by expression

cbt1:


src/types.ts
```diff
@@ -88,7 +88,7 @@ export interface TableLayout extends Omit<EngineAPI.IGenericHyperCubeLayout, 'qH
 
 export interface Cell {
   qText?: string;
-  qAttrExps: EngineAPI.INxAttributeExpressionValues;
+  qAttrExps?: EngineAPI.INxAttributeExpressionValues;
```
cbt1, line 91: I made this change as it reflects reality and is inline with the API spec [here](https://qlik.dev/apis/json-rpc/qix/schemas#%23%2Fdefinitions%2Fschemas%2Fentries%2FNxCell).

But it did end up not being necessery for the feature itself.

haxxmaxx: that's fine

src/table/components/virtualized-table/utils/__tests__/get-cell-style.test.ts
```diff
@@ -71,6 +99,49 @@ describe('getCellStyle', () => {
           'repeating-linear-gradient(-45deg, rgba(200,200,200,0.08), rgba(200,200,200,0.08) 2px, rgba(200,200,200,0.3) 2.5px, rgba(200,200,200,0.08) 3px, rgba(200,200,200,0.08) 5px), bodyBackground',
       });
     });
+
+    test('with column styling and selection state is INACTIVE', () => {
```
haxxmaxx, line 103: So technically there should be 16 cases here: 4 selection states * 2 hover cases * 2 column styling cases. Now there are 14. I haven't looked at all in detail. But I'd say either we have all 16 or we remove some, since some are them are basically the same thing. First off, inactive and possible are treated in the same way I think. And selection  will always override everything (including hover). If you cut the inactive ones at least, that is down to 12

haxxmaxx (commented):
implementation looks good, just the question about redundant tests

src/table/components/virtualized-table/utils/__tests__/get-cell-style.test.ts
cbt1: Technically there are way more than 16 cases. It’s is pure function and in an ideal world all different inputs should be tested. But in reality the number of different inputs are just way too many to consider testing them all. Does the function have too much responsibility? Yes it does. Should it be refactored? Yes, I just have not found a way to reduce the complexity of it. But when I do, I will.

So I went with some kind of middle ground where I added tests for plausible inputs rather than ALL possible once.

Looks like I could not really decide on a middle ground here between the two times I’ve added tests for this function. As I started with `4 selection states * 2 hover states` but somehow now ended up with: `(4 selection states * 2 hover states) + (3 selection states * 2 column states) = 14`.

You seem to prefer `3 selection state * 2 hover states * 2 column states` so will go with that.

haxxmaxx: sounds good

haxxmaxx (approved):

PR #807, opened 2023-01-19
refactor: use baseProps on context for pagination implementation

haxxmaxx:
- use `baseProps` for pagination table, instead of props drilling
- add `rootElement` since it is very commonly used
- add `embed` since it is passed all the way to `listBoxWrapper` and that is the only place it is used

I think we can add more stuff. Anything that triggers a rerender all the way from index, would not create extra renders if they are put on the context

### DIFF
--- a/src/__test__/test-with-providers.tsx
+++ b/src/__test__/test-with-providers.tsx
@@ -20,6 +20,8 @@ interface ProviderProps {
   theme?: ExtendedTheme;
   keyboard?: stardust.Keyboard;
   direction?: 'ltr' | 'rtl';
+  rootElement?: HTMLElement;
+  embed?: stardust.Embed;
 }
 
 const TestWithProviders = ({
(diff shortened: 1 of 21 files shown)


src/__test__/test-with-providers.tsx
```diff
@@ -42,6 +44,8 @@ const TestWithProviders = ({
   cellCoordMock = undefined,
   selectionDispatchMock = undefined, // Can be used to avoid selectionDispatch infinite loop
   direction = 'ltr',
+  rootElement = {} as HTMLElement,
+  embed = {} as stardust.Embed,
```
haxxmaxx, line 48: casting to make sure they are not optional. We know they are, but acccording to the type it can be undefined, since we are sharing the type with the native solution. Will change this when we know the future of the native table in this repo

cbt1 (approved):
🚢

PR #809, opened 2023-01-20
fix: remove sort options for head menu

haxxmaxx:
Now only showing the search option, thus there is no menu for measures currently. There will be as soon as we add options that exist for both measures and dimensions

### DIFF
--- a/src/table/components/head/HeadCellMenu.tsx
+++ b/src/table/components/head/HeadCellMenu.tsx
@@ -2,8 +2,6 @@ import React, { useRef, useState, useMemo } from 'react';
 import Menu from '@mui/material/Menu';
 import More from '@qlik-trial/sprout/icons/More';
 import Search from '@qlik-trial/sprout/icons/Search';
-import Descending from '@qlik-trial/sprout/icons/Descending';
-import Ascending from '@qlik-trial/sprout/icons/Ascending';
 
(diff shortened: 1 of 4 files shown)


src/table/components/head/HeadCellMenu.tsx
```diff
@@ -66,10 +35,10 @@ export default function HeadCellMenu({
           ]
         : []),
     ],
-    [translator, isInteractionEnabled, isCurrentColumnActive, sortDirection, isDimension]
+    [translator, isDimension]
   );
 
-  return (
+  return menuItemGroups.length ? (
```
cbt1, line 41: Nit-picking but it might be easier to read the code of you exit early and return null when it is a measure instead. But that is mostly me not being a fan of long ternary operator statements

cbt1 (approved):

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: hmm. yeah I think I'd prefer that as well. I will keep the spreading in the menuItemGroups array though, since we will need that as soon as we add more options.

haxxmaxx: actually you can't early return, because then you get hooks called conditionally. This will be cleaned up soon anyway

PR #810, opened 2023-01-20
fix: reduce flickering with stale data

cbt1:
This PR reduces the occurrence of flickering when data becomes stale, ex. new layout triggered by selections in some other object, property changes or changing page.

**`Body` component re-render flow:**

BEFORE:
1. Reset data state to an empty array
2. Render Body with the empty array as data (this would cause the flickering)
3. Call engine to fetch new data
4. Set state with new data
5. Render Body with new data

AFTER:
1. Render Body with previous data (use the deferred row count to guard against row count changes that does not match the length of the previous data)
2. Call engine to fetch new data
3. Set state with new data
4. Render Body with new data

There as still an issue with the scroll position being reset on step 1, which under certain circumstances can be seen as a kind of flickering.

### DIFF
--- a/src/table/components/virtualized-table/hooks/use-data/index.ts
+++ b/src/table/components/virtualized-table/hooks/use-data/index.ts
@@ -1,6 +1,5 @@
 import { useCallback, useEffect, useState } from 'react';
 import { PageInfo, Row, Column, TableLayout } from '../../../../../types';
-import useOnPropsChange from '../use-on-props-change';
 import { COLUMN_DATA_BUFFER_SIZE, ROW_DATA_BUFFER_SIZE } from '../../constants';
 import useGetHyperCubeDataQueue from '../use-get-hypercube-data-queue';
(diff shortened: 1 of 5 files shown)


src/table/components/virtualized-table/hooks/use-data/index.ts
```diff
@@ -1,6 +1,5 @@
 import { useCallback, useEffect, useState } from 'react';
 import { PageInfo, Row, Column, TableLayout } from '../../../../../types';
-import useOnPropsChange from '../use-on-props-change';
```
haxxmaxx, line 3: is this still used elsewhere?

haxxmaxx (approved):

src/table/components/virtualized-table/hooks/use-data/index.ts
cbt1: Yes in 2 places

PR #811, opened 2023-01-23
feat: use pagination property

cbt1:
Enables the user to toggle pagnination mode ON/OFF. Default is ON and toggle is hidden behind a feature flag.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -85,7 +85,7 @@ export default function supernova(env: Galaxy) {
       const embed = useEmbed();
 
       const [pageInfo, setPageInfo] = useState(initialPageInfo);
-      const shouldRenderVirtualizedTable = false; // layout.scrollMode === 1;
+      const shouldRenderVirtualizedTable = areBasicFeaturesEnabled && layout.presentation?.usePagination === false;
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -85,7 +85,7 @@ export default function supernova(env: Galaxy) {
       const embed = useEmbed();
 
       const [pageInfo, setPageInfo] = useState(initialPageInfo);
-      const shouldRenderVirtualizedTable = false; // layout.scrollMode === 1;
+      const shouldRenderVirtualizedTable = areBasicFeaturesEnabled && layout.presentation?.usePagination === false;
```
haxxmaxx, line 88: use a `!` here? or are we missing a type on usePagination?

cbt1: Both `presentation` and `presentation.usePagination` are optional. But the default value is to use pagination. So need to check that `usePagination` is explicitly set otherwise it would break the default value.

Once we do change default value to we can do `!`.

haxxmaxx: ah ok makes sense

haxxmaxx (approved):

PR #813, opened 2023-01-24
build: use node lts version

cbt1:


cbt1:
We cannot bump the required node version to 18 yet as the docker image `mcr.microsoft.com/playwright` is still running v16 due to issues described [here](https://github.com/microsoft/playwright/commit/70065ba6ddcd37702a12c64fff971cae2cbb8848)

haxxmaxx:
> We cannot bump the required node version to 18 yet as the docker image `mcr.microsoft.com/playwright` is still running v16 due to issues described [here](https://github.com/microsoft/playwright/commit/70065ba6ddcd37702a12c64fff971cae2cbb8848)

Does that mean we can't get new versions of nebula in?

cbt1:
> > We cannot bump the required node version to 18 yet as the docker image `mcr.microsoft.com/playwright` is still running v16 due to issues described [here](https://github.com/microsoft/playwright/commit/70065ba6ddcd37702a12c64fff971cae2cbb8848)
> 
> Does that mean we can't get new versions of nebula in?

No, but circleci needs to install and build using v18 and locally you should use v18 as well

haxxmaxx (approved):

PR #814, opened 2023-01-24
Crl/buttons in header

haxxmaxx:
Introduces the `HeadCellContent`. It should be reusable for both tables. Right now only used in the pagination table.

**Things that break with this PR**
- head font size
  - I think we might need some design input on this
- keyboard navigation

I'm leaving some things commented that I might need to rewrite, so I don't have to redo it from scratch. When this header work is completed, I think it is a good time to do some "e2e" RTL tests for sorting


src/table/hooks/use-head-icons.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { useMemo } from 'react';
+import Lock from '@qlik-trial/sprout/icons/react/Lock';
+import Descending from '@qlik-trial/sprout/icons/react/Descending';
+import Ascending from '@qlik-trial/sprout/icons/react/Ascending';
+import { Column } from '../../types';
+
+const useHeadIcons = (column: Column) => {
```
haxxmaxx, line 7: unsure if this should be a hook or not. no need to memo really, but it uses components. It is mostly to get the logic out of the component

src/table/components/head/TableHeadWrapper.tsx
```diff
@@ -52,58 +36,37 @@ function TableHeadWrapper({ tableData, changeSortOrder, areBasicFeaturesEnabled
...
 
           return (
             <StyledHeadCell
               headerStyle={headerStyle}
               key={column.id}
               align={column.align}
               className="sn-table-head-cell sn-table-cell"
-              tabIndex={tabIndex}
               aria-sort={ariaSort}
-              aria-pressed={isCurrentColumnActive}
```
haxxmaxx, line 80: I'm pretty sure we don't need this. It is mostly used for checkboxes and stuiff. aria-sort is undefined if it is not sorted on that column, should be enough

src/table/hooks/use-head-icons.tsx
cbt1: There is some potential to wrap it in `useMemo` as the return value is a non-primitive value. Which would always have triggered a re-render of the `StyledSortButton` component. But if that component is not wrapped in `React.memo` that will occur anyway.

By the looks of it `StyledSortButton` is not using `React.memo` so the `useMemo` call should not be needed. BUT might be that the `styled` function is memoizing the component but I don't really know.

haxxmaxx: ok yeah no real need to memo anything here. Anytime `HeadCellContent ` rerenders it should update the icons. So this file should just should just be named `get-head-icons.tsx`?

cbt1: Yeah I think you can. It could be useful for the virtual scroll table but might as well wait and see with that

haxxmaxx: fixed

src/table/components/head/HeadCellContent.tsx
```diff
@@ -0,0 +1,58 @@
...
+  isActive,
+  areBasicFeaturesEnabled,
+}: HeadCellContentProps) {
+  const { constraints, selectionsAPI, keyboard, translator } = useContextSelector(
+    TableContext,
+    (value) => value.baseProps
+  );
+  const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
+
+  const { startIcon, endIcon, lockIcon } = useHeadIcons(column);
```
cbt1, line 23:
```suggestion
  const { startIcon, endIcon, lockIcon } = getHeadIcons(column);
```

src/table/components/head/HeadCellContent.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+
+import { useContextSelector, TableContext } from '../../context';
+import { HeadCellContentProps } from '../../types';
+import { FullSortDirection } from '../../constants';
+import useHeadIcons from '../../utils/get-head-icons';
```
cbt1, line 6:
```suggestion
import getHeadIcons from '../../utils/get-head-icons';
```

cbt1:
Also dont forget to merge with correct commit message

PR #817, opened 2023-01-26
feat: sorting and head menu in virtual scroll

haxxmaxx:
- using `HeadCellContent` inside `HeaderCell`
- moving `changeSortOrder` to context

### DIFF
--- a/src/table/virtualized-table/HeaderCell.tsx
+++ b/src/table/virtualized-table/HeaderCell.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
 import { Column } from '../../types';
+import { useContextSelector, TableContext } from '../context';
 import { GeneratedStyling } from '../types';
-import CellText from '../components/CellText';
+import HeadCellContent from '../components/head/HeadCellContent';
 
 interface HeaderCellProps {
   index: number;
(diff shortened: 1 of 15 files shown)


src/table/virtualized-table/HeaderCell.tsx
```diff
@@ -17,8 +18,10 @@ const HeaderCell = ({ index, style, data }: HeaderCellProps) => {
     columns,
     headerStyle: { ...applicableStyle },
   } = data;
+  const { layout } = useContextSelector(TableContext, (value) => value.baseProps);
   const datum = columns[index];
```
cbt1, line 22: Might as well rename `datum` to `column`

cbt1 (approved):
Cannot believe how easy that looked. Great work!

src/table/virtualized-table/HeaderCell.tsx
haxxmaxx: agree

src/table/virtualized-table/HeaderCell.tsx
```diff
@@ -37,7 +40,7 @@ const HeaderCell = ({ index, style, data }: HeaderCellProps) => {
         fontWeight: 'bold',
       }}
     >
-      <CellText singleLine>{datum.label}</CellText>
+      <HeadCellContent column={datum} columnIndex={index} isActive={isActive} areBasicFeaturesEnabled />
```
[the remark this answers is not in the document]
haxxmaxx: yeah that would be nice, it is kind of used all over the place.

haxxmaxx:
> Cannot believe how easy that looked. Great work!

Thanks! That was kind of all the prep work for it. Hopefully we can use the same way of implementing other data representations down the line, if we need them in both tables

PR #820, opened 2023-01-31
feat (Listbox): listbox popover for normal dimensions

a-m-dev:
PR for embedding listbox only for normal dimensions. 
We have check for master dimensions based on `qLibraryId`


src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,23 +1,35 @@
...
 
 export default function HeadCellMenu({ columnIndex, isDimension, tabIndex }: HeadCellMenuProps) {
-  const { translator } = useContextSelector(TableContext, (value) => value.baseProps);
+  const { translator, embed, layout } = useContextSelector(TableContext, (value) => value.baseProps);
   const [openMenuDropdown, setOpenMenuDropdown] = useState(false);
-  const [openListboxDropdown, setOpenListboxDropdown] = useState(false);
-  const anchorRef = useRef<HTMLButtonElement>(null);
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const listboxRef = useRef<HTMLDivElement>(null);
+  const isMasterDimension = layout.qHyperCube.qDimensionInfo[columnIndex]?.qLibraryId;
```
haxxmaxx, line 16: @cbt1 did we conclude that this is always the case for masterDimensions?

haxxmaxx: also, I think this wont work for dimensions that are placed after measures. the `columnIndex` passed in here does not correspond to where the index in qDimensionInfo

haxxmaxx: I think we should pass column instead of `columnIndex` to `HeadCellMenu`, especially since that `columnIndex` exists on column 🙃

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,23 +1,35 @@
...
+    embed.__DO_NOT_USE__.popover(listboxRef.current, fieldId, {
+      anchorOrigin: { vertical: 'bottom', horizontal: 'left' },
+      transformOrigin: { vertical: 'top', horizontal: 'left' },
+    });
+  };
 
   const menuItemGroups = useMemo<MenuItemGroup[]>(
     () => [
-      ...(isDimension
+      ...(isDimension && !isMasterDimension
```
haxxmaxx, line 32: we should make this into one bool, reducing the deps for this useMeno

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: then we can also remove isDimension, since that is also on the column... I forgot all this when refactoring the header

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -35,24 +47,28 @@ export default function HeadCellMenu({ columnIndex, isDimension, tabIndex }: Hea
...
+          id="sn-table-head-menu-button"
+          aria-controls={openMenuDropdown ? 'sn-table-head-menu' : undefined}
+          aria-expanded={openMenuDropdown ? 'true' : undefined}
+          aria-haspopup="true"
+          onClick={() => setOpenMenuDropdown(!openMenuDropdown)}
+        >
+          <More size="small" />
+        </StyledMenuIconButton>
+
+        <div ref={listboxRef} />
```
haxxmaxx, line 70: why does this need to be wrapped in a div?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,23 +1,35 @@
...
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const listboxRef = useRef<HTMLDivElement>(null);
+  const isMasterDimension = layout.qHyperCube.qDimensionInfo[columnIndex]?.qLibraryId;
+
+  const embedListbox = () => {
+    if (!layout || !embed) return;
+    const fieldId = layout.qHyperCube.qDimensionInfo[columnIndex]?.qFallbackTitle;
+    if (!fieldId) return;
+    // @ts-ignore `__DO_NOT_USE__` api does not require type check
+    // eslint-disable-next-line
```
haxxmaxx, line 23: I think this is a nice pattern to do this https://github.com/qlik-oss/sn-table/blob/main/src/table/utils/selections-utils.ts#L41

src/table/components/head/HeadCellMenu.tsx
a-m-dev: because the parent element is flex and we only want to show 2 things in a header cell (title button +  '...' menu ). if we put it without a wrapper div, we will end up showing 3 flexed items stretching over head cell.

src/table/components/head/HeadCellMenu.tsx
a-m-dev: You mean adding a TODO before @ts-ignore description? 
we also need lint ignore because of using __ before api name that we don't have control over it

haxxmaxx: yeah

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: Ah I see the popover itself is not added here in the tree, but it is mounted on this div, and that makes the alignment in the flexbox strange?

src/table/components/head/HeadCellMenu.tsx
cbt1: @haxxmaxx Yes I would think so.

Also I would think that storing the `isMasterDimension` on the type `Column` would make more sense

haxxmaxx: that is a good idea!

src/table/components/head/HeadCellMenu.tsx
a-m-dev: not actually, this small div is acting like an anchor point for popover. That is the only use case for it. but because of existence of this div directly in HeadCellMenu while it has a display:flex -> it makes it's direct children to spread within it. so wrapping it inside of a div avoids that problem.

src/table/components/head/HeadCellMenu.tsx
a-m-dev: ✅ updated

src/handle-data.ts
```diff
@@ -65,11 +65,13 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
   const { qError, qFallbackTitle, textAlign, qAttrExprInfo, qSortIndicator, qReverseSort, qApprMaxGlyphCount } = info;
   const isHidden = qError?.qErrorCode === 7005;
   const isLocked = isDim && (info as ExtendedNxDimensionInfo).qLocked;
+  const isMasterDim = isDim && !!(info as ExtendedNxDimensionInfo).qLibraryId;
```
haxxmaxx, line 68: I think it makes more sense to check this for both measure and dimension
```suggestion
  const isMasterItem = !!info.qLibraryId;
```

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,23 +1,39 @@
 import React, { useRef, useState, useMemo } from 'react';
 import Menu from '@mui/material/Menu';
-import More from '@qlik-trial/sprout/icons/react/More';
-import Search from '@qlik-trial/sprout/icons/react/Search';
+import More from '@qlik-trial/sprout/icons/More';
+import Search from '@qlik-trial/sprout/icons/Search';
```
haxxmaxx, line 4: why this change?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,23 +1,39 @@
...
-export default function HeadCellMenu({ columnIndex, isDimension, tabIndex }: HeadCellMenuProps) {
-  const { translator } = useContextSelector(TableContext, (value) => value.baseProps);
+export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
+  const { translator, embed } = useContextSelector(TableContext, (value) => value.baseProps);
   const [openMenuDropdown, setOpenMenuDropdown] = useState(false);
   const [openListboxDropdown, setOpenListboxDropdown] = useState(false);
-  const anchorRef = useRef<HTMLButtonElement>(null);
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const listboxRef = useRef<HTMLDivElement>(null);
+  const showListboxMenu = column.isDim && !column.isMasterDim;
```
haxxmaxx, line 17: this isn't really what it is. This is if we should show the search menu item (at this point). So should be something like `showSearchMenuItem`

src/table/components/head/__tests__/HeadCellMenu.spec.tsx
```diff
@@ -1,21 +1,29 @@
+/* eslint-disable no-underscore-dangle */
 import React from 'react';
 import { stardust } from '@nebula.js/stardust';
 import { render, fireEvent, screen, waitFor, waitForElementToBeRemoved } from '@testing-library/react';
-import { TableLayout } from '../../../../types';
+import { Column, TableLayout } from '../../../../types';
 import HeadCellMenu from '../HeadCellMenu';
 import TestWithProviders from '../../../../__test__/test-with-providers';
 
+type EmbedWithDoNotUse = stardust.Embed & {
```
haxxmaxx, line 9: DIdn't think of this before. I think we should do this extending as for `ExtendedTranslator` etc. In the index file, we cast to this new type `ExtendedEmbed` and then we use that type throughout  the code

src/handle-data.ts
a-m-dev: Does master measure has qLibId as well? i get type complains about it if i do so

src/table/components/head/__tests__/HeadCellMenu.spec.tsx
a-m-dev: I would suggest let's do it in another PR and not pollute this one with that change, seems like there are lots of them.

cbt1 (approved):
Have not played around with it but LGTM

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -5,19 +5,35 @@ import Search from '@qlik-trial/sprout/icons/react/Search';
...
+  const embedListbox = () => {
+    // @ts-ignore TODO: no types for `__DO_NOT_USE__`, it will improve when it becomes stable
+    // eslint-disable-next-line
+    embed.__DO_NOT_USE__.popover(listboxRef.current, column.fieldId, {
+      anchorOrigin: { vertical: 'bottom', horizontal: 'left' },
+      transformOrigin: { vertical: 'top', horizontal: 'left' },
+    });
+
+    // @ts-ignore TODO: no types for `__DO_NOT_USE__`, it will improve when it becomes stable
+    // eslint-disable-next-line
```
haxxmaxx, line 28: is this needed? Looks like it was copy pasted from above, at least the comment isn't really correct

haxxmaxx (approved):

src/table/components/head/HeadCellMenu.tsx
a-m-dev: will update it asap

PR #823, opened 2023-02-06
refactor: put table styling on context

haxxmaxx:
since font sizes and famalies will be needed for doing a proper "hug" calculation, we need to put the styling on the context

### DIFF
--- a/src/table/context/TableContext.tsx
+++ b/src/table/context/TableContext.tsx
@@ -4,6 +4,7 @@ import { createSelectorProvider } from './createSelectorProvider';
 import { ContextValue, ContextProviderProps } from '../types';
 import useSelectionReducer from '../hooks/use-selection-reducer';
 import useColumnWidths from '../hooks/use-column-widths';
+import useTableStyling from '../hooks/use-table-styling';
 import { TableData } from '../../types';
 
 // In order to not have typing issues when using properties on the context,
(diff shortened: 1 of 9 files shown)


src/table/context/TableContext.tsx
```diff
@@ -52,6 +54,7 @@ export const TableContextProvider = ({
       embed,
       changeSortOrder,
       applyColumnWidths,
+      styling,
```
cbt1, line 57: Not sure this should be a base property. We don't really have a definition for a base property but I would consider it to be something that is part of the nebula environment/context and NOT a value that we resolve/create ourself.

So in this example the styling is derived from some of that "nebula environment/context".

Maybe something structure like this would make more sense as to what the different values are:
```js
const value = {
    state: {
      headRowHeight,
      setHeadRowHeight,
      focusedCellCoord,
      setFocusedCellCoord,
      selectionState,
      selectionDispatch: selectionDispatchMock || selectionDispatch,
      hoverIndex,
      setHoverIndex,
    },
    props: {
      changeSortOrder,
      styling,
    },
   env: {
      selectionsAPI,
      layout,
      model,
      translator,
      constraints,
      theme,
      keyboard,
      rootElement,
      embed,
    },
  }
```

Anyway overkill for this PR but something to think about.

cbt1 (approved):
LGTM

src/table/context/TableContext.tsx
haxxmaxx: I see your point. So the props outside of baseProps are carefully put seperately, so you fetch them, they are the same unless its value had changed, so for example:
`const focusedCellCoord = useContextSelector(TableContext, (value) => value.focusedCellCoord);`
That being said if we would put that inside a something called state or similar, I think this should be the same thing:
`const focusedCellCoord = useContextSelector(TableContext, (value) => value.state.focusedCellCoord);`

haxxmaxx: I agree we should seperate them kind of like you suggested, but for now I think we can consider `baseProps` as everything that is kind of updated on layout/model changes or comes straight from index

haxxmaxx:
@a-m-dev going to skip putting styling on the theme. If anything I think we should put the background stuff on styling instead, since theme is straight from nebula, and we are adding stuff to that. I think I could be good to have a look at together with a more structured context like bengtsson suggested

PR #824, opened 2023-02-07
fix: calculate hug based on real font style

haxxmaxx:
using the head and body styling to calculate/estimate the width of header, totals and body. 

Currently ignoring the extra padding differences between header and body. My next step is to fix the header label button for text wrap and font size. Then it makes sense to get an optimized result for hug.

### DIFF
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -47,7 +47,7 @@ export function getTotalPosition(layout: TableLayout) {
  * Gets the totals text for a column
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
-  if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText || '';
+  if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText ?? '';
   if (colIdx === 0 && pageColIdx === 0) return layout.totals.label;
   return '';
 }
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -47,7 +47,7 @@ export function getTotalPosition(layout: TableLayout) {
  * Gets the totals text for a column
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
-  if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText;
+  if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText || '';
   if (colIdx === 0 && pageColIdx === 0) return layout.totals.label;
   return '';
 }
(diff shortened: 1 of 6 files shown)


src/handle-data.ts
```diff
@@ -47,7 +47,7 @@ export function getTotalPosition(layout: TableLayout) {
  * Gets the totals text for a column
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
-  if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText;
+  if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText || '';
```
cbt1, line 50: Consider using nullish coalescing operator `??` instead of `||` to deal with `null` and `undefined`. There can be some unexpected interactions with `||` when you have falsy values.

Not really expected anything to change here but in general I'd prefer ?? over || when you want to deal with null and undefined.

src/table/hooks/use-column-widths.ts
```diff
@@ -79,14 +79,18 @@ export const getColumnWidths = (
...
+  tableWidth: number,
+  { head, body }: TableStyling
 ): [number[], React.Dispatch<React.SetStateAction<number[]>>] => {
-  const measureTextMethods = useMeasureText('13px', 'Arial');
-  // TODO use actual font size (and font eventually)
-  const [columnWidths, setColumnWidths] = useState(getColumnWidths(columns, tableWidth, measureTextMethods));
+  const measureHeadLabel = useMeasureText(head.fontSize, head.fontFamily).measureText;
+  const { measureText, estimateWidth } = useMeasureText(body.fontSize, body.fontFamily);
+  const getHugWidth = (headLabel: string, totalsLabel: string, glyphCount: number) =>
+    Math.max(measureHeadLabel(headLabel), measureText(totalsLabel), estimateWidth(glyphCount));
```
cbt1, line 88: "padding", "border" and that menu button will also occupy space so that should be accounted for when measuring the require width. I do think that is out-of-scope for this PR but just an FYI.

src/table/hooks/use-column-widths.ts
```diff
@@ -79,14 +79,18 @@ export const getColumnWidths = (
...
+
+  const [columnWidths, setColumnWidths] = useState(getColumnWidths(columns, tableWidth, getHugWidth));
```
cbt1, line 90: Try to avoid calling `getColumnWidths` every time this hook runs. As it can become costly to run that function everytime the state is updated just to discard the "initial value".

Image a table with 2000 column. Running `getColumnWidths` every state update.

cbt1:
```suggestion
  const [columnWidths, setColumnWidths] = useState(() => getColumnWidths(columns, tableWidth, getHugWidth));
```

src/table/hooks/use-column-widths.ts
haxxmaxx: yes I mention it in the description

src/handle-data.ts
haxxmaxx: you are right, this is a numerical value so it could be 0. good catch

src/table/hooks/use-column-widths.ts
haxxmaxx: Oh I didn't know that, good to know

a-m-dev (approved):

cbt1 (approved):

PR #825, opened 2023-02-08
fix: scrollbar width vt mode

cbt1:
As the width and height is computed by the table, when the browsers add a scrollbar, some of the space included in the width/height is occupyed by the scrollbar. This can cause the effect that a vertical scrollbar can trigger a horizontal scrollbar to appear, and vice-versa.

I had hoped I would be able to find a better solution but none so far. So in this PR there is an attempt to discover if there is a scrollbar, if so, measure its width and reduce that from the width/height of the table and re-render it.

### DIFF
--- a/src/table/virtualized-table/hooks/use-data/index.ts
+++ b/src/table/virtualized-table/hooks/use-data/index.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { PageInfo, Row, Column, TableLayout } from '../../../../types';
 import { COLUMN_DATA_BUFFER_SIZE, ROW_DATA_BUFFER_SIZE } from '../../constants';
 import { SetCellSize } from '../../types';
@@ -25,6 +25,7 @@ const useData = (
   setCellSize: SetCellSize
 ): UseData => {
   const [rowsInPage, setRowsInPage] = useState<Row[]>(Array(rowCount).fill(undefined));
+  const mutableRowsInPage = useRef(rowsInPage);
 
   const getDataPages = useCallback(
     async (pages: EngineAPI.INxPage[]) => model.getHyperCubeData('/qHyperCubeDef', pages),
@@ -35,6 +36,7 @@ const useData = (
     (dataPages: EngineAPI.INxDataPage[]) =>
       setRowsInPage((prevRows) => {
         const nextRows = [...prevRows];
+        mutableRowsInPage.current = nextRows;
 
         dataPages.forEach((dataPage) => {
           dataPage.qMatrix.forEach((matrixRow, matrixRowIdx) => {
@@ -56,7 +58,7 @@ const useData = (
 
         return nextRows;
       }),
-    [pageInfo, columns, layout, setCellSize]
+    [pageInfo, columns, layout, setCellSize, mutableRowsInPage]
   );
 
   // The queue takes a EngineAPI.INxPage object as items and adds them to a queue and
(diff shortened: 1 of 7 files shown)


src/table/virtualized-table/hooks/use-data/index.ts
```diff
@@ -109,11 +111,12 @@ const useData = (
...
 
-    getDataPages([{ qLeft: 0, qTop, qHeight, qWidth }]).then((dataPages) => {
+    const onBeforeHandlePages = () => {
       setRowsInPage(Array(rowCount).fill(undefined)); // Reset rows to initial value
-      handleDataPages(dataPages);
-    });
-  }, [getDataPages, handleDataPages, layout, visibleRowCount, visibleColumnCount, pageInfo, queue, rowCount]);
+    };
+
+    queue.enqueue({ qLeft: 0, qTop, qHeight, qWidth }, onBeforeHandlePages);
```
cbt1, line 118: The change in this PR can result in 2-3 renders in the `Table` component that would all request new data with possibly slightly different call parameters as qWidth and/or qHeight might change. Potentially triggering a race condition.

So I changed to use the queue her for fetching data as the `clear()` call in this useEffect will cancel any ongoing requests.

src/table/virtualized-table/hooks/use-scrollbar-width.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (ref.current) {
+      setYScrollbarWidth(ref.current.offsetWidth - ref.current.clientWidth);
+      setXScrollbarWidth(ref.current.offsetHeight - ref.current.clientHeight);
+    }
+  }, [ref]);
+
+  useOnPropsChange(() => {
+    setYScrollbarWidth(widthDiff);
+    setXScrollbarWidth(heightDiff);
+  }, [widthDiff, heightDiff]);
```
haxxmaxx, line 21: I missed when this hook was introduced, had a look at  the definition. How is it different from a useEffect?

src/table/virtualized-table/hooks/use-scrollbar-width.ts
```diff
@@ -0,0 +1,26 @@
...
+  const heightDiff = offsetHeight - clientHeight;
+
+  useEffect(() => {
+    if (ref.current) {
+      setYScrollbarWidth(ref.current.offsetWidth - ref.current.clientWidth);
+      setXScrollbarWidth(ref.current.offsetHeight - ref.current.clientHeight);
+    }
+  }, [ref]);
+
+  useOnPropsChange(() => {
```
haxxmaxx, line 18: so the useEffect runs when the ref changes, I guess when the object size becomes bigger for some reason? And useOnPropsChange runs if the scrollbar appears?

haxxmaxx (approved):

src/table/virtualized-table/hooks/use-scrollbar-width.ts
cbt1: The `useEffect` is there to trigger the first check for a scrollbar. It's in a useEffect with `ref` as a dep because we want to trigger it first when the `Table` (where the ref is coming from) to have been rendered. `useOnPropsChange` would not be able to handle that case.

`useOnPropsChange` will then trigger on all other renders and update scrollbar width if needed.

src/table/virtualized-table/hooks/use-scrollbar-width.ts
cbt1: It's actually a wrapper for this https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes and turns out be useful in this case as well.

PR #826, opened 2023-02-08
feat: field base selections in dropdown

a-m-dev:
This PR provides selection options directly in the columns dropdown menu. 
What not exists in this PR is having these selection menus inside of s sub-menu like designs. but it will be in the upcoming PR.

This PR requires https://github.com/qlik-oss/sn-table/pull/820 to be merged first! because it is based on that and that is the reason why there are lots of that change included here as well. 

I tried to cherry-pick the commit to a new branch but seems like the current main branch still using old implementation of listbox (inline version). so that kept me from doing it!

https://user-images.githubusercontent.com/29652890/217547077-54386c77-1d92-41b3-8cc5-b45876e26d9d.mov


src/locale/locales/en.json
```diff
@@ -55,6 +55,26 @@
     "value": "Search",
     "comment": "label for menu item to search in dimension."
   },
+  "SNTable.MenuItem.SelectAll": {
+    "value": "Select all",
+    "comment": "label for select all items in dimensions."
```
haxxmaxx, line 60: technically you are selecting values in a field

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -46,15 +65,89 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
 
+  const checkStateCountByKey = <T,>(keys: (keyof T)[], obj: T): boolean => {
+    return keys.some((key) => obj[key] > 0);
+  };
+
+  const runSelectionAnalysis = (layout: TableLayout) => {
+    const dimInfo = layout.qHyperCube.qDimensionInfo.find((dim) => dim.qFallbackTitle === column.label);
+    if (!dimInfo) return;
+    setSelectionAnalysis({
+      canSelectAll: checkStateCountByKey(['qOption', 'qAlternative', 'qExcluded', 'qDeselected'], dimInfo.qStateCounts),
```
haxxmaxx, line 137: shouldn't `canSelectAll` be always true?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -46,15 +65,89 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+      canSelectPossible: checkStateCountByKey(['qOption'], dimInfo.qStateCounts),
+      canSelectAlternative: checkStateCountByKey(['qAlternative'], dimInfo.qStateCounts),
+      canSelectExcluded: checkStateCountByKey(['qAlternative', 'qExcluded'], dimInfo.qStateCounts),
+    });
+  };
+
+  const handleOpenDropdown = async () => {
+    if (!model) return;
+    setOpenMenuDropdown(!openMenuDropdown);
+    model.getLayout().then((l) => runSelectionAnalysis(l as TableLayout));
```
haxxmaxx, line 148: So on first render, are you able to see that the options are all disabled? then some become enabled? If so, what happens when the next time you open it? I thinking there could be a problem with some options being enabled from last time it was open, and there is a minimal change to click something that has not been disabled yet

haxxmaxx: And actually, we might only want to get the layout when opening, not when closing. Then maybe something like this
```
if (!openMenuDropdown) {
  const layout =  await model.getLayout();
  runSelectionAnalysis(l as TableLayout);
}
setOpenMenuDropdown(!openMenuDropdown);

File: src/table/components/head/HeadCellMenu.tsx [hunk 4]
```diff
@@ -46,15 +65,89 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+                enabled: true,
+              },
+            ],
+            [
+              {
+                id: 1,
+                itemTitle: translator.get('SNTable.MenuItem.SelectAll'),
+                onClick: async () => {
+                  await fieldInstance?.selectAll();
+                  setOpenMenuDropdown(false);
```
haxxmaxx, line 77: any reason to await this? We might as well close the menu directly, no need for the selection call to finish?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,20 +1,39 @@
...
 export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
-  const { translator, embed } = useContextSelector(TableContext, (value) => value.baseProps);
+  const { app, translator, embed, model } = useContextSelector(TableContext, (value) => value.baseProps);
   const [openMenuDropdown, setOpenMenuDropdown] = useState(false);
   const [openListboxDropdown, setOpenListboxDropdown] = useState(false);
   const anchorRef = useRef<HTMLDivElement>(null);
   const listboxRef = useRef<HTMLDivElement>(null);
   const showSearchMenuItem = column.isDim && !column.isMasterItem;
+  const [fieldInstance, setFieldInstance] = useState<EngineAPI.IField | null>(null);
+  const [selectionAnalysis, setSelectionAnalysis] = useState<Record<string, boolean>>({
```
haxxmaxx, line 25: I would call this something like `enabledSelectionActions`

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,20 +1,39 @@
...
+    canSelectAll: false,
+    clearSelections: false,
```
haxxmaxx, line 27: `canClearSelections` or remove can from the others

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -46,15 +65,89 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
     ],
-    [translator, showSearchMenuItem]
+    [translator, showSearchMenuItem, fieldInstance, selectionAnalysis]
   );
 
+  const checkStateCountByKey = <T,>(keys: (keyof T)[], obj: T): boolean => {
+    return keys.some((key) => obj[key] > 0);
+  };
+
+  const runSelectionAnalysis = (layout: TableLayout) => {
```
haxxmaxx, line 133: looks like these only depend on `layout`, so I think they can be put in a util file

haxxmaxx, line 133: with the name suggestion above, I would call this something like `updateEnabledSelectionActions`

src/table/components/head/HeadCellMenu.tsx
a-m-dev: if you do a select all -> then select all being enabled becomes a bug! 
Screenshot 2023-02-08 at 16 08 35

src/table/components/head/HeadCellMenu.tsx
a-m-dev: awaiting because all of those function returns a promise in the type file. also of course it is async call. I'll move state change above it 👍

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: yes seems like that is the case for the nebula listbox as well

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -1,58 +1,172 @@
...
               },
             ],
           ]
         : []),
     ],
-    [translator, isDimension]
+    [translator, showSearchMenuItem, fieldInstance, enabledSelectionActions]
   );
 
+  const checkStateCountByKey = <T,>(keys: (keyof T)[], obj: T): boolean => {
```
cbt1, line 129: Pure function so can be declared outside of this component

src/table/components/head/HeadCellMenu.tsx
a-m-dev: what happens now is, i really cant see any items disabled at the first render, because `model.getLayout()` call happens really fast. I could reproduce your idea when i had explicitly called `model.getLayout()` after 2 sec inside of a setTimeout. and in that case, i was initially seeing it disabled and after 2 sec and getLayout call i was seeing the items being enabled. 

So, I'm thinking may be, we need to disable them all when we close the popover (and menu items will be fine after trying to open popover again because of getLayout call)? what do you think?

cbt1: Yes looks like something like that could be needed. I guess you could as keep a "isLoading" state and always disable them when that is true.

a-m-dev:
Just moved the entire logic for selection into a separate hook of `use-field-selection.ts` file. Now, it's better isolated and it became much more easier to test.

src/table/components/head/HeadCellMenu.tsx
a-m-dev: after implementing "disabling all when dropdown closes" it solved the issue. Now, every time you open dropdown -> call getLayout -> mean while items disabled (which is **not** noticeable) -> you get layout -> every thing is in order **and** when you close -> disabled them first -> then close dropdown => which literally fading dropdown and disabling happens at the same time and it's barely noticeable.

I think this cycle 👆 covers the need for `isLoading` state and that state would be needed in a scenario like we have dropdown open, and then selection changes -> which would require us to call getLayout **while** dropdown is still open.

So I think we don't need it until designs require us somehow! **OR** maybe we need it for the sub-menu case? (probably not either)

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -30,13 +43,25 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
     });
   };
 
+  useEffect(() => {
+    if (!openMenuDropdown) resetSelectionActionsEnabledStatus();
+  }, [openMenuDropdown]);
```
cbt1, line 48: This gives an eslint warning that I think you should fix

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -45,13 +70,65 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+                },
+                icon: <SelectExcluded />,
+                enabled: selectionActionsEnabledStatus.canSelectExcluded,
               },
             ],
           ]
         : []),
     ],
-    [translator, showSearchMenuItem]
+    [translator, showSearchMenuItem, fieldInstance, selectionActionsEnabledStatus]
```
cbt1, line 131: This gives an eslint warning that I think you should fix

cbt1 (approved):

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -28,6 +41,18 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
     embed.on('fieldPopoverClose', () => {
       setOpenListboxDropdown(false);
     });
+  }, [embed, column.fieldId]);
+
+  useEffect(() => {
+    if (!openMenuDropdown) resetSelectionActionsEnabledStatus();
+  }, [openMenuDropdown, resetSelectionActionsEnabledStatus]);
```
haxxmaxx, line 48: Late to the party here, but i don't see why you need this. It doesn't matter what the statuses are when the menu is closed. We only need to make sure they are correct when you open the menu

haxxmaxx: nvm now I remember. If they are in the wrong state if the getLayout is slow, they sholud be disabled and nothing else

PR #830, opened 2023-02-09
fix: column sort order issues

cbt1:
Fixed isses related to user dragging a dimension column to a different index:
- Single select did not work
- Multicell select did not work
- Totals value did not match column

### DIFF
--- a/src/table/utils/selections-utils.ts
+++ b/src/table/utils/selections-utils.ts
@@ -171,7 +171,7 @@ export const getMultiSelectedRows = (
     newSelectedRows[cell.qElemNumber] = cell.rowIdx;
     // also add the next or previous cell to selectedRows, based on which arrow is pressed
     const idxShift = evt.key === KeyCodes.DOWN ? 1 : -1;
-    const nextCell = pageRows[cell.pageRowIdx + idxShift][`col-${cell.colIdx}`] as Cell;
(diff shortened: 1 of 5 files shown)


src/table/utils/selections-utils.ts
```diff
@@ -171,7 +171,7 @@ export const getMultiSelectedRows = (
     newSelectedRows[cell.qElemNumber] = cell.rowIdx;
     // also add the next or previous cell to selectedRows, based on which arrow is pressed
     const idxShift = evt.key === KeyCodes.DOWN ? 1 : -1;
-    const nextCell = pageRows[cell.pageRowIdx + idxShift][`col-${cell.colIdx}`] as Cell;
+    const nextCell = pageRows[cell.pageRowIdx + idxShift][`col-${cell.pageColIdx}`] as Cell;
```
haxxmaxx, line 174: have you tried this with pagination?

cbt1: Yes see my other comment :) Not all that much but did test with a moved column, a non-moved column and using keyboard multiselect

a-m-dev (approved):
over all looks good, but not sure about which path should take on `colIdx` or `pageColIdx`.

haxxmaxx (approved):
I tested it and I can't break it. It was actually strange that it worked before. We set column.id to col-${pageColIdx} so this probably fixes some undiscovered bug 👍

PR #832, opened 2023-02-10
feat: keyboard navigation for new headers

haxxmaxx:
- keys to go between header cells, from header to body and vice versa
- tabbing to header elements

not in this PR:
- totals support
- logic for keyboard.enabled


src/table/components/head/__tests__/HeadCellContent.spec.tsx
```diff
@@ -114,15 +114,15 @@ describe('<HeadCellContent />', () => {
...
-  it('should not call changeSortOrder when clicking a header cell and cells are selected', () => {
-    selectionsAPI = {
-      isModal: () => true,
-    } as ExtendedSelectionAPI;
-    renderTableHead();
-    fireEvent.click(screen.getByText(column.label));
-
-    expect(changeSortOrder).not.toHaveBeenCalled();
-  });
+  // it('should not call changeSortOrder when clicking a header cell and cells are selected', () => {
```
haxxmaxx, line 117: not sure what to do on this one. We now check if selections are enabled using the context. So you would have to mock a selection to do that. Probably involving calling selection dispatch

haxxmaxx:
losing coverage due to skipping totals and not running a couple of functions in accessibility-utils (that contain no logic)

cbt1:
I'm getting an error thrown when I use the keys to navigate the header and do the following:
1. I have a table with 6 columns
2. Focus the first column header
3. Navigate using the keys the last colum and then continue by pressing right key, so that focus is on the first cell (0,0 in body)
4. Press the left key

Skärmavbild 2023-03-07 kl  15 16 07

src/table/components/head/HeadCellContent.tsx
```diff
@@ -6,25 +6,38 @@ import { FullSortDirection } from '../../constants';
...
 
 function HeadCellContent({ children, column, isActive, areBasicFeaturesEnabled }: HeadCellContentProps) {
-  const { constraints, selectionsAPI, keyboard, translator, changeSortOrder } = useContextSelector(
+  const { constraints, keyboard, translator, rootElement, changeSortOrder } = useContextSelector(
     TableContext,
     (value) => value.baseProps
   );
+  const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const isFocusInHead = useContextSelector(TableContext, (value) => value.focusedCellCoord[0] === 0);
+  const isInSelectionMode = useContextSelector(TableContext, (value) => value.selectionState.colIdx > -1);
```
cbt1, line 18: Why could you not use `selectionsAPI.isModal()`?

src/table/types.ts
```diff
@@ -150,10 +150,22 @@ export interface HandleHeadKeyDownProps {
   evt: React.KeyboardEvent;
   rootElement: HTMLElement;
   cellCoord: [number, number];
-  column: Column;
-  changeSortOrder: ChangeSortOrder;
+  setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>;
   isInteractionEnabled: boolean;
+  areBasicFeaturesEnabled: boolean;
+  // isLabelLast: boolean;
```
cbt1, line 156: Remove commented code

haxxmaxx:
> EDIT: Does not happen if I remove the totals row

yeah it is broken when you have a totals row at this point

src/table/components/head/HeadCellContent.tsx
haxxmaxx: because the header is memoed and does not update when you start selecting. But the selectionState on the context does. We need to change the tabindex of the header once you start selecting. This was the simplest way I could think of

cbt1: What do you mean by "memoed"? The component is not wrapped in `React.memo`? I think you mean `selectionsAPI` "property" is not updated when you start selecting?

You MIGHT be able to do something like:
```js
const isInSelectionMode = useContextSelector(TableContext, (value) => value.baseProps.selectionsAPI.isModal());
```

So that `selectionsAPI.isModel()` is still used everywhere we need to check if we are in "selection mode".

haxxmaxx: well the head wrapper component is memoed, so this one is indirectly. But looking at the selectionsAPI straight on the context is better. Then I think I can test it better as well!

cbt1 (approved):
LGTM

PR #834, opened 2023-02-10
test: use global measureText mock

cbt1:
All test will now use a global measureText mock.

Had to put in int a `beforeEach` callback as some test calls `jest.restoreAllMock()` which would reset it for all subsequent tests.

### DIFF
--- a/src/table/hooks/__tests__/use-column-widths.spec.ts
+++ b/src/table/hooks/__tests__/use-column-widths.spec.ts
@@ -5,18 +5,11 @@ import { TableStyling } from '../../types';
 import useColumnWidths from '../use-column-widths';
 
 describe('use-column-widths', () => {
-  let measureTextMock: jest.Mock<{ width: number }>;
   let columns: Column[];
   let tableWidth: number;
   let styling: TableStyling;
 
   beforeEach(() => {
-    measureTextMock = jest.fn();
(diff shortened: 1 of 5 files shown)


cbt1:
Forget to update the other tests that are also using a cusotm measureText mock so will do that now..

src/table/hooks/__tests__/use-column-widths.spec.ts
```diff
@@ -77,7 +66,6 @@ describe('use-column-widths', () => {
       });
 
       it('should return one size for hug and equal sizes for two columns with fill', () => {
-        measureTextMock.mockReturnValue({ width: 10 });
```
cbt1, line 80: Removing this did not actually change the outcome of the test so not sure that the intention was here.

cbt1:
This PR also removes all the error logs we got from tests that had issues with the canvas context

src/table/hooks/__tests__/use-column-widths.spec.ts
haxxmaxx: ahh, yes I noticed this to today. The width is smaller than the accepted min value, so that is why it returns 120. I have changed this in my PR

haxxmaxx (approved):
Mange takk

PR #835, opened 2023-02-10
fix: totals border color in vt mode

cbt1:
Fixes an issue where the totals border would not be colored correctly.

### DIFF
--- a/src/table/virtualized-table/Table.tsx
+++ b/src/table/virtualized-table/Table.tsx
@@ -1,6 +1,5 @@
 import React, { memo, useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { VariableSizeList } from 'react-window';
-import { getColumns, getTotalPosition } from '../../handle-data';
 import useColumnSize from './hooks/use-column-size';
 import Body from './Body';
 import FullSizeContainer from './FullSizeContainer';
@@ -17,7 +16,8 @@ import getHeights from './utils/get-height';
 import useScrollbarWidth from './hooks/use-scrollbar-width';
 
 const Table = (props: TableProps) => {
-  const { rect, pageInfo, paginationNeeded } = props;
+  const { rect, pageInfo, tableData } = props;
+  const { totalsPosition, columns, paginationNeeded } = tableData;
   const { layout, theme, styling } = useContextSelector(TableContext, (value) => value.baseProps);
   const ref = useRef<HTMLDivElement>(null);
   const headerRef = useRef<VariableSizeList>(null);
@@ -25,7 +25,6 @@ const Table = (props: TableProps) => {
   const bodyRef = useRef<BodyRef>(null);
   const innerForwardRef = useRef() as React.RefObject<HTMLDivElement>;
   const { xScrollbarWidth, yScrollbarWidth } = useScrollbarWidth(ref);
-  const totals = useMemo(() => getTotalPosition(layout), [layout]);
   const bodyStyle = useMemo<BodyStyle>(
     () => ({
       ...styling.body,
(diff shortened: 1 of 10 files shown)


src/table/virtualized-table/Table.tsx
```diff
@@ -17,24 +16,23 @@ import getHeights from './utils/get-height';
 import useScrollbarWidth from './hooks/use-scrollbar-width';
 
 const Table = (props: TableProps) => {
-  const { rect, pageInfo, paginationNeeded } = props;
+  const { rect, pageInfo, tableData } = props;
+  const { totalsPosition, columns, paginationNeeded } = tableData;
```
a-m-dev, line 20: I think you can combine these two lines. because it's already passed from your check at `index.js` and you are 100% sure that `totalsPosition` is returning an object of `{ atTop: boolean, atBottom: boolean }` at any case, so my suggestion would be: 
```jsx
const { rect, pageInfo, tableData: { totalsPosition, columns, paginationNeeded } } = props;
```

src/table/virtualized-table/Table.tsx
```diff
@@ -17,24 +16,23 @@ import getHeights from './utils/get-height';
...
     () => ({
       ...styling.body,
       background: theme.background.color ?? 'transparent',
     }),
     [layout, theme.name()] // eslint-disable-line react-hooks/exhaustive-deps
   );
-  const { headerRowHeight, bodyRowHeight, headerAndTotalsHeight } = getHeights(styling.head, bodyStyle, totals);
-  const columns = useMemo(() => getColumns(layout), [layout]);
+  const { headerRowHeight, bodyRowHeight, headerAndTotalsHeight } = getHeights(styling.head, bodyStyle, totalsPosition);
   const tableRect = useMemo(() => toTableRect(rect, paginationNeeded), [rect, paginationNeeded]);
```
a-m-dev, line 36: can't we move other stuff to `getTableData` function as well? like `tableRect`, `useColumnSize` and `useTableCount` ? seems like they somehow need some piece of layout data in their core and you already have layout in `getTableData`

src/table/virtualized-table/Wrapper.tsx
```diff
@@ -2,18 +2,18 @@ import React, { useMemo, useState } from 'react';
...
 import { MAX_PAGE_SIZE } from './constants';
 import useOnPropsChange from './hooks/use-on-props-change';
 import { TableContext, useContextSelector } from '../context';
 
 export default function Wrapper(props: WrapperProps) {
-  const { rect } = props;
-  const { layout, theme, constraints } = useContextSelector(TableContext, (value) => value.baseProps);
-  const totalRowCount = layout.qHyperCube.qSize.qcy;
+  const { rect, tableData } = props;
+  const { paginationNeeded, totalRowCount } = tableData;
```
a-m-dev, line 14: same here for `tableData`, but i think you probably don need them if you use `tableData.totalRowCount` and `tableData.paginationNeeded`. unless variable name is important in this case

src/table/virtualized-table/__tests__/Table.spec.tsx
```diff
@@ -87,9 +104,9 @@ describe('<Table />', () => {
   afterEach(() => jest.restoreAllMocks());
 
   it('should render', async () => {
-    renderTable();
+    await renderTable();
```
a-m-dev, line 107: Do we need await here because of act? if that is true, why we should await for act?

src/table/virtualized-table/Table.tsx
cbt1: I can but it would end up being 5 lines of code instead of 2 thanks to prettier so I will not change that

src/table/virtualized-table/Table.tsx
cbt1: We "can" but they are only used by the virtual scroll table so I don't think there is much value in it.

Currently `TableData` is used by the virtual scroll table because some of the shared code between the two tables depends on it. If that would have not been the case I think I would have gone with a different solution.

src/table/virtualized-table/Wrapper.tsx
cbt1: It the same story here. Prettier will turn 2 easly readable lines of code into 3 lines of code. Which IMO does not add any value.

Also I think this is just my personal preference that I prefer to write `paginationNeeded` over `tableData.paginationNeeded` as it's shorter

src/table/virtualized-table/__tests__/Table.spec.tsx
cbt1: `act` returns a Promise. So we wait for that to resolve. If not, the test would log errors telling us that wrap the call in  `act`. Don't fully understand as to why it's needed tho.

cbt1: Right, so quick rendering tells me that `await act` is needed because `Table` component does some async data fethcing in a `useEffect`.  So that should be a way to await for that data fetch to complete.

a-m-dev (approved):

PR #836, opened 2023-02-10
test: remove errors logged during test

cbt1:
When you run unit test some of them logged various errors. This PR attempts to fix the reason behind the errors, thus removing the errors logged.

### DIFF
--- a/src/table/components/body/__tests__/withSelections.spec.tsx
+++ b/src/table/components/body/__tests__/withSelections.spec.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import TableCell from '@mui/material/TableCell';
 
-import * as withSelections from '../withSelections';
+import withSelections from '../withSelections';
 import { Announce, ExtendedSelectionAPI, Cell, Column } from '../../../../types';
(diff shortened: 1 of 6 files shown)


src/table/components/body/__tests__/withSelections.spec.tsx
```diff
@@ -20,20 +20,26 @@ describe('withSelections', () => {
...
+              </HOC>
+            </tr>
+          </tbody>
+        </table>
       </TestWithProviders>
     );
 
   beforeEach(() => {
-    HOC = withSelections.default((props: CellHOCProps) => <TableCell {...props}>{props.children}</TableCell>);
+    HOC = withSelections((props: CellHOCProps) => <TableCell>{props.children}</TableCell>);
```
cbt1, line 42: It might seem odd that I have do not pass props here anymore. But it was never validated in any way so all it did was cause errors to be logged.

haxxmaxx: yeah these tests are a bit strange anyway

haxxmaxx (approved):
you are really spoiling RTL now, I liked giving it something to complain about 🙃

PR #837, opened 2023-02-10
test: add rendering tests for vt mode

cbt1:
Adds rendering test for the the virtual scroll table. I've most just copied the old scenarios and run those as well except in two cases:
- Scenario 1 has a "broken" layout where the qHyperCude defintion does not match the data from getHyperCudeData. So for virtual scroll the scenario is the same but with data from a different app.
- Scenario 12 was not applicable as scroll a item into view does not work the same.

Also added a new Scenario 14 with 1M data rows to trigger the pagination footer in virtual scroll table.

Change the initial "waitForSelector" so that both tables could use the same selector when waiting for the table to have rendered.


a-m-dev (commented):
are these screen shots correct? 
seems like when we have less rows in table, it renders a white space below rows but in other cases there are empty rows


Screenshot 2023-02-13 at 15 08 53
Screenshot 2023-02-13 at 15 08 38
Screenshot 2023-02-13 at 15 08 22

cbt1:
> are these screen shots correct? seems like when we have less rows in table, it renders a white space below rows but in other cases there are empty rows

@a-m-dev  Well yes and no. The first image is a new fixture that I created because the one used by the pagination table was "corrupt", so that is 100% correct in that it does not show any empty cells.

For the other two images, It's correct in the sense that the output is based on the layout and the result of the `getHyperCubeData` call. BUT here I just copied the existing fixtures that the pagination table use and they do not match how the virtual scroll table fetches data. The `layout` essentially says that that X number of rows exist but the `getHyperCubeData` only returns a handful of them and not enought to fill the screen. So what you see is the result of that.

I think we should revise the fixtures as they have a lot of overlapping scenarios. But I would consider that for another PR as it would change all rendering tests.

a-m-dev (approved):
we need to update test scenarios for vt

PR #839, opened 2023-02-13
test: run rendering tests in parallel

cbt1:
Each test suit must run on unique ports to not conflicts with each other

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -26,7 +26,7 @@ const config: PlaywrightTestConfig = {
   // The maximum number of retry attempts per test, two retries on CI
   retries: process.env.CI ? 2 : 0,
   // Limit the number of workers
-  workers: 1,
+  workers: 2,
   // Multiple "projects" can run your tests in multiple browsers and configurations
   projects: [
     {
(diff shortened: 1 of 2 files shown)


playwright.config.ts
```diff
@@ -26,7 +26,7 @@ const config: PlaywrightTestConfig = {
   // The maximum number of retry attempts per test, two retries on CI
   retries: process.env.CI ? 2 : 0,
   // Limit the number of workers
-  workers: 1,
+  workers: 2,
```
a-m-dev, line 29: I tried to use `os.cpus().length` to run test on multiple workers but seems like it only picks up 2 of them, any idea about why?

a-m-dev (approved):

playwright.config.ts
cbt1: @a-m-dev  Yes, there are only 2 test suits (files) that it can run in parallel. So it won't help by increasing the number of workers.

I did some test with running tests inside a test file also in parallel, that way you can increase the number of workers but found it too unstable. Also CI only have 2 CPU:s anyway there it would always run 2.

PR #840, opened 2023-02-13
fix: selection listener memleak

cbt1:
Noticed that while toggling between usePagination a few times an error was logged to console about a possible event memory leak.

Tracked it down to this useEffect that did not properly return the cleanup function.

### DIFF
--- a/src/__test__/test-with-providers.tsx
+++ b/src/__test__/test-with-providers.tsx
@@ -41,7 +41,11 @@ const TestWithProviders = ({
   app = { getField: () => Promise.resolve({}) } as unknown as EngineAPI.IApp,
   layout = generateLayout(1, 1, 5),
   constraints = {} as stardust.Constraints,
-  selectionsAPI = { isModal: () => false } as ExtendedSelectionAPI,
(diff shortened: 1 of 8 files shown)


a-m-dev (commented):
This is a bit confusing to me, more in this perspective of why we need to run `addSelectionListeners` while tearing down the component?

my guess is when you click `usePagination` checkbox fast, this functions take a bit time to finish it's work, Do we get selection listeners correctly in that case?

cbt1:
> `addSelectionListeners`

`addSelectionListeners` returns a clean-up function that removes the listeners from selections api. Since the clean-up function was not returned inside the useEffect the listeners never was removed.

It should probably just be a custom hook instead of a function that returns a clean-up. That would be easier to use and understand.

cbt1:
@a-m-dev I ended up refactoring this into a custom hook instead for clearity. But that did require a lot more changes in the code.

a-m-dev (approved):
✅ double great!

haxxmaxx:
I actually had this on my TODO list, should have been in Jira I guess. Great find!

PR #843, opened 2023-02-14
feat: VIZ-2338 property definition for dimension and measure

sharleneaqlik:
This PR holds the new property definition for dimension and measure for side-panel. The old components are rendered suing the definition for the time being.


sharleneaqlik:
duplicate

PR #844, opened 2023-02-14
fix: error on remove column

cbt1:
Fixes an issue where an error was throw after a dimension or measure was removed from the virtual scroll table.

I think this is a bit of a work around for the real issue which is that `tableData` is created based on the layout where the dimension/measure have been removed. BUT the due to how the`useContextSelector` appears to work, any cosumer of it will first get a "stale" layout (where the dimension/measure still exist) and than be re-rendered again with the updated layout.

Not really sure yet how to best solve this and prevent any further issue where you get stale vs non-stalve value.


cbt1:
[Here](https://github.com/dai-shi/use-context-selector/issues/69) is a thread on a npm package that implements a useContextSelector pattern. It mentions the double rendering as a limitation there so I doubt we'll find a way around that.

cbt1:
Closing this in favour of #847

PR #846, opened 2023-02-15
feat: selection actions in sub-menu

a-m-dev:
This PR puts selections actions into a sub-menu abstraction level, like below. 

Screenshot 2023-02-15 at 10 28 59

But the interesting part is the structure that it brings in to the menu itself. By leveraging recursive call into `<MenuList />` component, whenever we give submenus to a specific menu, we are being able to create as many groups deep as we can. 
Just need to provide the correct data in `<HeadCellMenu />` and it takes care of everything. 

https://user-images.githubusercontent.com/29652890/218988800-003d7576-7040-4ef5-b72e-6439daefe93e.mov


src/table/components/head/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { MenuItemGroup } from '../../../types';
+import MenuGroupWrapper from './MenuGroupWrapper';
+
+interface MenuListProps {
+  open: boolean;
+  anchorEl: HTMLDivElement | null;
+  onClose: () => void;
+  menuGroups: MenuItemGroup[];
+  transformOrigin?: PopoverOrigin;
+  anchorOrigin?: PopoverOrigin;
```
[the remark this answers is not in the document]
haxxmaxx: see my hard code comment

src/table/components/head/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const MenuList = ({
+  anchorEl,
+  open,
+  onClose,
+  menuGroups,
+  transformOrigin = { horizontal: 'left', vertical: 'top' },
+  anchorOrigin = { horizontal: 'left', vertical: 'bottom' },
```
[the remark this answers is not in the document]
a-m-dev: what we are giving to `MenuList` as **default values** -> it makes dropdown open from **bottom left**
but in case of sub menu -> it should open from **top right**

Actually, we can remove them! it still picks up default provided mui values and will work correctly!

src/table/components/head/MenuList/MenuList.tsx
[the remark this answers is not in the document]
a-m-dev: Never mind, we can remove them! it still picks up default provided mui values and will work correctly!

src/table/components/head/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,42 @@
...
+}: MenuListProps) => {
+  if (!menuGroups.length) return null;
+  return (
+    <Menu
+      className="sn-table-head-menu"
+      aria-labelledby="sn-table-head-menu-button"
```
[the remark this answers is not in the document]
a-m-dev: I think we can do something like giving an index as much as user goes deep in sub menus. 
something like:
```md
sn-table-head-menu-button-level-1 || menu name
sn-table-head-menu-button-level-2
sn-table-head-menu-button-level-3
...
```
It should be ok for us to assign different labels for `aria-labeledby` in different levels of submenu

src/table/components/head/MenuList/MenuList.tsx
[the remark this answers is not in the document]
a-m-dev: disabled that rule for those props

src/table/components/head/MenuList/MenuList.tsx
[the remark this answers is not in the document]
a-m-dev: Now it's only for main dropdown: 
Screenshot 2023-02-17 at 09 59 54

src/table/components/head/MenuList/RecursiveMenuList.tsx
```diff
@@ -0,0 +1,44 @@
...
+}: RecursiveMenuListProps) => {
+  if (!menuGroups.length) return null;
+  return (
+    <Menu
+      className="sn-table-head-menu"
+      open={open}
+      anchorEl={anchorEl}
+      onClose={onClose}
+      autoFocus={false}
+      {...(ariaLabel ? { [`aria-labelledby`]: ariaLabel } : {})}
```
cbt1, line 35: Can you not do like this?

```suggestion
     ariaLabelledby={ariaLabel} // It's either undefined or a string
```

src/table/components/head/MenuList/RecursiveMenuList.tsx
```diff
@@ -0,0 +1,44 @@
...
+      {...(anchorOrigin ? { anchorOrigin } : {})}
```
cbt1, line 36: Think these two can do the same as well?

a-m-dev: passing `undefined` (`anchorOrigin={anchorOrigin}`) causes below effect (MUI default anchor origin is top left of anchor ref):
https://user-images.githubusercontent.com/29652890/219617153-9ef6cc2d-d8a0-4a7f-b891-fe20e4e3e613.mov

So it will not work for popover related stuff but it does for `aria-labeledby`, will update it asap

cbt1: So not passing in anchorOrigin vs anchorOrigin=undefined gives different results. I'd say I did not expect that

src/table/components/head/MenuList/MenuGroup.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const MenuGroupItems = ({ id, onClick, itemTitle, icon, enabled, subMenus }: HeadCellMenuItem) => {
+  const [openMenu, setOpenMenu] = useState(false);
+  const anchorRef = useRef<HTMLDivElement | null>(null);
+
+  const handleOnClick = (evt: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+    if (onClick) onClick(evt);
+    if (subMenus?.length) {
+      setOpenMenu(true);
+      subMenusOpenStatusCache[itemTitle] = setOpenMenu;
```
cbt1, line 38: Would `id` not be a better key here? Not that I think we will run into duplicate item titles

a-m-dev: No i think, because each group can have an item with id of 1, 2 ...

a-m-dev: No i think, because each group can have an item with id of 1, 2 and so on. The thing that makes the keys of that object unique is going to be the titles because we cannot have a menu with same titles but each group can have duplicate ids

cbt1: Actually I wonder if this will work because `subMenusOpenStatusCache` will be shared by all sn-table instances

src/table/components/head/MenuList/RecursiveMenuList.tsx
a-m-dev: Yes, passing undefined fall into default behaviour of mui which we dont want

src/table/components/head/MenuList/MenuGroup.tsx
a-m-dev: but everything happens in scope of one table instance (in sense client for example)

cbt1: What I mean is that, even if you had 5 sn-table instances on a sheet and open the submenu for all of them. `const subMenusOpenStatusCache: SubMenusOpenStatusCache = {};` is only called once. So they all share the same cache.

But since they all have the same key (Selections) and can never be opened at the same time. It should not be a real issue. At least not for this case.

a-m-dev: Yes, since if there is one dropdown open already, user cannot open another dropdown from other table in a sheet => it will not going to be the case then!

cbt1: > also each cache is specific for one table, so if we have 5 table instances there will be 5 cache objects for each

No I don't think so. The creation of the cache is only run once. So ALL instances will share the same cache?

cbt1 (approved):
🚢  :shipit:

src/table/components/head/MenuList/MenuGroup.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type SubMenusOpenStatusCache = Record<string, React.Dispatch<React.SetStateAction<boolean>>>;
+const subMenusOpenStatusCache: SubMenusOpenStatusCache = {};
+
+const MenuGroupItems = ({ id, onClick, itemTitle, icon, enabled, subMenus }: HeadCellMenuItem) => {
+  const [openMenu, setOpenMenu] = useState(false);
+  const anchorRef = useRef<HTMLDivElement | null>(null);
+
+  const handleOnClick = (evt: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+    if (onClick) onClick(evt);
```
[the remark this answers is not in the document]
a-m-dev: we need to reset cache in this if check, so will keep it

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -77,53 +77,65 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+                      itemTitle: translator.get('SNTable.MenuItem.ClearSelections'),
+                      onClick: async () => {
+                        setOpenMenuDropdown(false);
+                        await fieldInstance?.clear();
+                      },
+                      icon: <ClearSelections />,
+                      enabled: selectionActionsEnabledStatus.canClearSelections,
+                    },
+                  ],
+                  [
```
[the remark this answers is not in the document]
a-m-dev: I am using this image below, it's from "column - search and select page"  figma:

Screenshot 2023-02-17 at 14 11 00

cbt1:
Still LGTM, cannot speak for the design spec.

haxxmaxx (approved):

PR #847, opened 2023-02-15
fix: table data out of sync

cbt1:
This PR puts the `tableData` on the context to make sure that it's in sync with the layout that it's based on. Only for Virtual scroll table to keep it smaller. Will create a new PR for pagination table.

But one of the bigger changes that I think was needed to make sense of the fix, was to change how `tableData` was created. It was wrapped in a promise that was only really applicable for the pagniation table, which had the side effect of creating double renders for the virtual scroll table.

So I ended up moving all logic that belong to each type of table (pagination, virtual scroll and carbon) into custom nebula hooks. With that it was easy to see which props was shared and which was not (ex. `tableData`).


haxxmaxx (approved):
very nice. This will make if very easy to remove the native code if it comes to that

PR #849, opened 2023-02-15
test: fix act error for `useFieldSelection`

a-m-dev:
fixed act error for `useFieldSelection`

root cause was this error here:https://github.com/qlik-oss/sn-table/blob/7734f4b60b38f2f675eafbd91c39d187cbe99140/src/table/hooks/use-field-selection.ts#L32-L35

because it was setting state after `app.getField(...)` being resolved!

### DIFF
--- a/src/table/hooks/__tests__/use-field-selection.spec.tsx
+++ b/src/table/hooks/__tests__/use-field-selection.spec.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { renderHook, act } from '@testing-library/react';
+import { renderHook, act, waitFor } from '@testing-library/react';
 import { Column, TableLayout } from '../../../types';
 import useFieldSelection from '../use-field-selection';
 import TestWithProviders from '../../../__test__/test-with-providers';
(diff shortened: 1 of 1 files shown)


src/table/hooks/__tests__/use-field-selection.spec.tsx
```diff
@@ -65,78 +65,92 @@ describe('useFieldSelection()', () => {
 
     it('`canSelectAll` and `canSelectPossible` and should be true after calling `updateSelectionActionsEnabledStatus` with `qOptions`', () => {
       const state = { qOption: 1 } as EngineAPI.INxStateCounts;
-      expect(triggerHook(state)).toMatchObject({
-        canSelectAll: true,
-        canClearSelections: false,
-        canSelectPossible: true,
-        canSelectAlternative: false,
-        canSelectExcluded: false,
+      waitFor(() => {
```
cbt1, line 68: I think you need an await before waitFor. Otherwise the test will never fail

cbt1 (approved):

PR #851, opened 2023-02-16
refactor: use table data from context

cbt1:


src/table/context/TableContext.tsx
```diff
@@ -37,7 +38,9 @@ export const TableContextProvider = ({
   tableWidth = 0,
 }: ContextProviderProps) => {
   const [headRowHeight, setHeadRowHeight] = useState(0);
-  const [focusedCellCoord, setFocusedCellCoord] = useState((cellCoordMock || [0, 0]) as [number, number]);
+  const [focusedCellCoord, setFocusedCellCoord] = useState(
```
cbt1, line 41: `[0, 0]` caused an extra render to occur every time the user changed page

haxxmaxx: nice catch

haxxmaxx (approved):

PR #853, opened 2023-02-16
fix: header alignment

haxxmaxx:
found some issues with the header content
- the inverse flex direction was not applied in vs
- the padding was not including the button padding in vs
- the auto margin was broken

for the record, the headers in vs does not work with multi line yet


cbt1 (approved):
LGTM but you do recall the reason to put StyledMenuIconButton in a div?

src/table/utils/styling-utils.ts
```diff
@@ -8,10 +8,11 @@ import { SELECTION_STYLING, COLORING } from '../styling-defaults';
 
 export const LINE_HEIGHT = 4 / 3;
 export const CELL_PADDING_HEIGHT = 8;
+export const HEAD_LABEL_BUTTON_PADDING = 8;
 export const CELL_BORDER_HEIGHT = 1;
 
 export const fontSizeToRowHeight = (fontSize: string) =>
-  parseInt(fontSize, 10) * LINE_HEIGHT + CELL_PADDING_HEIGHT + CELL_BORDER_HEIGHT;
+  parseInt(fontSize, 10) * LINE_HEIGHT + CELL_PADDING_HEIGHT + HEAD_LABEL_BUTTON_PADDING + CELL_BORDER_HEIGHT;
```
cbt1, line 15: Hmm think this will affect the row height of the body as well in VT

haxxmaxx: yes I pushed a fix for that

src/table/virtualized-table/utils/get-height.ts
```diff
@@ -1,11 +1,11 @@
 import { COMMON_CELL_STYLING } from '../../styling-defaults';
 import { GeneratedStyling } from '../../types';
 import { fontSizeToRowHeight } from '../../utils/styling-utils';
-import { MIN_BODY_ROW_HEIGHT, MIN_HEADER_HEIGHT } from '../constants';
+import { MIN_BODY_ROW_HEIGHT, MIN_HEADER_HEIGHT, PADDING_TOP_BOTTOM } from '../constants';
 import { BodyStyle, Rect, Totals } from '../types';
 
 const getHeaderRowHeight = ({ fontSize = COMMON_CELL_STYLING.fontSize }: GeneratedStyling) =>
-  Math.max(MIN_HEADER_HEIGHT, fontSizeToRowHeight(fontSize));
+  Math.max(MIN_HEADER_HEIGHT, fontSizeToRowHeight(fontSize) + PADDING_TOP_BOTTOM) * 2;
```
cbt1, line 8: Why `* 2`? Looks like the header height in the rendering tests are too large?

haxxmaxx: it should be ontSizeToRowHeight(fontSize) + PADDING_TOP_BOTTOM * 2) 🙃, already pushed it

PR #856, opened 2023-02-17
refactor: move pagination table

cbt1:
Moving the components for the pagination table into `src/table/pagination-table/components`.

Some components are shared by both tables, those remain in the `src/table/components` folder.

So the basic idea here is to have a folder structure that looks something like this:
- /src/table
  - /components/utils/hooks/ (shared code)
  - /pagination-table/ (only code used by this table)
  - /virtualized-table/ (only code used by this table)

This should mean that it will be easier to identify shared code and non-shared code. So that when you edit for example shared code, it should mean that both tables are affected and should be "regression tested" in the changed area.


src/table/pagination-table/components/head/ColumnAdjuster.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { useRef } from 'react';
-import { AdjusterProps } from '../../types';
-import { useContextSelector, TableContext } from '../../context';
-import { AdjusterHitArea, AdjusterHeadBorder, AdjusterBodyBorder } from './styles';
-import { ColumnWidthTypes, MIN_COLUMN_WIDTH, PAGINATION_HEIGHT } from '../../constants';
-import { BORDER_WIDTH } from '../../styling-defaults';
+import { AdjusterProps } from '../../../types';
```
haxxmaxx, line 2: the `ColumnAdjuster` will be used for the scrolling as well, so might as well move it now, the styles seem to be there already

haxxmaxx (approved):
move the column adjuster and it looks great, this is probably the best way to do it

src/table/pagination-table/components/head/ColumnAdjuster.tsx
cbt1: ✅

PR #860, opened 2023-02-20
feat: column resize for virtual scroll

haxxmaxx:


src/table/virtualized-table/TotalsCell.tsx
```diff
@@ -31,7 +34,10 @@ const TotalsCell = ({ index, style, data }: TotalsCellProps) => {
         display: 'flex',
         alignItems: 'center',
         borderStyle: 'solid',
-        borderWidth: isLastColumn ? '0px' : '0px 1px 0px 0px',
+        borderLeftWidth: '0px',
```
haxxmaxx, line 37: For some reason there seems to be a 3px border by default or inherited from somewhere. I couldn't see it anywhere in the devtools except under the layout tab

cbt1: Yeah I've noticed that as well, could be some default browser behaviour

src/table/virtualized-table/__tests__/Table.spec.tsx
```diff
@@ -100,6 +103,10 @@ describe('<Table />', () => {
     isModal = false;
 
     constraints = {};
+
+    rootElement = {
+      getBoundingClientRect: () => ({ height: rect.width }),
```
cbt1, line 108: A bit confusing have the height equal the width. Why not do set it from `rect.height`?

haxxmaxx: suppose to be height, fixed

cbt1 (approved):

PR #862, opened 2023-02-21
fix: empty cells on initial render

cbt1:
Fixes an issue where the scroll table would sometimes render with empty cells


src/table/virtualized-table/hooks/use-get-hypercube-data-queue.ts
```diff
@@ -9,6 +9,10 @@ const useGetHyperCubeDataQueue = (
   const queued = useRef(new Set<EngineAPI.INxPage>()); // Keep track of all unique pages that should be retrieved
   const ongoing = useRef(new Set<EngineAPI.INxPage[]>()); // Keep track of ongoing request. Should be aborted if page info or layout is changed
   const finished = useRef(new Set<string>()); // Keep track of all unqiue pages that have added to the queue
+  const mutableGetDataPages = useRef(getDataPages);
```
cbt1, line 12: useRef here to ensure that a single `queue` instance is used during the charts lifecycle. This should fix the race-condition issue that caused the empty cells.

haxxmaxx (approved):

PR #864, opened 2023-02-23
 feat: use initial data fetch

cbt1:
Enabled the `qInitialDataFetch` property by default when creating a new sn-table.

This allows the virtual scroll table to render instantly without first rendering a table with empty cells. It also improves the initial estimated row heights.

Pagination table does not use the qDataPages from the layout. It's something that could be done by not going to be super straight forward as it would need to figure out if the number of columns and rows in the layout is enough.

It should still be possible to use already created sn-table:s that does not have the`qInitialDataFetch` property set. Then it will work just like before.


src/table/virtualized-table/hooks/use-mutable-prop.ts
```diff
@@ -0,0 +1,10 @@
+import { useRef } from 'react';
+
+const useMutableProp = <T>(value: T) => {
+  const mutableRef = useRef<T>(value);
```
haxxmaxx, line 4: I don't really understand this. is it that the value is set on every render? That is why you set current to value on the next line?

cbt1: I'm afraid it's a hack.. I've been using this instead of [useEffectEvent](https://beta.reactjs.org/learn/separating-events-from-effects#declaring-an-effect-event), since that is only experimental.

haxxmaxx (approved):

PR #869, opened 2023-02-27
feat: dynamic row height for header and totals

haxxmaxx:
- Making sure that the row height increases when the text wraps. A bit more complex for headers since it depends on which icons are showing. Also need to take into account that the text is bold in the text measuring
- fixing a small padding issue in the `HeadCellContent`
- removing unused use-column-size files


src/table/virtualized-table/utils/cell-width-utils.ts
```diff
@@ -1,8 +1,20 @@
...
 const AMPLIFIED_PADDING = PADDING_LEFT_RIGHT + 2;
+const SORT_ICON_WIDTH = 20;
+const LOCK_ICON_WIDTH = 24;
+const MENU_ITEM_WIDTH = 28;
 
 export const subtractCellPaddingAndBorder = (width: number) => width - AMPLIFIED_PADDING * 2 - BORDER_WIDTH;
 
+export const subtractCellPaddingIconsAndBorder = (width: number, column: Column) => {
+  let newWidth = width - AMPLIFIED_PADDING * 2 - BORDER_WIDTH - SORT_ICON_WIDTH;
+  newWidth -= column.isLocked ? 24 : LOCK_ICON_WIDTH;
```
cbt1, line 14: `LOCK_ICON_WIDTH` is also 24?

haxxmaxx: fixed

src/table/virtualized-table/hooks/use-measure-text.ts
```diff
@@ -8,10 +8,14 @@ export interface MeasureTextHook {
...
-export default function useMeasureText(fontSize: string | undefined, fontFamily: string | undefined): MeasureTextHook {
+export default function useMeasureText(
+  fontSize: string | undefined,
+  fontFamily: string | undefined,
+  boldText?: boolean
+): MeasureTextHook {
   const { estimateWidth, measureText } = useMemo((): MeasureTextHook => {
     const context = document.createElement('canvas').getContext('2d') as CanvasRenderingContext2D;
-    context.font = `${fontSize} ${fontFamily}`;
+    context.font = `${boldText ? '600' : ''} ${fontSize} ${fontFamily}`;
```
cbt1, line 18: Why `600`? We use `bold`, which corresponds to a value of 700, as a styling attribute in `HeaderCell.tsx`

src/table/virtualized-table/utils/cell-width-utils.ts
```diff
@@ -1,8 +1,20 @@
...
+const SORT_ICON_WIDTH = 18;
+const LOCK_ICON_WIDTH = 24;
+const MENU_BUTTON_WIDTH = 28;
 
 export const subtractCellPaddingAndBorder = (width: number) => width - AMPLIFIED_PADDING * 2 - BORDER_WIDTH;
 
+export const subtractCellPaddingIconsAndBorder = (width: number, column: Column) => {
+  let newWidth = width - AMPLIFIED_PADDING * 2 - BORDER_WIDTH - SORT_ICON_WIDTH;
+  newWidth -= column.isLocked ? LOCK_ICON_WIDTH : 0;
+  newWidth -= column.isDim && !column.isMasterItem ? MENU_BUTTON_WIDTH : 0;
```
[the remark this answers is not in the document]
haxxmaxx: sure, it will only add some more whitespace anyway

src/table/virtualized-table/hooks/use-measure-text.ts
haxxmaxx: It actaully doesn't go through to the HeadCellContent. According to the designs it should be 600. I'll change the totals so that is 600 as well

haxxmaxx: I noticed now that there is no difference on the default font family, which the designs are based on. The change is more consistent though

cbt1 (approved):

PR #870, opened 2023-02-28
feat: listbox for master dims

a-m-dev:
this pr makes the listbox available for master dims

**we need to wait for a nebula bump ([this PR should be in that release](https://github.com/qlik-oss/nebula.js/pull/1088))**
---
https://user-images.githubusercontent.com/29652890/221845297-4757ea9d-9116-40e0-9a76-3023c78850e6.mov

### DIFF
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -74,7 +74,6 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
   } = info;
   const isHidden = qError?.qErrorCode === 7005;
   const isLocked = isDim && (info as ExtendedNxDimensionInfo).qLocked;
-  const isMasterItem = !!info.qLibraryId;
   const autoAlign = isDim ? 'left' : 'right';
 
   let fieldIndex = 0;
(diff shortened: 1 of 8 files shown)


src/handle-data.ts
```diff
@@ -97,6 +97,7 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
       columnWidth,
       id: `col-${pageColIdx}`,
       label: qFallbackTitle,
+      qLibraryId: isMasterItem ? info.qLibraryId : undefined,
```
haxxmaxx, line 100: isnt this the same as
```suggestion
      qLibraryId: info.qLibraryId,
```

haxxmaxx: I think you can remove isMasterItem, since we don't need that bool anymore. And we can always check if qLibraryId exists or not

a-m-dev: ✅ done!

src/handle-data.ts
```diff
@@ -97,6 +95,7 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
       columnWidth,
       id: `col-${pageColIdx}`,
       label: qFallbackTitle,
+      qLibraryId: info.qLibraryId,
```
haxxmaxx, line 98: We are destructing info above, you could do that for qLibraryId also

haxxmaxx (approved):

PR #873, opened 2023-03-01
fix: update cell height on column resize

cbt1:
This PR makes the cells in the body change height while the user is re-sizing a column:

https://user-images.githubusercontent.com/16608020/222409438-f2f02b3e-ac26-423b-bf30-ae65d87be19e.mov

Similar to how it behaves in the pagination table.


src/table/hooks/use-column-widths.ts
```diff
@@ -93,7 +93,7 @@ const useColumnWidths = (
 
   const [columnWidths, setColumnWidths] = useState(() => getColumnWidths(columns, tableWidth, getHugWidth));
 
-  useDidUpdateEffect(() => {
+  useOnPropsChange(() => {
```
cbt1, line 96: Using `useDidUpdateEffect` caused an flow of updates that triggered 3 renders each time the layout changed. The render cycles would look something like this
1. Stale layout and columnWidths is provied by`useSelectionContext`
2. The new layout is provided by `useSelectionContext` but the columnWidths remain stale
3. The new columnWidths is provided by `useSelectionContext`

By not wrapping the adjusting the state in a useEffect, the step 3 is removed.

src/table/virtualized-table/Body.tsx
```diff
@@ -102,6 +98,14 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
     gridRef.current.resetAfterIndices({ columnIndex: 0, rowIndex: 0, shouldForceUpdate: true });
   }, [layout, pageInfo.page, gridRef, columnWidth, rowMeta, theme.name()]); // eslint-disable-line react-hooks/exhaustive-deps
 
+  useLayoutEffect(() => {
+    gridRef.current?.scrollTo({ scrollTop: 0 });
+  }, [columnWidth, rowCount, columns.length, pageInfo]);
+
+  useOnPropsChange(() => {
+    updateCellHeight(rowsInPage);
```
cbt1, line 106: This is what triggers the cells to be re-calculated and re-rendered when the user resizes a column

src/table/hooks/use-column-widths.ts
```diff
@@ -93,7 +93,7 @@ const useColumnWidths = (
...
     setColumnWidths(getColumnWidths(columns, tableWidth - yScrollbarWidth, getHugWidth));
   }, [columns, tableWidth, yScrollbarWidth]);
```
cbt1, line 98: This is a bug here. Where it needs more deps that triggers a new column widht. But will fix in another PR.

src/table/virtualized-table/Wrapper.tsx
```diff
@@ -25,8 +24,15 @@ const Wrapper = (props: WrapperProps) => {
     [pageSize, page]
   );
 
-  // Reset page on new layout
-  useOnPropsChange(() => setPage(0), [layout]);
+  useOnPropsChange(() => {
```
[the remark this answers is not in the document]
cbt1: If that is the case then I think we should, so that they behave the same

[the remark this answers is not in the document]
cbt1: I have another PR that is based on this PR. Where I have move and restructured the page logic. It will make that change in that PR instead. Ok?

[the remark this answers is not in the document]
cbt1: So if I understand you correctly. If the data size changes, essentionally `qSize.qcy or qSize.qcx` changes. The pagination table should reset to page `0`?  I'm not seeing that behaviour or that kind of logic in the code

src/table/virtualized-table/Body.tsx
```diff
@@ -77,29 +73,31 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
     [rowsInPage, columns, bodyStyle, isHoverEnabled, maxLineCount]
   );
 
-  const bodyHeight = getBodyHeight(rect, headerAndTotalsHeight, deferredRowCount, estimatedRowHeight);
+  const bodyHeight = getBodyHeight(rect, headerAndTotalsHeight, rowCount, estimatedRowHeight);
 
-  useEffect(() => {
+  useLayoutEffect(() => {
```
a-m-dev, line 78: is below `useLayoutEffect` trigger, not syncing height on every render? if so, do we still need this one too?

src/table/virtualized-table/Header.tsx
```diff
@@ -6,8 +6,9 @@ import { useContextSelector, TableContext } from '../context';
 import useResetHeader from './hooks/use-reset-header';
 
 const Header = (props: HeaderProps) => {
-  const { rect, forwardRef, columns, columnWidths, pageInfo, headerStyle, rowHeight } = props;
+  const { rect, forwardRef, columns, pageInfo, headerStyle, rowHeight } = props;
   const { layout } = useContextSelector(TableContext, (value) => value.baseProps);
+  const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
```
a-m-dev, line 11: may be silly question but, why not getting both at the same call to `useContextSelector`? is it something about re-rendering? if so, do we need to care about it at this level?

src/table/virtualized-table/Body.tsx
cbt1: That is correct and intended. The first `useLayoutEffect` (with deps) tells `syncHeigt` to forcefully run no matter what. Sort like a `forceUpdate` call.

The second `useLayoutEffect` is using a guard inside that syncHeight function, so that in most renders, it does not actually do anything. Only given some condition is meet does it update a state.

src/table/virtualized-table/Wrapper.tsx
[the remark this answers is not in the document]
a-m-dev: I was suspected that there might be bug with `useOnPropsChange` regarding using referencing object values as deps, and i think i was not mistaken. 

I think it should be fine with primitive data types but as you can see for object values => if you change something deep inside, it will always detect `hasChanged` as true!

Screenshot 2023-03-06 at 09 53 27

src/table/virtualized-table/hooks/use-data/index.ts
```diff
@@ -37,7 +36,7 @@ const useData = (
   const mutableSetCellSize = useMutableProp<SetCellSize>(setCellSize);
   const memoizedToRows = useCallback(
     (qDataPages: EngineAPI.INxDataPage[], prevState?: Row[]) => {
-      const nextState = Array.isArray(prevState) ? [...prevState] : createEmptyState(rowCount);
+      const nextState = Array.isArray(prevState) ? prevState.slice(0, rowCount) : createEmptyState(rowCount);
```
a-m-dev, line 39: could we use `gridState.current.overscanRowStopIndex` (Stop index) instead of `rowCount` or That is not intended for this use case here?

src/table/virtualized-table/Header.tsx
cbt1: I do think that because of how `useContextSelector` works. It triggers a re-render that is tighly coupled with what the callback function returns. So every time the callback function returns a new value. The component is re-rendered.

So you cannot really get both values in the same call, as something like `const { columnWidths, baseProps } = useContextSelector(TableContext, (value) => value);` would always return trigger a re-render as `value` is always a new object on every render.

If that is what you are asking?

src/table/virtualized-table/hooks/use-data/index.ts
cbt1: No that would not be possible. The "stop index" represent only what react-window has rendered on it's last render. Wheres `rowCount` is the total number of rows in the table (for a page).

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -84,7 +78,14 @@ const useDynamicRowHeight = ({
 
   const setCellSize = useCallback(
     (text: string, rowIdx: number, colIdx: number) => {
+      const key = `${rowIdx}-${colIdx}`;
+      if (rowMeta.current.measuredCells.has(key)) {
```
a-m-dev, line 82: Do you need this guard while using Set? is it for avoiding replacement of keys old value?

src/table/virtualized-table/Wrapper.tsx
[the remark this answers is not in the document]
haxxmaxx: You're right, the logic isn't like that. If you end up on an empty page, you go back to the first one. So it is just a simpler version of what you have here. When needed, you should run setPage(0)

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -103,10 +104,43 @@ const useDynamicRowHeight = ({
...
-    (index: number) => rowMeta.current.heights[index] ?? estimatedRowHeight,
+    (rowIdx: number) => rowMeta.current.heights[rowIdx] ?? estimatedRowHeight,
     [rowMeta, estimatedRowHeight]
   );
 
+  /**
+   * Some user actions and events can trigger row heights to be invalidated
+   * - A column is re-sized
+   * - New layout (ex. due to property changes or outside selections)
+   * - Change of page (which would load new data)
```
a-m-dev, line 115: If i'm not mistaking, you mean scrolling right? (which would trigger a fetch of new page data in VT)

cbt1: No scrolling is different event handled by `src/table/virtualized-table/hooks/use-items-rendered-handler.ts`.

VT can still require multiple pages if the number of rows exceeds 250k.

src/table/virtualized-table/Wrapper.tsx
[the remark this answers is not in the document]
haxxmaxx: and yes leave it to the next pr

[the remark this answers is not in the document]
cbt1: @a-m-dev In your example code, you assign `newProp` with a new object? In which case it should detect that it has changed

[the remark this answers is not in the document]
a-m-dev: Yes, I mean if something has changes in layout, you will receive a new entire layout object and if that was not the case, still the result will be the same while something changes inside of same object 👇


Screenshot 2023-03-06 at 10 42 08

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -103,10 +104,43 @@ const useDynamicRowHeight = ({
...
+    if (!gridState) return;
+
+    const { overscanRowStartIndex: rowStart, overscanRowStopIndex } = gridState.current;
+    const rowStop = Math.min(overscanRowStopIndex, layout.qHyperCube.qSize.qcy - 1);
+
+    for (let rowIdx = rowStart; rowIdx <= rowStop; rowIdx++) {
+      const row = rows[rowIdx] ?? {};
+      Object.values(row).forEach((cell) => {
+        if (typeof cell === 'object') {
+          setCellSize(cell.qText ?? '', rowIdx, cell.pageColIdx);
```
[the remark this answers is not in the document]
cbt1: To clarify a concepts here. `estimatedRowHeight` is the average height of all measured rows. There is an assumption that all other rows that we have not measured are going to be roughtly equal to the rows we already have measured. This is a crude approach and there are multiple scenarios that is just not going to be the case. But generally it still enough for `react-window` to behave nicely.

If each cell on a row gets smaller line height (3 -> 2 -> 1), the row height for that row. Will be "3 lines". So `setEstimatedRowHeight` would be called for each measured cell. But the value passed to `setEstimatedRowHeight` would not change after the "highest" cell had been measured. 

I could avoid the costly operation if I knew which column was being re-sized. Which I don't, but I think it would be possible to expose it, ex. from the `ColumnAdjuster` component.

src/table/virtualized-table/Header.tsx
a-m-dev: Yes, that makes sense then 👍

src/table/virtualized-table/Wrapper.tsx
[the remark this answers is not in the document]
cbt1: This is still the correct behavior? `oldProp[0] === newProp[0]; // Are not the same`.

So I've created some unit tests for`useOnPropsChange` locally and it behaves as I would expect. So mayby I'm missing your point?

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
a-m-dev: I was asking because you've used Set for `measuredCells` and since that would not store duplicate values then thinking that check might be not needed, but since it is for preventing re-calc then it's great to have that check ✅

src/table/virtualized-table/Wrapper.tsx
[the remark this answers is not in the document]
a-m-dev: That is correct, my point is in above example, if you update the qcy with **1**, it still going to detect as object has  changed. 
(sorry for wrong example 😅)

Screenshot 2023-03-06 at 10 56 48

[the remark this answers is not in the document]
cbt1: That is what I want. If the property is a new layout it should trigger the callback function. There are cases where we could say pass in `layout.qHyperCube.qSize.qcy` directly as a dependency instead, but it tends to be hard to know which layout change should trigger a change or not. So I mostly just use `layout` directly to be on the safe side.

a-m-dev (approved):
nice 👍 🚀

PR #875, opened 2023-03-01
fix: move header and totals row heights to table level

haxxmaxx:
- create a new hook that keeps track of totals and header row height. Use this instead of `getHeights`
- rename `columnWidth` to `columnWidths` since it is an array of multiple widths


cbt1:
This is a `refactor` and not a `fix` right?

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
```diff
@@ -0,0 +1,75 @@
...
+  pageInfo: PageInfo;
+  totalsPosition: TotalsPosition;
+  headerRef: React.RefObject<VariableSizeList<any>>;
+  totalsRef: React.RefObject<VariableSizeList<any>>;
+}
+
+const getBodyRowHeight = ({ fontSize = COMMON_CELL_STYLING.fontSize }: GeneratedStyling) =>
+  Math.max(MIN_BODY_ROW_HEIGHT, fontSizeToRowHeight(fontSize));
+
+const useHeaderAndTotalsHeight = ({
```
cbt1, line 23: Since it also includes the body height. Better name it something more generic? Ex. "useHeights"

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
```diff
@@ -0,0 +1,75 @@
...
+  columns,
+  columnWidths,
+  pageInfo,
+  headerRef,
+  totalsRef,
+  totalsPosition,
+}: UseHeaderAndTotalsHeightProps) => {
+  const { layout, styling } = useContextSelector(TableContext, (value) => value.baseProps);
```
cbt1, line 31: We should move this into the `useDynamicRowHeight` hook. Well maybe need to keep styling here too but that would skip us having to pass two props to the hook

haxxmaxx:
> This is a `refactor` and not a `fix` right?

it is to fix the problem where the body rect is too big, so the header or totals can overflow

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
haxxmaxx: that we can do

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
haxxmaxx: ah yes, I added the bodyHeight after naming it. I'll change it

a-m-dev (approved):
Code looks LGTM

cbt1:
By moving the logic to the table component. Re-sizing the column will now also re-render the Table component while the interaction is taking place.

Maybe that we the fix? Otherwise the solution do seem to effect the performance in a negative way

src/table/virtualized-table/utils/get-body-height.ts
```diff
@@ -0,0 +1,18 @@
+import { Rect } from '../types';
```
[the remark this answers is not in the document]
cbt1: Why not?

src/table/virtualized-table/types/index.ts
```diff
@@ -58,7 +58,7 @@ export interface HeaderProps {
   pageInfo: PageInfo;
   forwardRef: React.RefObject<VariableSizeList<any>>;
   columns: Column[];
-  columnWidth: number[];
+  columnWidths: number[];
```
cbt1, line 61: Should say that I have removed columnWidths as a property from these component in my branch. They are now coming from the useSelectorContext instead. Which I think makes more sense

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
haxxmaxx: since we are using the useDynamicRowHeight twice, we need to pass different styles. So I'll only move layout

src/table/virtualized-table/types/index.ts
haxxmaxx: ok sure, that works for me

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
```diff
@@ -0,0 +1,75 @@
...
+    pageInfo,
+    boldText: true,
+  });
+
+  useEffect(() => {
+    columns.forEach((col, idx) => {
+      headerHeight.setCellSize(col.label, 0, idx);
+      totalsHeight.setCellSize(col.totalInfo, 0, idx);
+    });
+  }, [columns, headerHeight, totalsHeight]);
```
cbt1, line 59: You might as well remove the `useEffect` as `headerHeight` and `totalsHeight` will be new instances on every render. I think all this useEffect will do is trigger a stale render first

src/table/virtualized-table/utils/get-body-height.ts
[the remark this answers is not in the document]
haxxmaxx: just more a question if it fits in to your idea of the file structure 🙃  if it is fine with you then we are good

src/table/virtualized-table/hooks/use-header-and-totals-height.ts
haxxmaxx: cool I'll change

src/table/virtualized-table/utils/get-body-height.ts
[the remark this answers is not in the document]
cbt1: I tend to love files with just a single function ;)

cbt1 (approved):

PR #876, opened 2023-03-02
feat: in edit mode disable head menu button

LiKang6688:


cbt1 (approved):
LGTM but I would think hiding the menu and sort buttons would be where we want to go

haxxmaxx (changes requested):
I'm with bengtsson on this one. It just shouldn't show at all. which makes it easier, since you can just not render it

PR #877, opened 2023-03-03
fix: totals label on first col

haxxmaxx:
The first if checks if it is a measure, meaning we only need to check if the column is the one furthest left, which is pageColIdx === 0

### DIFF
--- a/src/__test__/handle-data.spec.ts
+++ b/src/__test__/handle-data.spec.ts
@@ -205,17 +205,17 @@ describe('handle-data', () => {
       expect(getTotalInfo(layout, 3, 3, 2)).toBe('');
     });
 
-    it('should return new total label value', () => {
+    it('should return specified total label', () => {
       layout.totals.label = 'Whatever';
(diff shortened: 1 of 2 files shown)


src/__test__/handle-data.spec.ts
```diff
@@ -214,6 +214,10 @@ describe('handle-data', () => {
       layout.totals.label = '';
       expect(getTotalInfo(layout, 0, 0, 2)).toBe('');
     });
+
+    it('should return first dimension total value as Totals when colIdx is not 0', () => {
```
a-m-dev, line 218: Can we get Total values on dimensions? I was thinking totals row is only for when we have measures

cbt1: That would be correct I think. So it should also be a condition that measures.length > 0

LiKang6688:
It does not cause any issues, right?

cbt1 (approved):

PR #879, opened 2023-03-03
feat: hide head sort buttons/head cell menu buttons when they are not needed

LiKang6688:
- should hide the head sort button/head menu button when constraints.active is true in edit mode
- should hide the head sort button/head menu button when clicking a header cell and cells are selected

Screenshot 2023-03-03 at 11 31 26


haxxmaxx:
Not what we want to do. We want to hide the sort icon and hide the ... menu button. We still want to keep the label button, disabling it somehow

PR #881, opened 2023-03-03
fix: copy cell only on cells

a-m-dev:
only show copy cell value in context menu in 'TD' and 'TH' elements


LiKang6688:
LGTM, shall we add an integration test with React-Testing-Library?

cbt1 (changes requested):
Does not work in virtual scroll table. Table element is not used there. You will need to check something like a class name

src/nebula-hooks/use-context-menu.ts
```diff
@@ -2,10 +2,13 @@
 import { onContextMenu } from '@nebula.js/stardust';
 import copyCellValue from '../table/utils/copy-utils';
 
+const tagNamesWhiteList = new Set(['TD', 'TH']);
+
 export default function useContextMenu(areBasicFeaturesEnabled: boolean) {
   onContextMenu?.((menu: any, event: any) => {
     areBasicFeaturesEnabled &&
       event.target &&
+      tagNamesWhiteList.has(event.target.closest('.sn-table-cell')?.tagName) &&
```
haxxmaxx, line 11: I think the class should be enough, all cells should have it

haxxmaxx (approved):

LiKang6688 (approved):

cbt1 (approved):

PR #885, opened 2023-03-06
refactor: extract functions from `handle-key-press` and `accessibility-utils`

haxxmaxx:
- rename handle-key-press to handle-keyboard
- create new file keyboard-utils and move all functions from handle-keyboard to that file, that are not event handlers
- craete new file get-element-utils that and move all non-accessibility related functions from accessiblity-utils there
- create enum for focus types


src/table/utils/get-element-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    top: number;
+    bottom: number;
+  } = { top: 0, bottom: 0 }
+): [number, number] => {
+  const rowCount = rootElement.getElementsByClassName('sn-table-row').length;
+  const columnCount = rootElement.getElementsByClassName('sn-table-head-cell').length;
+  let [nextRow, nextCol] = cellCoord;
+
+  switch (evt.key) {
+    case KeyCodes.DOWN:
```
haxxmaxx, line 20: only thing new here is using the enum

LiKang6688 (approved):
Are you ready to be reviewed? The title is not correct. The refactoring looks good to me

haxxmaxx:
> Are you ready to be reviewed? The title is not correct. The refactoring looks good to me

yes, I changed the title now

PR #886, opened 2023-03-06
feat: use keyboard to resize column

LiKang6688:
https://user-images.githubusercontent.com/4471489/223090631-eceeab86-3e29-4c11-a65d-fecb0abb7938.mov


cbt1:
Have you tested this with the virtual scroll table?

LiKang6688:
> Have you tested this with the virtual scroll table?

Yes, the difference is looking:
Screenshot 2023-03-06 at 12 31 08

Screenshot 2023-03-06 at 12 31 38

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -56,13 +56,51 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
+    if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
+      const initWidth = columnWidths[pageColIdx];
+      tempWidths.current.initWidth = initWidth;
+      const columnWidth = event.key === 'ArrowLeft' ? initWidth - 5 : initWidth + 5;
+      const adjustedWidth = Math.max(columnWidth, MIN_COLUMN_WIDTH);
+      tempWidths.current.columnWidth = adjustedWidth;
+
+      const newColumnWidths = [...columnWidths];
+      newColumnWidths[pageColIdx] = adjustedWidth;
+      setColumnWidths(newColumnWidths);
```
haxxmaxx, line 78: This is duplicate code, should be able to create a `updateWidth` function

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -56,13 +56,51 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
+    } else if (event.key === ' ' || event.key === 'Enter') {
+      event.stopPropagation();
+      event.preventDefault();
+      if (tempWidths.current.columnWidth !== tempWidths.current.initWidth) {
+        const newWidthData = { type: ColumnWidthTypes.PIXELS, pixels: tempWidths.current.columnWidth };
+        applyColumnWidths(newWidthData, column);
```
haxxmaxx, line 84: same here, this could be a `confirmWidth` function that can be reused

src/locale/locales/en.json
```diff
@@ -79,6 +79,14 @@
     "value": "Select excluded",
     "comment": "label for select excluded items in field."
   },
+  "SNTable.MenuItem.ChangeColumnPosition": {
```
haxxmaxx, line 82: this is not used, pls remove

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -56,13 +56,51 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
     document.addEventListener('mouseup', mouseUpHandler);
   };
 
+  const focusHeadMenuButton = (event: React.KeyboardEvent | React.FocusEvent<HTMLDivElement>) => {
```
haxxmaxx, line 59: this would make sense if it lives in a11y utils file

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -56,13 +56,51 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
+    const target = event.target as HTMLDivElement;
+    target.setAttribute('tabIndex', '-1');
+    const headMenuButton = target
+      ?.closest('.sn-table-cell')
+      ?.querySelector('#sn-table-head-menu-button') as HTMLElement;
```
haxxmaxx, line 64: this is a HTMLButtonElement

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -57,6 +58,16 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
     setOpenMenuDropdown(!openMenuDropdown);
   };
 
+  const setFocusOnClosetColumnAdjuster = () => {
```
haxxmaxx, line 61: this would also make sense it a11y utils

cbt1:
> > Have you tested this with the virtual scroll table?
> 
> Yes, the difference is looking...

Yeah that is a different issue. But great that it works in virtual scroll as well then 👍

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -57,6 +58,16 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+    setTimeout(() => {
+      const adjusterHitArea = anchorRef.current
+        ?.closest('.sn-table-cell')
+        ?.querySelector('#adjuster-hit-area') as HTMLElement;
+      adjusterHitArea?.setAttribute('tabIndex', '0');
```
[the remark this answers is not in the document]
LiKang6688: To be able to focus on it, I have to set tabIndex to 0. The default value is -1.

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -56,13 +56,51 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
+      event.stopPropagation();
+      event.preventDefault();
+      if (tempWidths.current.columnWidth !== tempWidths.current.initWidth) {
+        const newWidthData = { type: ColumnWidthTypes.PIXELS, pixels: tempWidths.current.columnWidth };
+        applyColumnWidths(newWidthData, column);
+      }
+      focusHeadMenuButton(event);
+    } else if (event.key === 'Escape') {
+      focusHeadMenuButton(event);
+    }
```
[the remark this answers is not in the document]
LiKang6688: IMO, it is the purpose of the esc case which does not do the update. Otherwise, it is the same as the space bar or enter.

src/table/components/head/ColumnAdjuster.tsx
LiKang6688: Sure, I will do that after https://github.com/qlik-oss/sn-table/pull/885 merged

src/table/components/head/ColumnAdjuster.tsx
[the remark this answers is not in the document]
haxxmaxx: in that case we would need to reset it to the init value, if it is set correctly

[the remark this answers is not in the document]
LiKang6688: Right, good point, I will do that

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -56,13 +64,39 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
     document.addEventListener('mouseup', mouseUpHandler);
   };
 
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === KeyCodes.LEFT || event.key === KeyCodes.RIGHT) {
+      const RESIZE_DISTANCE = 5;
+      const initWidth = columnWidths[pageColIdx];
```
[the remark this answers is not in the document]
LiKang6688: sure

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -139,6 +141,20 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
                 ],
               },
             ],
+            [
+              {
```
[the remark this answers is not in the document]
LiKang6688: Good point.I will fix it.

src/table/utils/accessibility-utils.ts
```diff
@@ -5,6 +5,31 @@ import { DEFAULT_FOCUS_CELL_COORD, FocusTypes } from '../constants';
 import { CellFocusProps, HandleResetFocusProps } from '../types';
 import { findCellWithTabStop, getCellElement, getNextCellCoord } from './get-element-utils';
 
+/**
+ * Add the tab stop for adjuster hit area and focus that
+ */
+export const setFocusOnClosetColumnAdjuster = (anchorRef: React.RefObject<HTMLDivElement>) => {
+  setTimeout(() => {
```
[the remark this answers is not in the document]
LiKang6688: Without setTimeout,  adjusterHitArea?.focus() does not work. any suggestion for a better solution?

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -139,8 +141,37 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+                id: 1,
+                itemTitle: translator.get('SNTable.MenuItem.AdjustColumnSize'),
+                onClick: (evt: React.MouseEvent<HTMLLIElement>) => {
+                  evt.stopPropagation();
```
a-m-dev, line 149: I think we can use `preventDefaultBehavior` function for menu items

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -21,48 +23,84 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
 
     tempWidths.current = {
       initX: evt.clientX,
       initWidth: columnWidths[pageColIdx],
       columnWidth: columnWidths[pageColIdx],
     };
+
     document.addEventListener('mousemove', mouseMoveHandler);
-    document.addEventListener('mouseup', mouseUpHandler);
+    document.addEventListener('mouseup', (event) => mouseUpHandler(event, adjuster));
```
haxxmaxx, line 67: I don't think this will remove the listener properly later. Pretty sure it has to be the same instance of the callback for the remover to find it

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -139,8 +141,37 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
...
+                  setOpenMenuDropdown(false);
+                  setFocusOnClosetColumnAdjuster(anchorRef);
+                },
+                icon: <ColumnSize />,
+                enabled: true,
+              },
+            ],
           ]
-        : []),
+        : [
```
haxxmaxx, line 159: why not just add the object last in the array?

src/table/utils/accessibility-utils.ts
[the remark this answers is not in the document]
haxxmaxx: not really, @cbt1 ?

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -21,48 +23,84 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
     }
   };
 
+  const mouseMoveHandler = (evt: MouseEvent) => {
+    const deltaWidth = evt.clientX - tempWidths.current.initX;
+    const adjustedWidth = Math.max(tempWidths.current.initWidth + deltaWidth, MIN_COLUMN_WIDTH);
+    updateWidth(adjustedWidth);
+  };
+
+  const mouseUpHandler = (evt: MouseEvent, adjuster: HTMLDivElement) => {
```
haxxmaxx, line 46: do we really need to pass the adjuster? Can you use the evt.target or is that sometimes the wrong element?

LiKang6688: that sometimes is the wrong element

src/table/components/head/HeadCellMenu.tsx
LiKang6688: Not sure I understand.  `showSearchMenuItem ? []: []`

src/table/utils/accessibility-utils.ts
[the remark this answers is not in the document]
a-m-dev: the reason is `focusHeadMenuButton()` called once 
without settimout it first sets tabindex to 0,
then it sets again to -1 by calling `focusHeadMenuButton`
I'm trying to see how can fix this

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -21,48 +23,84 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
+    preventDefaultBehavior(evt);
+    document.removeEventListener('mousemove', mouseMoveHandler);
+    document.removeEventListener('mouseup', () => mouseUpHandler(evt, adjuster));
+
+    confirmWidth();
+    adjuster.blur();
```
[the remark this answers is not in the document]
LiKang6688: Yes, it triggers handleBlur and the focus will be on the menu button. But there is no indication that the focus is on the menu visually.

src/table/components/head/ColumnAdjuster.tsx
LiKang6688: Fixed.

src/table/components/head/HeadCellMenu.tsx
haxxmaxx: yeah but that then spread into an array so you could do this
```
() => [
      ...(showSearchMenuItem
        ? [
          ...
          ]
        : []),
      {columnWidthItme},
    ],
```

src/table/components/head/ColumnAdjuster.tsx
haxxmaxx: ok then could we store it on the `tempWidths` ref, so you don't have to pass it to the mouseup handler and risk a memory leak

src/table/components/head/ColumnAdjuster.tsx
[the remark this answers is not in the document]
haxxmaxx: ah then that I think is fine

src/table/components/head/HeadCellMenu.tsx
LiKang6688: true

src/table/components/head/HeadCellMenu.tsx
LiKang6688: very true

src/table/utils/accessibility-utils.ts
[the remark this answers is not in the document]
a-m-dev: 👆 it's been called from `onBlur` of Adjuster, need a way to prevent it

src/table/components/head/ColumnAdjuster.tsx
LiKang6688: Sure. Done

src/table/utils/accessibility-utils.ts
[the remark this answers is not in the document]
a-m-dev: well, after selecting something from dropdown, the ... menu disappears when using settimeout => which means mui cannot set the focus back into the ... menu and (I guess) ignores it silently. 

**when not using settimeout** => ... menu still is there => which mui sets focus back to ... menu and overrides what ever we do in `setFocusOnClosetColumnAdjuster`

src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -16,53 +18,93 @@ const ColumnAdjuster = ({ column, isLastColumn }: AdjusterProps) => {
...
     }
+    adjusterHitArea.blur();
   };
 
   const mouseDownHandler = (evt: React.MouseEvent) => {
     // evt.preventDefault();
     evt.stopPropagation();
 
     tempWidths.current = {
+      adjusterHitArea: evt.target,
```
haxxmaxx, line 64: I would say better to do `evt.target.closest('#adjuster-hit-area')` to remove the logic in `mouseUpHandler`

src/table/utils/accessibility-utils.ts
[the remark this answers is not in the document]
haxxmaxx: I'm fine with this for now

a-m-dev:
✅ The solution is elegant and it could become more perfect if we could be able to skip setTimeout. but if everyone is ok we can keep it for now, just to unblock the path for picking up new tasks.

src/table/utils/accessibility-utils.ts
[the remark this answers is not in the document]
cbt1: @a-m-dev Seems possible

a-m-dev (approved):

haxxmaxx (approved):

PR #888, opened 2023-03-07
fix: add default value font size

eliseeborn:
tested with various themes on V&A tenant, shows theme font size and color for all of them


src/ext/property-panel/settings.js
```diff
@@ -28,7 +35,10 @@ const stylingPanel = {
                       width: 9,
                       min: 5,
                       max: 300,
-                      placeholder: '',
+                      defaultValue(item, data, args) {
+                        const currentTheme = args.theme.current();
+                        return stripPx(currentTheme.object?.straightTable?.header?.fontSize || currentTheme.fontSize);
```
cbt1, line 40:
```suggestion
                        return parseInt(currentTheme.object?.straightTable?.header?.fontSize || currentTheme.fontSize, 10);
```

src/ext/property-panel/settings.js
```diff
@@ -66,7 +80,10 @@ const stylingPanel = {
                       width: 9,
                       min: 5,
                       max: 300,
-                      placeholder: '',
+                      defaultValue(item, data, args) {
+                        const currentTheme = args.theme.current();
+                        return stripPx(currentTheme.object?.straightTable?.content?.fontSize || currentTheme.fontSize);
```
cbt1, line 85:
```suggestion
                        return parseInt(currentTheme.object?.straightTable?.content?.fontSize || currentTheme.fontSize, 10);
```

cbt1:
Looks like a zip file got commited by accident

cbt1:
You are using the OR (||) operator but, it will give inconsistent results when the user sets a value in the theme that is truthy or falsy but not a valid font size/color. I think a safer approach is to use the nullish (??) coalesce operator instead. IMO let the user notice that a theme as invalid font size or color.

src/ext/property-panel/__test__/settings.spec.js
```diff
@@ -61,6 +61,36 @@ describe('settings', () => {
...
 
+    describe('headerFontSize/fontSize/headerFontColor/fontColor.defaultValue', () => {
+      const { headerFontSize, headerFontColor } = headerSection.items.headerFontItem.items.headerFontWrapper.items;
+      const { contentFontSize, contentFontColor } = contentSection.items.contentFontItem.items.contentFontWrapper.items;
+      let args;
+
+      beforeEach(() => {
+        args = {
+          theme: {
+            current: () => {
```
cbt1, line 72: Is the case where you get the font-size/color from the straight table theme tested some where else? I'm not seeing it here

cbt1 (approved):
LGTM, maybe it lacks a test or two

LiKang6688 (approved):

src/ext/property-panel/__test__/settings.spec.js
eliseeborn: no, I haven't tested that case, I'll add it now before merging 👍

cbt1 (approved):

PR #890, opened 2023-03-07
fix: white space before totals

cbt1:
Fixes an issue in the pagination table where the totals row would be incorrectly positioned (see screenshot below). This could happen when the table was re-sized. Ex. via object re-size in Qlik Sense or browser window re-sizing.

It did not happen every time. You had to find a scenario where the height of the header row would grow smaller as the table was re-sized.

**Before:**
Skärmavbild 2023-03-07 kl  13 15 44

**After:**
Skärmavbild 2023-03-07 kl  13 17 52

### DIFF
--- a/src/table/pagination-table/components/head/TableHeadWrapper.tsx
+++ b/src/table/pagination-table/components/head/TableHeadWrapper.tsx
@@ -19,8 +19,10 @@ function TableHeadWrapper({ areBasicFeaturesEnabled }: TableHeadWrapperProps) {
   const headRowRef = useRef<HTMLTableRowElement>(null);
 
   useEffect(() => {
-    headRowRef.current && setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
-  }, [headRowRef.current, styling.head.fontSize, headRowRef.current?.getBoundingClientRect().height]);
+    if (headRowRef.current) {
+      setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
+    }
+  }, [styling.head.fontSize, columnWidths, setHeadRowHeight]);
 
   return (
     <TableHead>


src/table/pagination-table/components/head/TableHeadWrapper.tsx
```diff
@@ -19,8 +19,10 @@ function TableHeadWrapper({ areBasicFeaturesEnabled }: TableHeadWrapperProps) {
   const headRowRef = useRef<HTMLTableRowElement>(null);
 
   useEffect(() => {
-    headRowRef.current && setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
-  }, [headRowRef.current, styling.head.fontSize, headRowRef.current?.getBoundingClientRect().height]);
+    if (headRowRef.current) {
+      setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
+    }
+  }, [styling.head.fontSize, columnWidths, setHeadRowHeight]);
```
cbt1, line 25: There a couple of scenarios when this use effect needs to run:
- Column is re-sized
- Container element is re-sized
- Layout changes
- Theme change

The `columnWidths` will take care of the 3 first scenarios and the `styling.head.fontSize` the last.

cbt1: Not entirely true as columnWidths have a bug where it does not run it's effect every time it needs to. Will make a new PR for that fix.

haxxmaxx (approved):

a-m-dev (approved):

src/table/pagination-table/components/head/TableHeadWrapper.tsx
cbt1: @LiKang6688  `setHeadRowHeight` is a React hook dispatch function. It does not change between renders so it will not triggers the `useEffect` on every render.

I added it, because otherwise eslint would complain about a missing dependency. You don't really need dispatch functions in the deps array but because it's coming from the `useContextSelector`eslint doesn't understand that. So had to add it to get rid of the warning.

LiKang6688: aha, I see. Thanks

PR #893, opened 2023-03-08
fix: put totals behind flag

haxxmaxx:
Realized that we were passing the flag to the keyboard handler for totals, that totals is not behind flag. This should fix that


haxxmaxx:
getting some new code climate complaints. They will be solved when we remove the flag

cbt1 (approved):

PR #895, opened 2023-03-08
fix: scroll to top when page changes

a-m-dev:
PR for scroll to top on page change


cbt1 (approved):

haxxmaxx:
I think we want to scroll up when the page is changed automatically as well. For example when you make a selection and you are moved back to the first page. let's discuss on daily

a-m-dev:
I tried multiple times but after selection confirmed, I've noticed that if you confirm the selection by keyboard (aka, pressing enter) it will still reset the focus in top. 

but if you confirm selection from nebula action button (check button) it will not, 
So will try the useEffect idea on page change

cbt1 (approved):

LiKang6688 (approved):
LGTM but no long one-liner scroll to top on page change 😺

haxxmaxx (approved):

PR #898, opened 2023-03-09
fix(a11y): press head menu button cannot open head cell menu

LiKang6688:


haxxmaxx:
it is a bit more complicated than this, I'll address it in my pr

PR #899, opened 2023-03-10
fix: totals in virtual scroll respect column alignment

eliseeborn:


cbt1 (approved):

src/table/virtualized-table/TotalsCell.tsx
```diff
@@ -39,7 +40,7 @@ const TotalsCell = ({ index, style, data }: TotalsCellProps) => {
         borderTopWidth: totals.atBottom ? '1px' : '0px',
         borderBottomWidth: totals.atTop ? '1px' : '0px',
         padding: '4px 12px',
-        justifyContent: index === 0 ? 'left' : 'right',
+        justifyContent: index === 0 ? 'left' : align,
```
haxxmaxx, line 43: Any reason why the first column doesn't use `align`. Looks like this in the old table
Screenshot 2023-03-10 at 14 42 31

eliseeborn: ah, good question, I just assumed that I should leave it like it was, so no special reason from me

haxxmaxx: ok then I think we should simply set it to `align`

PR #900, opened 2023-03-10
fix: context selector state

cbt1:
Fixes an issue where `useContextSelector` could return a state value before it triggered a state update for that value. This could cause some weird side effects.


src/table/context/useContextSelector.ts
```diff
@@ -1,35 +1,28 @@
...
-  const latestSelector = useRef(selector);
-  const latestSelectedState = useRef<any>();
-
   const currentValue = accessor();
 
   if (currentValue === undefined) {
     throw new Error('You must call useContextSelector inside a valid context.');
   }
 
-  const selectedState = useMemo(() => selector(currentValue), [currentValue, selector]);
```
cbt1, line 20: `selector` is always a new value as we do not wrap the `selector` in a `useCallback` when creating it. Instead it's always re-create. Ex: `useContextSelector(TableContext, (value) => value.baseProps)`.

So the `selectedState` is resolved/update on each render.

src/table/context/useContextSelector.ts
```diff
@@ -38,7 +31,7 @@ export function useContextSelector<T, TSelected>(context: Context<T>, selector:
     return () => {
       listeners.delete(listener);
     };
-  }, [listeners]);
+  }, [latestSelectedState, latestSelector, listeners]);
```
a-m-dev, line 34: I think since `latestSelectedState` and `latestSelector` becomes a ref state now (by using `useMutableProp()`), does passing it as dependency in this effect makes any difference? IMO it's extra in this deps list

src/table/context/createSelectorProvider.tsx
```diff
@@ -43,7 +43,7 @@ export function createSelectorProvider<T>(OriginalContext: Context<T>): Function
       return contextValueRef.current;
     }, [contextValueRef]);
 
-    const contextValue: SelectorContextType<T> = useMemo(() => [getContextValue, listeners.current], [contextValueRef]);
+    const contextValue: SelectorContextType<T> = useMemo(() => [getContextValue, listeners.current], [getContextValue]);
```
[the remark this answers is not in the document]
cbt1: Hmm I would think that if not, `contextValue` would always be a new array. Possibly triggering a new "render" because the provider value would change every time `createSelectorProvider` is called.

src/table/context/useContextSelector.ts
cbt1: Don't think so, `currentValue` could still be a new value which would recreate `selectedState`. So the consumer of this hook could still get a new `selectedState` before the actual state update happens.

src/table/context/useContextSelector.ts
cbt1: It does not but eslint doesn't understand that they are refs. So instead of disable the eslint rule here. I just added them as a deps.

It would work, as in no need to have them in deps array, if I declared the ref inside this hook. But since I do it via `useMutableProp` it does not work anymore

haxxmaxx:
Do we want to proceed with this?

cbt1:
> Do we want to proceed with this?

@haxxmaxx  It's one of those fixes that does not affect any behaviour the user is currently experiencing. But it might avoid bugs at a later point in time.

It is not a risk free change tho. But I think it will add more value, the more work that is spent on this project. So I would put it at "risk vs future quality".

Donya-qlik:
A minor request: Could you tag the bug ticket number to your description, thanks.

cbt1:
> A minor request: Could you tag the bug ticket number to your description, thanks.

No bug ticket for one. It's a more of a DX issue where it might bite us in the foot later if not fixed.

cbt1:
Closing this in favor of using some established state manager like Jotai instead

PR #901, opened 2023-03-10
fix: text alignment should be based on numeric or not numeric

LiKang6688:
- head cell 
	- column is measure: right
	- column is dimension
		- 	qDimensionType is N: right
		- otherwise: left

- totals cell
	- column is dimension: left
	- column is measure right

- inner body cell
	- cell's qNum is number: right
	- cell's qNum is not number: left

user-set text alignment should override the above.

### DIFF
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -83,7 +83,7 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
   const isLocked = isDim && (info as ExtendedNxDimensionInfo).qLocked;
   const { qDimensionType } = info as ExtendedNxDimensionInfo;
   const autoHeadCellTextAlign = qDimensionType === 'N' || qDimensionType === undefined ? 'right' : 'left';
-  const autoTotalsCellTextAlgin = isDim ? 'left' : 'right';
+  const autoTotalsCellTextAlign = isDim ? 'left' : 'right';
 
   let fieldIndex = 0;
   let fieldId = '';
@@ -110,7 +110,7 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
       sortDirection: qSortIndicator && qSortIndicator !== 'N' ? qSortIndicator : 'A',
       totalInfo: getTotalInfo(layout, colIdx, pageColIdx, numDims),
       autoHeadCellTextAlign,
-      autoTotalsCellTextAlgin,
+      autoTotalsCellTextAlign,
       textAlign: {
         auto: !textAlign || textAlign.auto,
         align: textAlign?.align,
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -48,7 +48,8 @@ export function getTotalPosition(layout: TableLayout, areBasicFeaturesEnabled = 
 }
 
 /**
- * Gets the totals text for a column
+ * Gets the totals label in the first column
+ * Gets the qText for each of the rest of columns
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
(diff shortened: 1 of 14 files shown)


src/handle-data.ts
```diff
@@ -79,7 +81,9 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
   } = info;
   const isHidden = qError?.qErrorCode === 7005;
   const isLocked = isDim && (info as ExtendedNxDimensionInfo).qLocked;
-  const autoAlign = isDim ? 'left' : 'right';
+  const { qDimensionType } = info as ExtendedNxDimensionInfo;
+  const autoHeadCellTextAlign = qDimensionType === 'N' || qDimensionType === undefined ? 'right' : 'left';
+  const autoTotalsCellTextAlgin = isDim ? 'left' : 'right';
```
haxxmaxx, line 86: typo

src/handle-data.ts
```diff
@@ -101,11 +105,16 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
...
       stylingIDs: qAttrExprInfo.map((expr) => expr.id),
       // making sure that qSortIndicator is either A or D
       sortDirection: qSortIndicator && qSortIndicator !== 'N' ? qSortIndicator : 'A',
       totalInfo: getTotalInfo(layout, colIdx, pageColIdx, numDims),
+      autoHeadCellTextAlign,
+      autoTotalsCellTextAlgin,
+      textAlign: {
+        auto: !textAlign || textAlign.auto,
+        align: textAlign?.align,
+      },
```
haxxmaxx, line 117: I don't really see why we need all of this information for later. Wouldn't it be possible to figure out what the final align is for all three cases (head, totals, body) and store those three straight on the column

src/table/virtualized-table/Cell.tsx
```diff
@@ -18,21 +19,19 @@ interface CellProps {
 
 const Cell = ({ columnIndex, rowIndex, style, data }: CellProps) => {
   const { rowsInPage, columns, bodyStyle, isHoverEnabled, maxLineCount } = data;
-  const cell = rowsInPage[rowIndex]?.[`col-${columnIndex}`];
+  const cell = rowsInPage[rowIndex]?.[`col-${columnIndex}`] as CellType;
```
haxxmaxx, line 22: not sure you can cast this, @cbt1 can it be undefined at this stage?

src/table/pagination-table/components/body/TableBodyWrapper.tsx
```diff
@@ -61,8 +61,10 @@ function TableBodyWrapper({
           className="sn-table-row"
         >
           {columns.map((column, columnIndex) => {
-            const { id, align } = column;
+            const { id, textAlign } = column;
             const cell = row[id] as Cell;
+            const autoAlign = () => (cell.qNum && !Number.isNaN(+cell.qNum) ? 'right' : 'left');
+            const align = textAlign.auto ? autoAlign() : textAlign.align;
```
haxxmaxx, line 67: This we also could do on a handle-data level, when we create `rows`. Unfortunately we have do do it in a different place for the virtual scroll. But we could reuse the logic.

LiKang6688: Yes, that is how the old table behaves
Screenshot 2023-03-09 at 09 48 34

src/handle-data.ts
LiKang6688: we need data cell level. I do not see the easy solution to store them on the column

LiKang6688:
gonna move logic from component level to fetch data place

LiKang6688:
gonna move logic from component level to fetch data place

PR #902, opened 2023-03-10
fix: text alignment should be based on numeric or not numeric

LiKang6688:
- head cell 
	- column is measure: right
	- column is dimension
		- 	qDimensionType is N: right
		- otherwise: left

- totals cell
	- column is dimension: left
	- column is measure right

- inner body cell
	- cell's qNum is number: right
	- cell's qNum is not number: left

user-set text alignment should override the above.

Screenshot 2023-03-10 at 20 52 19


src/types.ts
```diff
@@ -126,7 +130,9 @@ export interface Column {
   colIdx: number;
   pageColIdx: number;
   label: string;
-  align: 'left' | 'center' | 'right';
+  headCellTextAlign: Align;
+  totalsCellTextAlign: Align;
+  textAlign: string | Align;
```
haxxmaxx, line 135: could type this more specifically to `Align | 'auto'`

haxxmaxx (approved):
Looks good, just some suggestions

src/table/utils/styling-utils.ts
```diff
@@ -311,3 +311,6 @@ export function getSelectionStyle(styling: CellStyle, cellSelectionState: Select
     ...selectionStyling,
   };
 }
+
+export const autoAlign = (cell: EngineAPI.INxCell) =>
```
a-m-dev, line 315: this one returns type of `Align`, I think by returning Align type from this probably you wont need to cast in line 180 of `handle-data.ts` file

but also it might return a default alignment as well because of boolean check

src/handle-data.ts
```diff
@@ -79,7 +82,11 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
   } = info;
   const isHidden = qError?.qErrorCode === 7005;
   const isLocked = isDim && (info as ExtendedNxDimensionInfo).qLocked;
-  const autoAlign = isDim ? 'left' : 'right';
+  const { qDimensionType } = info as ExtendedNxDimensionInfo;
+  const autoHeadCellTextAlign = qDimensionType === 'N' || qDimensionType === undefined ? 'right' : 'left';
```
a-m-dev, line 86: I remember that we had some table use case where column was showing some sort of image not numbers of strings. 

My questions is what we will get `qDimensionType` in those cases and will those be automatically right aligned?

src/table/virtualized-table/hooks/use-data/utils/index.ts
```diff
@@ -15,11 +16,16 @@ const createRow = (
   const pageRowIdx = pageRowStartIdx + cellRowIdx;
   const row: Row = rows[pageRowIdx] ?? { key: `row-${pageRowIdx}` };
 
+  const getBodyCellAlign = (textAlign: 'auto' | Align, cell: EngineAPI.INxCell) =>
```
a-m-dev, line 19: same align type here as well

a-m-dev (approved):

src/table/virtualized-table/hooks/use-data/utils/index.ts
cbt1, line 19: Why is this function created inside the `createRow` function? It looks pure to me?

LiKang6688:
```
const getBodyCellAlign = (textAlign: 'auto' | Align, cell: EngineAPI.INxCell) =>
  textAlign === 'auto' ? autoAlign(cell) : textAlign;

const createRow = (
```

is it okay? or should I move it to another file?

cbt1: It's OK

cbt1 (approved):

PR #906, opened 2023-03-13
chore: update qext

haxxmaxx:


assets/sn-table.qext
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "sn-table",
-  "version": "0.1.0",
-  "description": "",
+  "name": "Straight table",
+  "version": "1.24.0",
```
haxxmaxx, line 3: I put the latest version here. This is not update automatically. I checked a couple of other extensions, the version here is not up to date

cbt1: File have been removed. It no longer served any purpose

cbt1: The version is generated from the package.json file (https://github.com/qlik-oss/nebula.js/blob/0dd499ce8777ee421deb616b6ed45e0ad1ebb64e/commands/sense/lib/build.js#L399)

niekvanstaveren (approved):

PR #911, opened 2023-03-14
feat(api)!: add columnWidth and usePagination to API

haxxmaxx:


src/qae/object-properties.js
```diff
@@ -76,6 +76,11 @@ const properties = {
      */
     label: 'Totals',
   },
+  /**
+   * use pagination or continuous scroll. If undefined, defaults to using pagination
```
cbt1, line 80:
```suggestion
   * Use pagination or continuous scroll. If undefined, defaults to using pagination
```

src/qae/object-properties.js
```diff
@@ -76,6 +76,11 @@ const properties = {
...
+   * @type {boolean}
```
cbt1, line 81: As this is documented. The property is not optional

haxxmaxx: Sorry not following. Do you mean that it shouldn't be optional but it is now?

LiKang6688:
After we merge this PR, shall we release the sn-table as 2.0.0?

src/ext/property-panel/settings.js
```diff
@@ -255,11 +255,11 @@ const getChartExploration = (env) =>
 
 const getUsePagination = ({ flags }) =>
   flags.isEnabled('PS_18291_SN_TABLE_BASIC_FEATURES') && {
-    ref: 'presentation.usePagination',
+    ref: 'usePagination',
```
LiKang6688, line 258: I am curious. Why do we decide to remove presentation?

cbt1: The other presentation properties are on the root, so it made no sense to only have this prop there

LiKang6688: okay, it makes sense to me. Thx

src/qae/object-properties.js
haxxmaxx: Hmm you are right. I was thinking for a newly created chart this could be undefined, but then it willl be "filled in" by this definition. Then the second sentence becomes overkill to me

LiKang6688 (approved):

PR #913, opened 2023-03-14
fix: to have correct focus indication for head cell menu items

LiKang6688:


src/table/constants.ts
```diff
@@ -62,3 +62,5 @@ export const INITIAL_DATA_FETCH_HEIGHT = 100;
 
 export const FIRST_HEADER_CELL_COORD: [number, number] = [0, 0];
 export const FIRST_BODY_CELL_COORD: [number, number] = [1, 0];
+
+export const FOCUS_BOX_SHADOW = '0 0 0 2px #177fe6 inset, 0 0 0 3px #fff inset';
```
haxxmaxx, line 66: this boxshadow is specifically defined for the table cells, so that you get a small white border to create contrast. Especially against the selection green. We don't really need that for the menu options I think, so just the blue box shadow should be enough

LiKang6688: Okay

LiKang6688: `boxShadow: 'rgb(23, 127, 230) 0px 0px 0px 2px'`, is what mui uses. Is it okay?

a-m-dev (approved):

PR #915, opened 2023-03-14
fix: scroll to infinity in firefox

a-m-dev:
some border mis calc were fixed in this PR from column adjuster => which was the root cause of issue!


haxxmaxx (approved):

LiKang6688 (approved):
Magic!

PR #916, opened 2023-03-15
fix: keyboard navigation tab handling

haxxmaxx:
Noticed a bunch of errors with the keyboard navigation. Seems like they mostly stemmed from not handling tabs correctly. Even though the focus moved properly, the focusCellCoord and tab stops where not updated properly. This is what has changed:
- tab handling. In most cases we still rely on moving the focus using the tab order (the exceptions are moving from and to the header), but we now make sure that the `focusedCellCoord` is updated correctly. That is why the tab helpers had to be introduced. This means that we don't do `preventDefaultBehavior` for all cases in the handlers, meaning the call has moved inside the different key cases
- simplified the `shouldBubbleEarly` logic. What this function does is early return when some key press is handled in both the inner component (e.g. body) and the wrapper. So we need to early return for arrows and esc. 
- some updates on the jsdocs/comments for the logic
- clarifying some function names
- unit tests

### DIFF
--- a/src/table/pagination-table/components/body/TableTotals.tsx
+++ b/src/table/pagination-table/components/body/TableTotals.tsx
@@ -25,7 +25,6 @@ function TableTotals() {
       {columns.map((column, columnIndex) => {
         const cellCoord: [number, number] = [atTop ? 1 : rows.length + 1, columnIndex];
         const tabIndex = atTop && columnIndex === 0 && !keyboard.enabled ? 0 : -1;
-        console.log(tabIndex);
 
         return (
           <StyledTotalsCell
--- a/src/table/pagination-table/components/body/TableTotals.tsx
+++ b/src/table/pagination-table/components/body/TableTotals.tsx
@@ -3,7 +3,7 @@ import TableRow from '@mui/material/TableRow';
 
 import { useContextSelector, TableContext } from '../../../context';
 import { handleTotalKeyDown } from '../../../utils/handle-keyboard';
-import { areTabStopsEnabled, removeTabAndFocusCell } from '../../../utils/accessibility-utils';
+import { removeTabAndFocusCell } from '../../../utils/accessibility-utils';
 import { StyledTotalsCell } from './styles';
 import CellText from '../../../components/CellText';
 
@@ -24,7 +24,9 @@ function TableTotals() {
     <TableRow className="sn-table-row">
       {columns.map((column, columnIndex) => {
         const cellCoord: [number, number] = [atTop ? 1 : rows.length + 1, columnIndex];
-        const tabIndex = atTop && columnIndex === 0 && areTabStopsEnabled(keyboard) ? 0 : -1;
+        const tabIndex = atTop && columnIndex === 0 && !keyboard.enabled ? 0 : -1;
+        console.log(tabIndex);
+
         return (
           <StyledTotalsCell
             totalsStyle={styling.totals}
--- a/src/table/utils/__tests__/accessiblity-utils.spec.ts
+++ b/src/table/utils/__tests__/accessiblity-utils.spec.ts
@@ -1,8 +1,9 @@
 import { stardust } from '@nebula.js/stardust';
 import React from 'react';
 import { Announce, TotalsPosition } from '../../../types';
-import { FocusTypes } from '../../constants';
+import { FIRST_BODY_CELL_COORD, FocusTypes } from '../../constants';
 import * as accessibilityUtils from '../accessibility-utils';
+import * as getElementUtils from '../get-element-utils';
 
 describe('accessibility-utils', () => {
   let cell: HTMLTableCellElement | undefined;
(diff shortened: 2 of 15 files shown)


cbt1:
Not sure if related to this PR but I did some quick testing and found a potential issue:
1. Create a table with totals on the top
2. Do some selections and confirm
3. Now focus table and hit enter

The first cell in the totals row is focused. I was expecting either first header or body cell.

src/table/pagination-table/components/body/TableTotals.tsx
```diff
@@ -24,7 +24,8 @@ function TableTotals() {
     <TableRow className="sn-table-row">
       {columns.map((column, columnIndex) => {
         const cellCoord: [number, number] = [atTop ? 1 : rows.length + 1, columnIndex];
-        const tabIndex = atTop && columnIndex === 0 && areTabStopsEnabled(keyboard) ? 0 : -1;
+        const tabIndex = atTop && columnIndex === 0 && !keyboard.enabled ? 0 : -1;
```
a-m-dev, line 27: I tried to set tab index to 0 
```
const tabIndex = 0; // atTop && columnIndex === 0 && !keyboard.enabled ? 0 : -1;
```
and I was expecting to focus land on totals cell by default but this happened: 

https://user-images.githubusercontent.com/29652890/226284051-c68aa641-f9cb-4997-b065-709516584d2c.mov

src/table/utils/__tests__/accessiblity-utils.spec.ts
```diff
@@ -366,4 +367,37 @@ describe('accessibility-utils', () => {
       expect(keyboard.focus).toHaveBeenCalledTimes(1);
     });
   });
+
+  describe('focusBodyFromHead', () => {
+    let otherCell: HTMLTableCellElement;
+
+    beforeEach(() => {});
```
a-m-dev, line 374: oops!

src/table/utils/accessibility-utils.ts
```diff
@@ -85,13 +85,18 @@ export const moveFocus = (
...
+ */
 export const focusBodyFromHead = (
   rootElement: HTMLElement,
   setFocusedCellCoord: React.Dispatch<React.SetStateAction<[number, number]>>
 ) => {
-  const cell = findCellWithTabStop(rootElement);
-  const newCellCoord = getCellCoordFromCell(rootElement, cell);
-  cell.focus();
+  let cell = findCellWithTabStop(rootElement);
+  const newCellCoord = cell ? getCellCoordFromCell(rootElement, cell) : FIRST_BODY_CELL_COORD;
```
a-m-dev, line 97: I would suggest to change this function's name as well, 
`getCellCoords` should be fine, I think since we passing a cell into that function that should be obvious that we need coords of that cell and it requires a cell element

a-m-dev: also, I've checked that function, Not sure if it will work correctly in VT. have you tried that case? 

Seems like you cannot move tab stops more than last cell in the head 👇

https://user-images.githubusercontent.com/29652890/226286785-fecbf8cb-ce91-47ce-a5fc-f941edca8e55.mov

haxxmaxx:
> Not sure if related to this PR but I did some quick testing and found a potential issue:
> 
>     1. Create a table with totals on the top
> 
>     2. Do some selections and confirm
> 
>     3. Now focus table and hit enter
> 
> 
> The first cell in the totals row is focused. I was expecting either first header or body cell.

yupp, that is known. I think it is fine for now. It is pretty hard to keep track of what made the table update and where to reset focus. I've had some problems setting it to the header in all cases. I think we can wait for some feedback on this

src/table/utils/accessibility-utils.ts
```diff
@@ -85,13 +85,18 @@ export const moveFocus = (
   return nextCell;
 };
 
+/**
+ * Finds the cell with tab stops and focuses that cell.
+ * If no cells has focus, it focuses the first body cell instead
```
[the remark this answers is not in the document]
haxxmaxx: yeah the totals are in the body, so it implies that it is the first totals cell if the totals are on top

cbt1 (approved):
LGTM, not to say that @a-m-dev made some comments, that is worth looking into.

src/table/pagination-table/components/body/TableTotals.tsx
a-m-dev: I just could not make up to that case (atTop && colIdx and ...) => so i decided to change it to 0 to see what would happen. Just being curious 🤷‍♂️

haxxmaxx: ok I see :)

a-m-dev: PR looks good, we need to look at this issue in another one

a-m-dev (approved):

PR #917, opened 2023-03-15
fix: hide focus indicator when not using keyboard

eliseeborn:


haxxmaxx (commented):
Just to make sure, have you tested in virtual scroll? We don't have any focus handling there, just so we don't get any focus indicator when clicking

eliseeborn:
> Just to make sure, have you tested in virtual scroll? We don't have any focus handling there, just so we don't get any focus indicator when clicking

Virtual table shows no focus indicators

haxxmaxx (approved):

PR #919, opened 2023-03-15
fix: head cell content left | center | right alignment

a-m-dev:
Here is the preview of this PR changes which is not breaking the a11y. 

Also below you can see a left, center and right aligned Dim column. Last one is a measure and all tab stops working correctly

The space between lock icon and head cell text/button also fixed

https://user-images.githubusercontent.com/29652890/225703354-a0282030-0466-49eb-a853-7c7d589391d3.mov


src/table/components/head/HeadCellContent.tsx
```diff
@@ -33,9 +33,18 @@ function HeadCellContent({ children, column, isActive, areBasicFeaturesEnabled }
       areBasicFeaturesEnabled,
     });
 
+  const headCellContentClassName = useMemo(() => {
+    let className = '';
+    if (!column.isDim || column.headCellTextAlign === 'right') className += 'reverse ';
```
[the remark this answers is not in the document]
haxxmaxx: So you could probably use just use `headCellTextAlign` as the class name, then have special cases for `right` and `center`

src/table/components/head/styles.ts
```diff
@@ -49,23 +51,46 @@ export const VisuallyHidden = styled('span')({
...
+  justifyContent: 'center',
+  alignItems: 'center',
+});
+
 export const LockWrapper = styled(Box)({
   display: 'flex',
-  justifyContent: 'flex-end',
-  alignItems: 'flex-start',
+  justifyContent: 'center',
+  alignItems: 'center',
```
haxxmaxx, line 93: So these made sure that the icon was aligned correctly, make sure that alignment is preserved

haxxmaxx (commented):
so the menu button is in the wrong place if you left align a measure, same thing for center

Screenshot 2023-03-16 at 10 41 27

src/table/components/head/styles.ts
```diff
@@ -49,23 +51,46 @@ export const VisuallyHidden = styled('span')({
...
+      gridColumn: 2,
+    },
+
+    '& > div:last-child': {
+      gridColumn: 1,
+    },
+  },
+  '&.centered': {
+    gridTemplateColumns: `${DEFAULT_COLUMN_ICON_WIDTH}px 2fr ${DEFAULT_COLUMN_ICON_WIDTH}px`,
+  },
```
haxxmaxx, line 81: I think you can get away from classed here. All you would have to do is spread based on `align === 'center'` or something

src/table/constants.ts
```diff
@@ -54,6 +54,7 @@ export const MAX_COLUMN_WIDTH = 7680;
 export const MAX_COLUMN_PERCENTAGE_WIDTH = 100;
 export const DEFAULT_COLUMN_PIXEL_WIDTH = 200;
 export const DEFAULT_COLUMN_PERCENTAGE_WIDTH = 20;
+export const DEFAULT_COLUMN_ICON_WIDTH = 25;
```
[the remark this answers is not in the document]
a-m-dev: the default size of icon was 24 plus some decimal => so i rounded it up to 25 px

cbt1:
I tested this a bit together with the `fit to content` option on the fields just to see if it broke anything in the calculation.

It did not! So from that perspective it LGTM

haxxmaxx:
alignment looks better but doesn't deal with small column width well, the ... button is overlapping the label
Screenshot 2023-03-17 at 08 27 54

src/table/components/head/styles.ts
```diff
@@ -49,27 +54,58 @@ export const VisuallyHidden = styled('span')({
...
+  shouldForwardProp: (prop: string) => prop !== 'isLocked',
+})(({ theme, isLocked }) => ({
   width: '100%',
-  display: 'flex',
-  justifyContent: 'space-between',
+  display: 'grid',
   flexDirection: 'inherit',
-  alignItems: 'flex-end',
+  gridAutoFlow: 'dense',
   gap: theme.spacing(0.5),
```
haxxmaxx, line 64: do you still need this? I think in general since you are introducing a lot of new css, make sure to clean out stuff that is not needed

a-m-dev: Yes, that is because of `grid-column` property makes items in separate lines and making them dense -> puts them back into one line

a-m-dev:
Fixed the less column width problem

https://user-images.githubusercontent.com/29652890/225868191-6a551a51-0c02-4763-9383-9b5bba5a5635.mov

haxxmaxx (approved):

PR #921, opened 2023-03-17
feat: use new theme entry

haxxmaxx:
Switching the keyword for the theme from `straightTable` to `straightTableV2`


nebula.config.js
```diff
@@ -17,10 +17,21 @@ module.exports = {
     },
     keyboardNavigation: false,
     themes: [
+      {
```
haxxmaxx, line 20: I added this theme to emulate how it will look in the client

PR #924, opened 2023-03-20
fix: show right border when table is smaller than object

LiKang6688:
Screenshot 2023-03-20 at 09 40 52

virtualized-table:
Screenshot 2023-03-20 at 16 08 26

### DIFF
--- a/src/table/components/head/ColumnAdjuster.tsx
+++ b/src/table/components/head/ColumnAdjuster.tsx
@@ -13,7 +13,7 @@ import { focusHeadMenuButton } from '../../utils/accessibility-utils';
  * When you start dragging, mouse move and mouse up listeners are added.
  * While dragging the current column is updated, and on mouse up all other columns are updated.
  */
-const ColumnAdjuster = ({ column, isLastColumn, onColumnResize, paginationTableRef }: AdjusterProps) => {
+const ColumnAdjuster = ({ column, isLastColumn, onColumnResize }: AdjusterProps) => {
   const { pageColIdx } = column;
   const { rootElement, applyColumnWidths, constraints } = useContextSelector(TableContext, (value) => value.baseProps);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
@@ -21,16 +21,9 @@ const ColumnAdjuster = ({ column, isLastColumn, onColumnResize, paginationTableR
   const tempWidths = useRef({ adjusterHitArea: {}, columnWidth: 0, initX: 0, initWidth: 0 });
   const borderHeight = rootElement.getBoundingClientRect().height - PAGINATION_HEIGHT - BORDER_WIDTH;
 
-  const tableWidth = useContextSelector(TableContext, (value) => value.tableWidth);
-  const setBorderRight = useContextSelector(TableContext, (value) => value.setBorderRight);
-
   if (!applyColumnWidths || constraints.active) return null;
 
   const updateWidth = (adjustedWidth: number) => {
-    const columnsWidth = paginationTableRef?.current?.getBoundingClientRect?.()?.width;
-    if (columnsWidth && columnsWidth < tableWidth) setBorderRight(true);
-    else setBorderRight(false);
-
     onColumnResize?.();
     tempWidths.current.columnWidth = adjustedWidth;
     const newColumnWidths = [...columnWidths];
--- a/src/table/components/head/ColumnAdjuster.tsx
+++ b/src/table/components/head/ColumnAdjuster.tsx
@@ -13,7 +13,7 @@ import { focusHeadMenuButton } from '../../utils/accessibility-utils';
  * When you start dragging, mouse move and mouse up listeners are added.
  * While dragging the current column is updated, and on mouse up all other columns are updated.
  */
-const ColumnAdjuster = ({ column, isLastColumn, onColumnResize }: AdjusterProps) => {
+const ColumnAdjuster = ({ column, isLastColumn, onColumnResize, paginationTableRef }: AdjusterProps) => {
   const { pageColIdx } = column;
   const { rootElement, applyColumnWidths, constraints } = useContextSelector(TableContext, (value) => value.baseProps);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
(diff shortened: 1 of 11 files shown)


src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -13,17 +13,24 @@ import { focusHeadMenuButton } from '../../utils/accessibility-utils';
...
 
+  const tableWidth = useContextSelector(TableContext, (value) => value.tableWidth);
+  const setBorderRight = useContextSelector(TableContext, (value) => value.setBorderRight);
+
   if (!applyColumnWidths || constraints.active) return null;
 
   const updateWidth = (adjustedWidth: number) => {
+    const columnsWidth = paginationTableRef?.current?.getBoundingClientRect?.()?.width;
+    if (columnsWidth && columnsWidth < tableWidth) setBorderRight(true);
+    else setBorderRight(false);
```
haxxmaxx, line 32: I'm not sure we have to set this here. I think it is fine to set the borde once you are done changing the column width. That is when we update the other columns as well

src/table/pagination-table/components/styles.ts
```diff
@@ -27,12 +27,15 @@ export const StyledTableContainer = styled(TableContainer, {
 }));
 
 export const StyledTable = styled(Table, {
-  shouldForwardProp: (prop: string) => prop !== 'customWidth',
-})(({ customWidth }) =>
+  shouldForwardProp: (prop: string) => prop !== 'customWidth' && prop !== 'isBorderRight',
+})(({ customWidth, isBorderRight }) =>
```
haxxmaxx, line 31: not sure we want to put the border on the table. It will look slightly different than putting it on the cell. Also, this does not respect if we need the dark mode borders

src/table/components/head/ColumnAdjuster.tsx
haxxmaxx: This also does not take into account if the table resizes and we should show/hide the border

src/table/pagination-table/components/styles.ts
LiKang6688: Right, dark mode, Thx.

the border on the table and the cell is no different as far as I can see. And on the table is much simpler.
Screenshot 2023-03-20 at 11 19 43
Screenshot 2023-03-20 at 11 00 37

src/table/components/head/ColumnAdjuster.tsx
LiKang6688: set the borde once you are done changing the column width. Yes, that's better.

LiKang6688: right, table resizes. Maybe `tableWidth={props.rect.width} `can help

src/table/pagination-table/components/styles.ts
```diff
@@ -27,12 +27,15 @@ export const StyledTableContainer = styled(TableContainer, {
...
-})(({ customWidth }) =>
+  shouldForwardProp: (prop: string) => prop !== 'customWidth' && prop !== 'isBorderRight' && prop !== 'theme',
+})(({ customWidth, isBorderRight, theme }) =>
   customWidth
     ? {
         tableLayout: 'fixed',
         width: 'min-content',
+        borderWidth: isBorderRight && '0px 1px 0px 0px',
+        borderStyle: isBorderRight && 'solid',
+        borderColor: theme.background.isDark ? COLORING.DARK_MODE_BORDER : COLORING.BORDER_LIGHT,
```
haxxmaxx, line 38: we already do this calculation, the information is on the styling object

src/table/pagination-table/components/styles.ts
haxxmaxx: there is a small difference, but barely even noticable. If you zoom in on the corners, you see that you get a diagonal intersection kind of. I think we can keep the right border on table level

src/table/components/head/ColumnAdjuster.tsx
haxxmaxx: or you just calculate it on the table or context level. There is no point really setting it in the ColumnAdjuster. I'm pretty sure the value needs to change when columnWidths or rect.width changes.

src/table/pagination-table/components/styles.ts
LiKang6688: it `styling.body.borderRightColor`?

src/table/pagination-table/components/styles.ts
```diff
@@ -27,12 +27,15 @@ export const StyledTableContainer = styled(TableContainer, {
...
-})(({ customWidth }) =>
+  shouldForwardProp: (prop: string) => prop !== 'customWidth' && prop !== 'isBorderRight' && prop !== 'styling',
+})(({ customWidth, isBorderRight, styling }) =>
   customWidth
     ? {
         tableLayout: 'fixed',
         width: 'min-content',
+        borderWidth: isBorderRight && '0px 1px 0px 0px',
+        borderStyle: isBorderRight && 'solid',
+        borderColor: isBorderRight && styling.body.borderRightColor,
```
haxxmaxx, line 38: you could do this in one line with right-border
```suggestion
        borderRight: isBorderRight && `1px solid ${styling.body.borderRightColor}`,
```

src/table/utils/styling-utils.ts
```diff
@@ -18,6 +18,9 @@ export const LINE_HEIGHT = 4 / 3;
 export const CELL_PADDING_HEIGHT = 8;
 export const CELL_BORDER_HEIGHT = 1;
 
+export const getTableWidth = (columnWidths: number[]) =>
+  columnWidths && columnWidths.length > 0 && columnWidths.reduce((a, b) => a + b);
```
haxxmaxx, line 22:
```suggestion
export const showRightBorder = (columnWidths: number[], tableWidth: number) =>
  columnWidths && columnWidths.length > 0 ? columnWidths.reduce((a, b) => a + b) < tableWidth : false;
```

src/table/context/TableContext.tsx
```diff
@@ -51,6 +52,10 @@ export const TableContextProvider = ({
     tableWidth,
     styling
   );
+  const isBorderRight = useMemo(
```
haxxmaxx, line 55: Not sure the name makes sense, showRightBorder or hasRightBorder is more descriptive

src/table/virtualized-table/Cell.tsx
```diff
@@ -44,7 +47,7 @@ const Cell = ({ columnIndex, rowIndex, style, data }: CellProps) => {
           alignItems: 'center',
           borderWidth: '0px',
           borderBottomWidth: cell.isLastRow ? '0px' : `${BORDER_WIDTH}px`,
-          borderRightWidth: cell.isLastColumn ? '0px' : `${BORDER_WIDTH}px`,
+          borderRightWidth: cell.isLastColumn ? borderRightWidthInLastColum : `${BORDER_WIDTH}px`,
```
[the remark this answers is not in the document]
haxxmaxx: same thing for header and totals

haxxmaxx (commented):
solution looks good, just some refactoring comments

src/table/virtualized-table/Cell.tsx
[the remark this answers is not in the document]
LiKang6688: It does not make sense to me. Did you write it correctly?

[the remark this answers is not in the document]
LiKang6688: I think I get what you mean

src/table/context/TableContext.tsx
```diff
@@ -51,6 +52,10 @@ export const TableContextProvider = ({
     tableWidth,
     styling
   );
+  const showRightBorder = useMemo(
```
cbt1, line 55: This a case that is not considered here and that is when the use have a vertical scrollbar that take up space.

So you probably want to move this into `useColumnWidth` so that you can include the `yScrollbarWidth`. Otherwise it will show a right border for the last column even if the columns occupies all width available.

haxxmaxx (approved):
LGTM as long as it works correct with the scrollbar

PR #927, opened 2023-03-21
chore: remove `@mui/system` dependency

a-m-dev:
just a small change from @mui/system/styles -> @mui/material/styles in our imports + removing it from pkg.json file


yarn.lock
```diff
@@ -2358,7 +2358,7 @@
     "@babel/runtime" "^7.20.1"
     prop-types "^15.8.1"
 
-"@mui/system@5.11.12", "@mui/system@^5.11.12":
+"@mui/system@^5.11.12":
```
haxxmaxx, line 2361: I think if we redo yarn.lock from scratch we shouldn't see mui/system in here at all

haxxmaxx (commented):
I thought we would get complaints that the props in the styled components aren't typed. Do you know why not?

a-m-dev:
Both are same functions, and according to the docs the only difference is in using the default theme if there is none available through context. So I think should not be any problems in types.

yarn.lock
a-m-dev: I think `@mui/material` depends on it 👇

Screenshot 2023-03-21 at 10 40 11

haxxmaxx:
> Both are same functions, and according to the docs the only difference is in using the default theme if there is none available through context. So I think should not be any problems in types.

hmm ok then this might not affect bundle size. I still don't understand why typescript doesn't complain about the untyped callbacks in the styled functions...

haxxmaxx (approved):

PR #929, opened 2023-03-21
fix: add tooltips to table cells

eliseeborn:
adding tooltips using the html attribute title - same as for the old table.
I looked at adding it in CellText, but then the tooltip is only visible while hovering the text, not the rest of the cell.


src/table/pagination-table/components/body/withStyling.tsx
```diff
@@ -26,6 +26,7 @@ export default function withStyling(CellComponent: typeof StyledBodyCell) {
         align={align}
         tabIndex={tabIndex}
         className={`sn-table-cell ${selectedCellClass || ''}`}
+        title={(children as ReactElement).props.children}
```
cbt1, line 29: I don't really understand this one. Care to explain?

eliseeborn: children here is a StyledBodyCell,which has the text we want as a tooltip in its props, but children is of type ReactNode by default, which doesn't always have props.
So I'm type casting it to a ReactElement

cbt1: Ah now I see. So `children.props.children` turns out to the text node.

I do think this could break rather easily. A change in the depth of the text node and this will break.

I would consider a different solution. For example just passing the title as a prop the cell renderer?

eliseeborn: passing the title as a prop might work. If we add it on CellText, it will be only on the text itself and not the cell, and yeah, it might not work at all in the header.
I'll look into it some more :)

eliseeborn:
I have a question around the comment `//passive: turn off tooltips` - the old table has tooltips enabled in editmode, do we want to change that behavior?

cbt1 (approved):
LGTM

PR #931, opened 2023-03-21
refactor: move rect to table context

cbt1:
Moves the rect to the TableContext. It's required for fixing another issue, otherwise the VT table will run in to issue with the `columnWidths` and `rect` being out-of-sync.

### DIFF
--- a/src/table/pagination-table/components/TableWrapper.tsx
+++ b/src/table/pagination-table/components/TableWrapper.tsx
@@ -23,7 +23,7 @@ function TableWrapper(props: TableWrapperProps) {
 
   const { totalColumnCount, totalRowCount, totalPages, paginationNeeded, rows, columns, totalsPosition } =
     useContextSelector(TableContext, (value) => value.tableData);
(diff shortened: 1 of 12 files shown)


src/table/pagination-table/components/TableWrapper.tsx
```diff
@@ -132,7 +132,12 @@ function TableWrapper(props: TableWrapperProps) {
       </StyledTableContainer>
       {!constraints.active && (
         <FooterWrapper footerContainer={footerContainer} paginationNeeded={paginationNeeded}>
-          <PaginationContent {...props} handleChangePage={handleChangePage} isSelectionMode={isSelectionMode} />
+          <PaginationContent
+            {...props}
+            rect={rect}
```
haxxmaxx, line 137: why not use the context selector inside PaginationContent?

cbt1: I like that. Will fix!

haxxmaxx (approved):

PR #934, opened 2023-03-22
fix: slow loading with many pages

cbt1:
Fixes an issue where the pagination table would be really slow to load (5-6 seconds) when you had millions rows of data. This was because the number of page needed for a large data set would become huge and rendering a `<select>` element with 100k `<option>` was not good for performance.

This proposed solution set a max value on the number of pages can be rendered in the "Select page" dropdown. When the user selects a page that is at the end of the dropdown and there are more pages available, the dropdown will contain a page range that is:
- `startPageIndex = currentPage - 500`
- `endPageIndex = currentPage + 500`


eliseeborn (approved):
LGTM!

src/table/components/footer/PageOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const ELLIPSIS = '…';
+
+const PageOptions = ({ totalPages, page }: PageOptionsProps) => {
+  const pageOptionsLength = Math.min(totalPages, MAX_PAGE_OPTIONS_LENGTH);
+  const tooManyPages = pageOptionsLength < totalPages;
+  const endPageIndex =
+    tooManyPages && page > MAX_PAGE_OPTIONS_LENGTH / 2
+      ? Math.min(MAX_PAGE_OPTIONS_LENGTH / 2 + page, totalPages)
+      : Math.min(MAX_PAGE_OPTIONS_LENGTH, totalPages);
```
haxxmaxx, line 18:
```suggestion
      : pageOptionsLength;
```

src/table/components/footer/PageOptions.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const MAX_PAGE_OPTIONS_LENGTH = 1000;
+
+export const ELLIPSIS = '…';
+
+const PageOptions = ({ totalPages, page }: PageOptionsProps) => {
+  const pageOptionsLength = Math.min(totalPages, MAX_PAGE_OPTIONS_LENGTH);
+  const tooManyPages = pageOptionsLength < totalPages;
+  const endPageIndex =
+    tooManyPages && page > MAX_PAGE_OPTIONS_LENGTH / 2
```
haxxmaxx, line 16: this is checked twice, could maybe be called something, might improve readability

haxxmaxx (approved):

PR #935, opened 2023-03-22
fix: cell height on container or theme change

cbt1:


haxxmaxx (approved):
code lgtm

src/table/virtualized-table/Header.tsx
```diff
@@ -7,10 +7,10 @@ import useResetHeader from './hooks/use-reset-header';
 
 const Header = (props: HeaderProps) => {
   const { rect, forwardRef, columns, pageInfo, headerStyle, rowHeight, columResizeHandler } = props;
-  const { layout } = useContextSelector(TableContext, (value) => value.baseProps);
+  const { layout, theme } = useContextSelector(TableContext, (value) => value.baseProps);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
 
-  useResetHeader(forwardRef, layout, pageInfo, columnWidths);
+  useResetHeader(forwardRef, layout, pageInfo, columnWidths, theme);
```
a-m-dev, line 13: why not directly passing `theme.name()` from here?

src/table/virtualized-table/Totals.tsx
```diff
@@ -7,10 +7,10 @@ import useResetHeader from './hooks/use-reset-header';
 
 const Totals = (props: TotalsProps) => {
   const { rect, forwardRef, pageInfo, totals, rowHeight, columns } = props;
-  const { layout, styling } = useContextSelector(TableContext, (value) => value.baseProps);
+  const { layout, styling, theme } = useContextSelector(TableContext, (value) => value.baseProps);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
 
-  useResetHeader(forwardRef, layout, pageInfo, columnWidths);
+  useResetHeader(forwardRef, layout, pageInfo, columnWidths, theme);
```
a-m-dev, line 13: here too

src/table/virtualized-table/hooks/use-heights.ts
```diff
@@ -41,6 +42,14 @@ const useHeights = ({ columns, columnWidths, pageInfo, headerRef, totalsRef, tot
     boldText: true,
   });
 
+  const resizeAllHeaderCells = headerHeight.resizeAllCells;
+  const resizeAllTotalCells = totalsHeight.resizeAllCells;
+
+  const resizeCells = useCallback(() => {
+    resizeAllHeaderCells();
+    resizeAllTotalCells();
+  }, [resizeAllHeaderCells, resizeAllTotalCells]);
```
a-m-dev, line 51: These dependencies are functions, right? 
My questions is, why this useCallback should depend on another useCallback(ed) function, which does not return anything in result, just changes some internal states of `use-dynamic-row-height`?

If those functions are already in memoized in their own useCallback, what is the point of using useCallback here once more? and is using useCallback here, affects the performance?

src/table/virtualized-table/Header.tsx
cbt1: Will fix 👍

src/table/virtualized-table/hooks/use-heights.ts
cbt1: Not that much point in it to be honest. If I don't wrap them in a useCallback, I would have to return both functions from this hook.

But thinking of it now. Skipping the useCallback, might actually make it a bit more clear  what going in the consuming component, as I would be "force" to change naming a bit.

a-m-dev (approved):

PR #938, opened 2023-03-22
refactor: remove unused code for sorting

haxxmaxx:
Since we remove the ability to pass a specific sort direction (ascending/descending) we can remove some logic and corresponding tests.


cbt1 (approved):
LTGM but it seems more like a fix than a refactor as you appear to have changed some logic

haxxmaxx:
> LTGM but it seems more like a fix than a refactor as you appear to have changed some logic

I have, but it is the same result. We were sending the a patch for `qInterColumnSortOrder` every time, which was not needed when sending patch for `qReverseSort`. So it doesn't fix anything, just a slight logic optimization you could say?

PR #939, opened 2023-03-22
fix: add translations to ascending and descending

AmeliaAndersson:
Changed two strings to be translated instead. The old enums are still needed for logic somewhere else so I kept them and just added a new constant.


cbt1 (approved):

haxxmaxx (commented):
Needs to be slightly different after you merge main

haxxmaxx (approved):

PR #943, opened 2023-03-24
fix: click to focus head

haxxmaxx:
- put `handleMouseDownToFocusHead` function back onto the header cell. Focuses the label button  if you click anywhere on the cell, except any of the two buttons. Usually don't get a focus border, but sometimes I do. Seem pretty [hard to control](https://css-tricks.com/almanac/selectors/f/focus-visible/#aa-how-do-browsers-determine-when-something-is-focus-visible), but I think it is a tiny problem.
- rename functions to match the event (click -> mouseDown)
- move keyboard listener to cell level, was on the `HeadCellContent` before. Mostly to have both mouse and keyboard listener, but it also makes sense since VT does need any a11y support atm. Thus i am passing `isInteractionEnabled` to `HeadCellContent`
- simplifying some test. I have no test on the cell component for handleMouseDown and handleKeyDown, since they would basically test React/mui


cbt1 (approved):
LGTM, I did some quick testing and it seem to work fine 👍

PR #944, opened 2023-03-24
fix: move pageRows update to TableWrapper

haxxmaxx:
SInce vt has its own hook to update `pageRows` on the selection state, and the deps for the two hooks are different, sometimes `pageRows` is overridden with empty array coming from use-selection-reducer. Then select multiple crashes. So put that hook in TableWrapper instead so it runs exclusively for pagination

### DIFF
--- a/src/table/hooks/use-selection-reducer.ts
+++ b/src/table/hooks/use-selection-reducer.ts
@@ -8,7 +8,8 @@ import useDidUpdateEffect from './use-did-update-effect';
 
 const useSelectionReducer = (
   pageRows: Row[],
-  selectionsAPI: ExtendedSelectionAPI
+  selectionsAPI: ExtendedSelectionAPI,
+  isPagination: Boolean
 ): [SelectionState, SelectionDispatch] => {
   const [selectionState, selectionDispatch] = useReducer(reducer, {
     pageRows,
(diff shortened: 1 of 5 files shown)


src/table/hooks/use-selection-reducer.ts
```diff
@@ -18,8 +19,11 @@ const useSelectionReducer = (
     isSelectMultiValues: false,
   });
 
+  // Only update pageRows for pagination table from here, VT is doing it in a different hook
   useDidUpdateEffect(() => {
-    selectionDispatch({ type: SelectionActions.UPDATE_PAGE_ROWS, payload: { pageRows } });
+    if (isPagination) {
```
cbt1, line 24: Was it not possible to move this hook call into `TableWrapper`? Would not need the `isPagination` flag than

haxxmaxx: never tried, actually. worth giving a try

haxxmaxx: I think we get an extra render, but it 10 vs 11 so it doesn't really matter. It was for confirming selections, so not that strange that a lot of things happen. Performance is not that important when rows updates

cbt1 (approved):

PR #950, opened 2023-03-25
fix: wait for fonts to load

cbt1:
Fixes an issue where the width of some text, was measured before all fonts had been loaded.

This could result in that the length of the text in the browser, when it had been rendered, did not match the measured values. Producing artifacts where, for example, text would look like it required 2 lines but was only rendered on a single line.

~~I've only added this check for the virtual scroll table as I've yet to seen any issue with the pagination table.~~

Edit: it actually runs for pagination table too.

Edit2: Added the check for pagination table as it also depends on measuring text using a canvas context.

### DIFF
--- a/src/nebula-hooks/virtualized-table/use-wait-for-fonts.ts
+++ b/src/nebula-hooks/virtualized-table/use-wait-for-fonts.ts
@@ -2,11 +2,13 @@ import { usePromise } from '@nebula.js/stardust';
 import { getBodyStyle, getHeaderStyle } from '../../table/utils/styling-utils';
 import { ExtendedTheme, TableLayout } from '../../types';
 
-const useWaitForFonts = (theme: ExtendedTheme, layout: TableLayout) => {
+const useWaitForFonts = (theme: ExtendedTheme, layout: TableLayout, shouldRender: boolean) => {
   const { fontSize: headerFontSize, fontFamily: headerFontFamily } = getHeaderStyle(layout, theme, false);
   const { fontSize, fontFamily } = getBodyStyle(layout, theme);
 
   const [isFontLoaded, error] = usePromise(async () => {
+    if (!shouldRender) return false;
+
     await document.fonts.load(`600 ${headerFontSize} ${headerFontFamily}`);
     await document.fonts.load(`${fontSize} ${fontFamily}`);
 
@@ -17,7 +19,7 @@ const useWaitForFonts = (theme: ExtendedTheme, layout: TableLayout) => {
   // log the error and continue rendering the table
   if (error) {
     // eslint-disable-next-line no-console
-    console.warn('Failed to load font styling from theme:', error);
+    console.warn('Failed to load font:', error);
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


src/nebula-hooks/virtualized-table/use-wait-for-fonts.ts
```diff
@@ -0,0 +1,27 @@
+import { usePromise } from '@nebula.js/stardust';
+import { getBodyStyle, getHeaderStyle } from '../../table/utils/styling-utils';
+import { ExtendedTheme, TableLayout } from '../../types';
+
+const useWaitForFonts = (theme: ExtendedTheme, layout: TableLayout) => {
+  const { fontSize: headerFontSize, fontFamily: headerFontFamily } = getHeaderStyle(layout, theme, false);
```
cbt1, line 6: Could consider to convert the `useTableStyling` hook into a Nebula hook, so we don't need to resolve the styling twice. But it's not a costly operation so not really sure it's worth it.

src/nebula-hooks/virtualized-table/use-wait-for-fonts.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { fontSize, fontFamily } = getBodyStyle(layout, theme);
+
+  const [isFontLoaded, error] = usePromise(async () => {
+    await document.fonts.load(`600 ${headerFontSize} ${headerFontFamily}`);
```
cbt1, line 10: Might be better ways to check this but I could not come up with one..

src/nebula-hooks/virtualized-table/use-wait-for-fonts.ts
```diff
@@ -0,0 +1,29 @@
+import { usePromise } from '@nebula.js/stardust';
+import { getBodyStyle, getHeaderStyle } from '../../table/utils/styling-utils';
+import { ExtendedTheme, TableLayout } from '../../types';
+
+const useWaitForFonts = (theme: ExtendedTheme, layout: TableLayout, shouldRender: boolean) => {
```
cbt1, line 5: If font family is changed between 2 renders. For example by changing theme. This will first return `true` before it has checked that the new font has been loaded.

It a scenario that I think can happen, if a font is somehow dynamically loaded between the two renders. But I do find it unlikely to happen so I did end up not added some additional logic to handle this case.

src/nebula-hooks/virtualized-table/use-wait-for-fonts.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  // An error while loading fonts should not block the table from rendering. So if an error occur,
+  // log the error and continue rendering the table
+  if (error) {
+    // eslint-disable-next-line no-console
+    console.warn('Failed to load font:', error);
+    return true;
+  }
+
+  return !!isFontLoaded;
```
haxxmaxx, line 26: can this be undefined or why the `!!`

cbt1: Yes, before the promise have resolved for the first time, `isFontLoaded` will be undefined.

haxxmaxx (approved):

PR #953, opened 2023-03-28
fix: improve measure text

cbt1:
Aims to improve some issue that could occur when line count was estimated. Should reduce the likely hood of an line count estimate that is lower than what is actually being rendered. So text should not be "clipped" in half.

Known issues:
- Numeric cells can wrap onto multiple lines
- Performance of `estimateLineCount` can be improved by using binary search
- An issue can still occur where line count is larger than what is actually being rendered

### DIFF
--- a/src/table/virtualized-table/hooks/use-measure-text.ts
+++ b/src/table/virtualized-table/hooks/use-measure-text.ts
@@ -1,9 +1,11 @@
 import { useMemo } from 'react';
 import { memoize } from 'qlik-chart-modules';
+import { MAX_NBR_LINES_OF_TEXT } from '../constants';
 
 export interface MeasureTextHook {
   estimateWidth: (length: number) => number;
   measureText: (text: string) => number;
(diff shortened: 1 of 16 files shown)


src/table/virtualized-table/hooks/use-measure-text.ts
```diff
@@ -22,8 +24,29 @@ export default function useMeasureText(
     return {
       measureText: (text) => memoizedMeasureText(text).width,
       estimateWidth: (length: number) => memoizedMeasureText(MAGIC_DEFAULT_CHAR).width * length,
+      estimateLineCount: (text: string, maxWidth: number) => {
```
haxxmaxx, line 27: this is the one you would optimize later right?

haxxmaxx (approved):
looks good, nice to get more constants in one place

src/table/virtualized-table/hooks/use-measure-text.ts
cbt1: Yes

PR #954, opened 2023-03-28
chore: remove strict mode in VT 

a-m-dev:
because of it's side effect on selection actions in development mode (double action triggering and making hard to track them down)


LiKang6688:
I am not into the idea to remove Strict Mode

Screenshot 2023-03-28 at 14 05 15
https://react.dev/learn/render-and-commit
https://react.dev/reference/react/StrictMode

a-m-dev:
yes, I was wrong about what the jira ticket asks for
but thanks Lee 🙏

PR #955, opened 2023-03-28
fix: use adjusted column idx for selections

haxxmaxx:
When calling the select method on the selection API, the column index should be adjusted for hidden columns. That means we need an additional column index, which is the `colIdx` with the number of hidden columns prior to it subtracted. Since it is only relevant for dimensions, `-1` is added for measures.

This is based on the dimensions in `qColumnOrder` being in the rising order. Meaning that when reordering dimensions, qDimensions is always updated so the first appearing dimension is the first dimension in the column order. (E.g. if you have 4 columns with 2 dimensions, swithing places between the two columns results in the same `qColumnOrder` but the dims switch place in `qDimensionInfo`.) This is not currently the case in sense. So this has to wait for that to be fixed


cbt1 (approved):
LGTM, but have not tested running it

a-m-dev (approved):

PR #956, opened 2023-03-28
fix: `PS_18291_SN_TABLE_BASIC_FEATURES` off-boarding

a-m-dev:


src/table/utils/handle-click.ts
```diff
@@ -39,30 +39,29 @@ export const getSelectionMouseHandlers = (
...
-      if (areBasicFeaturesEnabled) {
-        selectionDispatch({ type: SelectionActions.SELECT_MOUSE_UP, payload: { cell, evt, announce } });
-      } else if (cell.isSelectable) {
+      if (cell.isSelectable) {
         selectionDispatch({ type: SelectionActions.SELECT, payload: { cell, evt, announce } });
+      } else {
+        selectionDispatch({ type: SelectionActions.SELECT_MOUSE_UP, payload: { cell, evt, announce } });
       }
     }
   };
```
a-m-dev, line 67: Is this should be like my change? 
Also it's tests I'm a bit confused if it should be like my changes or not. 

Pls pay a bit close attention to this part and let me know if something is wrong 🙏

src/locale/__tests__/translations.spec.ts
```diff
@@ -9,6 +9,7 @@ describe('translations', () => {
     beforeEach(() => {
       translator = {
         get: () => 'SNTable.Accessibility.RowsAndColumns',
+        language: () => 'someLanguage',
```
haxxmaxx, line 12: is this part of offboarding the flag?

src/table/utils/handle-click.ts
haxxmaxx: you should think of it as if `areBasicFeaturesEnabled` is always true. That means that the else block will never run and you can remove it

src/locale/__tests__/translations.spec.ts
a-m-dev: Nope, that was my changes from nebula => causing types error because I linked it already, ✅ fixed now

src/table/utils/handle-click.ts
a-m-dev: Then the entire`SelectionActions.SELECT_MOUSE_UP` becomes obsolete? (i mean in `selection-utils.js`)

haxxmaxx: I meant the opposite. If you look at the original code and areBasicFeaturesEnabled is always true. You can remove the else block, meaning `SelectionActions.SELECT` is never called from here.

haxxmaxx: this is what you get left
```
if (evt.button === 0) {
  selectionDispatch({ type: SelectionActions.SELECT_MOUSE_UP, payload: { cell, evt, announce } });
}

cbt1:
I'm not seeing any changes in the `ext/` folder?

haxxmaxx (approved):

PR #957, opened 2023-03-28
chore: migrate from yarn to pnpm

LiKang6688:
- use pnpm as the package manager for better performance
- add pnpm workspace to enable lint/type check projects in mashup-examples (remove mashup-examples in tsconfig.exclude and .eslintignore) https://pnpm.io/workspaces
- use prettier-plugin-packagejson to make the order of properties in all package.json files nice https://github.com/matzkoh/prettier-plugin-packagejson
- migrate from resolutions to overrides in package.json https://pnpm.io/package_json#pnpmoverrides
- update circleci config to use pnpm https://pnpm.io/continuous-integration#circleci
- update target in tsconfig to es2017 to support await in the root file for mashup files


LiKang6688:
To be decided:
- have a complete table mashup example in https://github.com/qlik-oss/nebula.js-examples
- have a  live table mashup example in qlik.dev 
- test table mashup using tenant or engima-mocker

PR #959, opened 2023-03-29
fix: no word break for numeric cells

cbt1:


a-m-dev:
by this change, the pagination table and VT will not be in sync in header cell content values alignment. in pagination head cell contents snapped to bottom, but in VT they are snapped to top of cell. 
Is it ok?

Screenshot 2023-03-29 at 17 00 43

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -77,14 +77,14 @@ const useDynamicRowHeight = ({
   );
 
   const setCellSize = useCallback(
-    (text: string, rowIdx: number, colIdx: number, batchStateUpdate = false) => {
+    (text: string, rowIdx: number, colIdx: number, isNumeric = false, batchStateUpdate = false) => {
```
a-m-dev, line 80: if you have some tests cases of this function specifically for `batchStateUpdate`, then those might be affected by addition of `isNumeric` argument here

src/table/virtualized-table/hooks/use-heights.ts
```diff
@@ -43,7 +43,7 @@ const useHeights = ({ columns, columnWidths, pageInfo, headerRef, totalsRef, tot
 
   columns.forEach((col, idx) => {
     headerHeight.setCellSize(col.label, 0, idx);
-    totalsHeight.setCellSize(col.totalInfo, 0, idx);
+    totalsHeight.setCellSize(col.totalInfo, 0, idx, !Number.isNaN(+col.totalInfo));
```
a-m-dev, line 46: so many use cases of this and probably a candidate to be a mini util function?

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -170,7 +170,7 @@ const useDynamicRowHeight = ({
       const row = rows[rowIdx] ?? {};
       Object.values(row).forEach((cell) => {
         if (typeof cell === 'object') {
-          setCellSize(cell.qText ?? '', rowIdx, cell.pageColIdx);
+          setCellSize(cell.qText ?? '', rowIdx, cell.pageColIdx, cell.isNumeric);
```
a-m-dev, line 173: just a question our of curiosity: 
why using `mutableSetCellSize` in line 132 (inside of loop)
but in here we are **inside a loop again** but using `setCellSize`?

is this function `resizeVisibleCells` only being called once at when user is done with resizing? (if so, i think that should be fine to use `setCellSize` here then)

a-m-dev (approved):
LGTM, just some minor suggestions

src/table/virtualized-table/Cell.tsx
```diff
@@ -43,7 +43,7 @@ const Cell = ({ columnIndex, rowIndex, style, data }: CellProps) => {
           ...style,
           ...cellStyle,
           display: 'flex',
-          alignItems: 'center',
+          alignItems: 'start',
```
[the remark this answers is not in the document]
cbt1: It does appear so yes. As the rendering test have not failed. It should render just as before with only 1 row.

src/table/virtualized-table/hooks/use-measure-text.ts
```diff
@@ -24,10 +24,14 @@ export default function useMeasureText(
     return {
       measureText: (text) => memoizedMeasureText(text).width,
       estimateWidth: (length: number) => memoizedMeasureText(MAGIC_DEFAULT_CHAR).width * length,
-      estimateLineCount: (text: string, maxWidth: number) => {
+      estimateLineCount: (text: string, maxWidth: number, isNumeric: boolean) => {
```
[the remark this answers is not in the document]
cbt1: Will fix

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
cbt1: `mutableSetCellSize` did not exist when this function was created and as it was not really needed. There might be some performance optimizations that in using `mutableSetCellSize` instead. But that is not for this PR.

src/table/virtualized-table/hooks/use-heights.ts
a-m-dev: Yes

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
cbt1: Good catch! There was not any test affect by it, but it was something that I forgot to check

PR #961, opened 2023-03-30
perf: measure text

cbt1:
Use binary search to find the estimated line count

### DIFF
--- a/src/table/virtualized-table/hooks/use-measure-text.ts
+++ b/src/table/virtualized-table/hooks/use-measure-text.ts
@@ -10,6 +10,37 @@ export interface MeasureTextHook {
 
 const MAGIC_DEFAULT_CHAR = 'N';
 
+function getNextLine(text: string, maxWidth: number, fixedMeasureText: (text: string) => number) {
+  let left = 0;
+  let right = text.length;
+
+  while (left <= right) {
+    const m = Math.floor((left + right) / 2);
+    const chunk = text.slice(0, m);
+    const prevChunk = text.slice(0, m - 1);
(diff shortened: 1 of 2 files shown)


src/table/virtualized-table/hooks/use-measure-text.ts
```diff
@@ -21,34 +64,38 @@ export default function useMeasureText(
...
+        return 1;
+      }
+
+      if (text.length <= 1) {
+        return 1;
+      }
+
+      if (fixedMeasureText(text) <= maxWidth) {
+        return 1;
+      }
```
haxxmaxx: why not use one if with `||`?

src/table/virtualized-table/hooks/use-measure-text.ts
```diff
@@ -2,14 +2,61 @@ import { useMemo } from 'react';
...
+    if (width < maxWidth) {
+      // Search right
+      left = m + 1;
+    } else {
+      // Search left
+      right = m;
+    }
+  }
+
+  return null;
```
haxxmaxx, line 57: seems like this row is not covered in the test, is it testable?

haxxmaxx (approved):

src/table/virtualized-table/hooks/use-measure-text.ts
cbt1: Good find! But no, my thinking is that it's not. The only time it would reach here is when `fixedMeasureText(text) <= maxWidth` is true. But there is a guard for that before the while statement as a performance improvement.

I will however do some more digging into before my final conclusion.

PR #962, opened 2023-03-30
fix: empty totals label in VT mode

a-m-dev:
when totals label is undefined in layout => VT crashes => this PR is the fix from root cause of it where we safely return an empty string instead in order to prevent crash

### DIFF
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -52,7 +52,7 @@ export function getTotalPosition(layout: TableLayout) {
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
   if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText ?? '';
-  return pageColIdx === 0 ? layout.totals.label : '';
+  return pageColIdx === 0 ? layout.totals.label ?? '' : '';
 }
 
 /**
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -52,7 +52,8 @@ export function getTotalPosition(layout: TableLayout) {
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
   if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText ?? '';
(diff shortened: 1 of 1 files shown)


src/handle-data.ts
```diff
@@ -52,8 +52,7 @@ export function getTotalPosition(layout: TableLayout) {
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
   if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText ?? '';
-  if (pageColIdx === 0) return layout.totals.label;
-  return '';
+  return pageColIdx === 0 ? layout.totals.label : '';
```
cbt1, line 55: I might be blind but I see no difference here, the logic looks the same to me 😂

a-m-dev: You are right my mistake here 😵😅

I tried to shorten this:
```ts
if (pageColIdx === 0) return layout.totals.label ?? ''
```
to this: 
```ts
return pageColIdx === 0 ? layout.totals.label : '';
```

but i forgot the nullish thing
now fixed it

a-m-dev: maybe this is even more readable ?
```ts
if (pageColIdx === 0) return layout.totals.label ?? ''
```

src/handle-data.ts
```diff
@@ -52,8 +52,7 @@ export function getTotalPosition(layout: TableLayout) {
  */
 export function getTotalInfo(layout: TableLayout, colIdx: number, pageColIdx: number, numDims: number) {
   if (colIdx >= numDims) return layout.qHyperCube.qGrandTotalRow[colIdx - numDims]?.qText ?? '';
-  if (pageColIdx === 0) return layout.totals.label;
-  return '';
+  return pageColIdx === 0 ? layout.totals.label ?? '' : '';
```
cbt1, line 55: I would say that the old syntax is easier to ready, so something like this `if (pageColIdx === 0) return layout.totals.label ?? '';`

cbt1 (approved):

LiKang6688 (changes requested):
totals itself is also optional

LiKang6688 (approved):

a-m-dev:
So this one is still good to go?

LiKang6688:
> So this one is still good to go? 🤨

I believe so

haxxmaxx (approved):

PR #963, opened 2023-03-30
feat: update api spec

LiKang6688:
- update from experimental to stable
- remove columnWidths
- usePagination is optional
- totals and its properties are optional
- fix percentage description

Syntax reference: https://jsdoc.app/tags-type.html

### DIFF
--- a/src/qae/object-properties.js
+++ b/src/qae/object-properties.js
@@ -26,8 +26,6 @@ const properties = {
     qMeasures: [],
     /** @type {number[]} */
     qColumnOrder: [],
-    /** @type {number[]} */
-    columnWidths: [],
     qInitialDataFetch: [{ qWidth: 50, qHeight: 100 }],
   },
   /**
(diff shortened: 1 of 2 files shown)


src/qae/object-properties.js
```diff
@@ -26,8 +26,6 @@ const properties = {
     qMeasures: [],
     /** @type {number[]} */
     qColumnOrder: [],
-    /** @type {number[]} */
```
cbt1, line 29: This would be a breaking. But I assume that we don't need and/or use it anymore?

haxxmaxx: we never used it, so I think we would be fine

a-m-dev (approved):
LGTM, But @haxxmaxx and @cbt1 have more knowledge than me on this one!

cbt1 (approved):
LGTM but we need to deal with the API Gov when do a new release and the spec gets updated. As it did appear that  gov check fails with these changes.

src/qae/object-properties.js
LiKang6688: it turns out that columnWidths is used in https://github.com/qlik-oss/sn-table/blob/main/src/ext/data.js#L45

PR #964, opened 2023-03-31
test: add line break scenario

cbt1:
There was a gap in the coverage when it comes to rendering with line break. This PR aims to improve that.


a-m-dev:
In two snapshots, totals row is exceeding the table wrapper toward right, seems like something is not going correct in those cases: 

Screenshot 2023-03-31 at 09 08 11

cbt1:
> In two snapshots, totals row is exceeding the table wrapper toward right, seems like something is not going correct in those cases:
> 
> Screenshot 2023-03-31 at 09 08 11

Yes that is a bug that should be fixed. But will not do it as part of this PR

a-m-dev (approved):

PR #965, opened 2023-03-31
fix: totals background

cbt1:


a-m-dev:
I think this is a general issue between pagination table and VT that changing background color does not fixing it from root. 

This is pagination screenshot captured by rendering tests: 
----
Screenshot 2023-03-31 at 11 50 06


as you can see, the pagination component is 100% filled the wrapper, which means, pagination is kinda decoupled from table itself.

**This could not be a real case scenario,** because tables overall width could not be less than it's container. Even based on column resize, if you shrink down the width of every column -> the last one will fill up to the end and will not be 120px as here where you se last column of pagination table to have 120px width: https://github.com/qlik-oss/sn-table/blob/1f99159af9a95638ea48d69b2ff9ac8c9032c758/test/rendering/__fixtures__/scenario_15.fix.js#L213

👆 last columns should be something like: 
```js
// lastCol -> playwright chrome default device width - 3 of 120px cols => remaining will give us the real width for 4th col
const lastColWidth = 1280px - 3*(120px)
```


I suppose that should not be the case for VT as well!
-----
The last columns should fill up to the end, otherwise  we are not testing a real world scenario in that case. 

Screenshot 2023-03-31 at 11 51 24
Screenshot 2023-03-31 at 11 51 10



So my conclusion would be: 
---
For testing line break problem, it's better to keep last columns width filled up to the end of wrapper (that nebula gives to us and we get it's info using `useRect()` and using it inside of VT like [for example in totals](https://github.com/qlik-oss/sn-table/blob/fix/totals-background/src/table/virtualized-table/Totals.tsx#L26))

By that, we don't need to make totals row's background transparent and that was the reason behind my long comment, so sorry for that -> i needed to figure it out why that is the case 🙂

cbt1:
@a-m-dev I'm rather certain that you should be able to set a pixel based column width for ALL columns, including the last column. That was one of the requirements for that feature. So the tests are testing a real world scenario.

Regarding the pagination footer. I think we want it like that. As it allows for more space for the footer. Otherwise we might have to hide some of the components in the footer (like the dropdown), if the is not enough space.

Also the scrollbars should wrap the container (`useRect`) and not the "table", as in the rendered columns. So I do think the current behavior is what we want.

a-m-dev (approved):
I was not aware of that, So this should be fine in that case ✅

PR #966, opened 2023-03-31
fix: wait for fonts

cbt1:
The `useWaitForFonts` hook was not working as intended when both tables had it's own usage of it. You could end up in  scenario where toggling the `usePagination` "switch" did not re-render with a different table.

Now both use the same output from the hook, which I think makes more sense anyway as we don't neeed to make duplicate async `load` calls.


LiKang6688 (approved):
make sense to me

PR #967, opened 2023-03-31
fix: empty cell height

cbt1:
In the pagination table. Empty cells or cells with only white-space does not render with correct height.

Before:
Skärmavbild 2023-03-31 kl  13 36 03

After:
Skärmavbild 2023-03-31 kl  13 35 32


LiKang6688 (approved):
Awesome

a-m-dev (approved):

PR #968, opened 2023-03-31
revert: columnWidths

LiKang6688:
we need columnWidths for chart conversion. The implementation of chart conversion will be in another PR.

### DIFF
--- a/src/qae/object-properties.js
+++ b/src/qae/object-properties.js
@@ -26,6 +26,8 @@ const properties = {
     qMeasures: [],
     /** @type {number[]} */
     qColumnOrder: [],
+    /** @type {number[]} */
+    columnWidths: [],
     qInitialDataFetch: [{ qWidth: 50, qHeight: 100 }],
   },
   /**


src/qae/object-properties.js
```diff
@@ -26,6 +26,8 @@ const properties = {
     qMeasures: [],
     /** @type {number[]} */
     qColumnOrder: [],
+    /** @type {number[]} */
```
cbt1, line 29: I think the jsdoc tag should be removed as we still don't want it to be a part of the API spec.

Also I wonder if a better fix is to actually remove the usage of it from the "ext/data" code instead.

haxxmaxx: why do we need it for the chart conversion? I think we need to map `columnWidths` to the corresponding `columnWidth` on each column's `qDef`, and vice versa, when converting between old and new table

src/ext/__test__/data.spec.js
```diff
@@ -68,56 +67,4 @@ describe('data', () => {
       expect(env.translator.get).toHaveBeenCalledWith('Visualizations.Descriptions.Column');
     });
   });
-
-  describe('data', () => {
```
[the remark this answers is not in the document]
LiKang6688: Sure

src/qae/object-properties.js
LiKang6688: In the main branch, there is no `columnWidths: []`, after we change the columnWidths to the old table, we cannot convert it to new table.
Same for the new table, after we change the columnWidth to the new table, we cannot convert it to the old table. 
Screenshot 2023-03-31 at 11 37 49
Screenshot 2023-03-31 at 11 37 18

LiKang6688: Yes, it is

haxxmaxx (approved):

PR #971, opened 2023-04-03
feat: mapping column widths for chart conversion

LiKang6688:
when a column does not have type or has auto type, set that column to be the old table column width if provided


cbt1:
I don't think this is the way to handle conversion. You need to have the logic as part of the `importProperties` and `exportProperties` of the `ext` object.

PR #972, opened 2023-04-04
feat: restores all qCalcCondition for chart conversion

LiKang6688:
chart conversion task 1:

- enable conditional show for chart conversion by `conversion.conditionalShow.unquarantine(propertyTree.qProperty);`	
	-	when a column is hidden in an old table, after converting that old table to the new one, that column remains hidden and vice versa.

chart conversion task 2: column width will be in the next PR

PR #973, opened 2023-04-05
chore: sort the keys of a package.json file

LiKang6688:


cbt1:
Do we really need a 3rd party package to "maintain" the sort order? Can we not just run it once locally and be fine with that?

LiKang6688:
> Do we really need a 3rd party package to "maintain" the sort order? Can we not just run it once locally and be fine with that?

it is a prettier plugin that takes effect in `prettier --write .`

LiKang6688:
so we do not need to manually figure out alphabetic order https://github.com/qlik-oss/sn-table/pull/972#discussion_r1158078695_

cbt1:
> so we do not need to manually figure out alphabetic order [#972 (comment)](https://github.com/qlik-oss/sn-table/pull/972#discussion_r1158078695)_

But if you use `yarn add --dev <name-of-package>` they should be added in correct order to the deps list? For the other properties, I don't think we change the scripts or properties often enough

LiKang6688:
> > so we do not need to manually figure out alphabetic order [#972 (comment)](https://github.com/qlik-oss/sn-table/pull/972#discussion_r1158078695)_
> 
> But if you use `yarn add --dev <name-of-package>` they should be added in correct order to the deps list? For the other properties, I don't think we change the scripts or properties often enough

yes, make sense to me

LiKang6688:
> > so we do not need to manually figure out alphabetic order [#972 (comment)](https://github.com/qlik-oss/sn-table/pull/972#discussion_r1158078695)_
> 
> But if you use `yarn add --dev <name-of-package>` they should be added in correct order to the deps list? For the other properties, I don't think we change the scripts or properties often enough

Yes, make sense to me

PR #974, opened 2023-04-05
feat: mapping column widths for chart conversion

LiKang6688:
- For converting "sn-table to table", columnWidth is -1 unless the width type is pixels.
- For converting "table to sn-table", -1 -> fitToContent and the rest -> pixels

fix types

### DIFF
--- a/src/ext/conversion/__test__/import-properties.spec.ts
+++ b/src/ext/conversion/__test__/import-properties.spec.ts
@@ -34,6 +34,21 @@ describe('importProperties', () => {
       },
       qAttributeExpressions: [],
     },
+    {
+      qDef: {
+        qDef: '',
+        textAlign: {
+          auto: true,
+          align: 'left' as Align,
+        },
+        columnWidth: {
+          type: 'FitToContent' as ColumnWidthType,
+          pixels: 200,
+          percentage: 20,
+        },
+      },
+      qAttributeExpressions: [],
+    },
   ];
 
   describe('getColumnInfo', () => {
@@ -55,16 +70,6 @@ describe('importProperties', () => {
 
       newColumnInfo = getColumnInfo(qMeasures, 1, columnWidths, 1);
       expect(newColumnInfo.qDef.columnWidth).toEqual({ pixels: 20, type: 'pixels' });
-    });
-
-    it('should get the correct columnInfo given columnWidths with -1 and pixels', () => {
-      const columnWidths = [400, -1];
-      let newColumnInfo = getColumnInfo(qDimensions, index, columnWidths);
-
-      expect(newColumnInfo.qDef.columnWidth).toEqual({ pixels: 400, type: 'pixels' });
-
-      newColumnInfo = getColumnInfo(qMeasures, 1, columnWidths, 1);
-      expect(newColumnInfo.qDef.columnWidth).toEqual({ type: 'fitToContent' });
     });
 
     it('should get the correct columnInfo given columnWidths with empty array', () => {
@@ -95,13 +100,26 @@ describe('importProperties', () => {
       expect(measures).toEqual(qMeasures);
     });
 
-    it('should get correct dimensions and measures when no qColumnOrder', () => {
+    it('should get correct dimensions and measures when no qColumnOrder provided', () => {
       const qColumnOrder = [] as number[];
-      const columnWidths = [400, -1];
+      const columnWidths = [400, -1, 300];
 
       const { dimensions, measures } = getMultiColumnInfo(qDimensions, qMeasures, qColumnOrder, columnWidths);
       expect(dimensions[0].qDef.columnWidth).toEqual({ pixels: 400, type: 'pixels' });
       expect(measures[0].qDef.columnWidth).toEqual({
+        type: 'fitToContent',
+      });
+      expect(measures[1].qDef.columnWidth).toEqual({ pixels: 300, type: 'pixels' });
+    });
+
+    it('should get correct dimensions and measures', () => {
+      const qColumnOrder = [2, 0, 1];
+      const columnWidths = [400, -1, 300];
+
+      const { dimensions, measures } = getMultiColumnInfo(qDimensions, qMeasures, qColumnOrder, columnWidths);
+      expect(dimensions[0].qDef.columnWidth).toEqual({ pixels: 400, type: 'pixels' });
+      expect(measures[0].qDef.columnWidth).toEqual({ pixels: 300, type: 'pixels' });
+      expect(measures[1].qDef.columnWidth).toEqual({
         type: 'fitToContent',
       });
     });
@@ -115,15 +133,15 @@ describe('importProperties', () => {
           measures: qMeasures,
           excludedDimensions: [],
           excludedMeasures: [],
-          interColumnSortOrder: [0, 1],
+          interColumnSortOrder: [0, 1, 2],
         },
       ],
       properties: {
         qHyperCubeDef: {
           qDimensions: [],
           qMeasures: [],
-          qColumnOrder: [0, 1],
-          columnWidths: [300, -1],
+          qColumnOrder: [0, 1, 2],
+          columnWidths: [300, -1, 400],
         },
       },
     } as unknown as ExportFormat;
@@ -147,6 +165,10 @@ describe('importProperties', () => {
       });
       expect(propertyTree.qProperty.qHyperCubeDef.qMeasures[0].qDef.columnWidth).toEqual({
         type: 'fitToContent',
+      });
+      expect(propertyTree.qProperty.qHyperCubeDef.qMeasures[1].qDef.columnWidth).toEqual({
+        pixels: 400,
+        type: 'pixels',
       });
     });
   });
--- a/src/ext/conversion/__test__/export-properties.spec.ts
+++ b/src/ext/conversion/__test__/export-properties.spec.ts
@@ -78,6 +78,12 @@ describe('exportProperties', () => {
       const columnWidths = getColumnWidths(qDimensions, qMeasures);
       expect(columnWidths).toEqual([-1, 200, -1]);
     });
+
+    it('should get correct columnWidths when qColumnOrder provided', () => {
+      const qColumnOrder = [1, 2, 0];
+      const columnWidths = getColumnWidths(qDimensions, qMeasures, qColumnOrder);
+      expect(columnWidths).toEqual([200, -1, -1]);
+    });
   });
 
   describe('exportProperties', () => {


src/ext/conversion/importProperties.ts
```diff
@@ -1,13 +1,49 @@
...
 
 import data from '../../qae/data';
-import { ExportFormat, PropTree } from '../../types';
+import { ExportFormat, PropTree, DimensionProperties, MeasureProperties } from '../../types';
+import { ColumnWidthTypes } from '../../table/constants';
+
+export const getColumnInfo = (
+  columnInfo: DimensionProperties[] | MeasureProperties[],
+  colIdx: number,
+  columnWidths: unknown,
```
haxxmaxx, line 11: I think you should be able to type this, isn't it always a number[] or undefined?

src/ext/conversion/importProperties.ts
```diff
@@ -18,7 +54,30 @@ const importProperties = (
...
+  const columnsLength = numDims + qMeasures.length;
+  const columnOrder = qColumnOrder?.length === columnsLength ? qColumnOrder : Array.from(Array(columnsLength).keys());
+
+  qDimensions = columnOrder
+    .filter((colIdx: number) => colIdx < numDims)
+    .map((colIdx: number) => getColumnInfo(qDimensions, colIdx, columnWidths));
+
+  qMeasures = columnOrder
+    .filter((colIdx: number) => colIdx >= numDims)
+    .map((colIdx: number) => getColumnInfo(qMeasures, colIdx, columnWidths, numDims));
```
haxxmaxx, line 75: you could just pass `colIdx - numdims` and you don't have to pass numDims and have logic for that in `getColumnInfo`

src/ext/conversion/importProperties.ts
```diff
@@ -1,13 +1,49 @@
...
+
+export const getColumnInfo = (
+  columnInfo: DimensionProperties[] | MeasureProperties[],
+  colIdx: number,
+  columnWidths: unknown,
+  numDims?: number
+) => {
+  let index = colIdx;
+  if (numDims) index = colIdx - numDims;
+  const column = columnInfo[index];
```
haxxmaxx, line 16: if you spread into a new variable here, you can just do column.qDef.columnWidth = { ... } in the if statement

src/ext/conversion/__test__/exportProperties.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+            },
+          ],
+        },
+      },
+    } as unknown as PropTree;
+    const hyperCubePath = undefined;
+
+    const expFormat = exportProperties(propertyTree, hyperCubePath);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
haxxmaxx, line 115: why do you have to do this?

src/ext/conversion/exportProperties.ts
```diff
@@ -1,12 +1,36 @@
...
   const expFormat = conversion.hypercube.exportProperties({
     propertyTree,
     hyperCubePath,
   });
+  const {
+    qHyperCubeDef: { qColumnOrder, qDimensions, qMeasures },
+  } = expFormat.properties;
+  const columnsLength = qDimensions.length + qMeasures.length;
+  const columnOrder = qColumnOrder?.length === columnsLength ? qColumnOrder : Array.from(Array(columnsLength).keys());
+  const columnWidths = columnOrder.map((colIdx: number) => getColumnWidth(colIdx, qDimensions, qMeasures));
```
haxxmaxx, line 27: you could bake this into one function, then it would be easy to test if it works without `qColumnOrder`.

src/ext/conversion/importProperties.ts
haxxmaxx: and then only have one return at the bottom

haxxmaxx (commented):
Also, since these are not component files, the names should be in kebab case, e.g. import-properties.ts

src/ext/conversion/__test__/exportProperties.spec.ts
LiKang6688: cause in types, propertyTree.qProperty.qHyperCubeDef.qMeasures[0].qDef does not have columnWidth

src/types.ts
```diff
@@ -5,24 +5,26 @@ export type Align = 'left' | 'center' | 'right';
...
+  align: Align;
 }
 
 // properties
 interface InlineDimensionDef extends EngineAPI.INxInlineDimensionDef {
   textAlign: TextAlign;
+  columnWidth: ColumnWidth;
 }
-interface InlineMeasureDef extends EngineAPI.INxInlineMeasureDef {
+interface InlineMeasureDef extends Omit<EngineAPI.INxInlineMeasureDef, 'qDef'> {
```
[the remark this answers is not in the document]
LiKang6688: Yes, we should. No, DimensionProperties/MeasureProperties have qDef.

src/ext/conversion/importProperties.ts
LiKang6688: empty array, [] as well?

haxxmaxx: not sure, but that would be covered by number[]

LiKang6688: right, then it is fine to type as `columnWidths?: number[],`

src/ext/conversion/importProperties.ts
LiKang6688: either have `colIdx - numdims` in parameters or inside function. I prefer to have that inside function

```
getColumnInfo = (
  columnInfo: DimensionProperties[] | MeasureProperties[],
  colIdx: number,
  columnWidths?: number[],
  measureIndex?: number
) => {
  let index = colIdx;
  if (measureIndex) index = measureIndex;
  const column = columnInfo[index];
...
```

src/types.ts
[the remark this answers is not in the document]
haxxmaxx: they have `qDef` yes, and that is set to `InlineMeasureDef` for measure. And now you omit `qDef` from `InlineMeasureDef`, is there a qMeasures[i].qDef.qDef property? In that case, is it only for measure?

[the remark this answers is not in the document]
LiKang6688: yes, it is only for measure. And It is not needed.
```

/**
     * NxInlineMeasureDef...
     */
    interface INxInlineMeasureDef {
		...
		/**
         * Definition of the expression in the measure.
         * Example: Sum (OrderTotal)
         * >> This parameter is mandatory.
         */
        qDef: string;
        ...
	}
```

src/ext/conversion/__test__/exportProperties.spec.ts
haxxmaxx: then can't we set the extended hypercube type on the propertyTree?

LiKang6688: that is in @types/qlik-engineapi and with a very deep hierarchy. I do not know how to do that yet.

src/types.ts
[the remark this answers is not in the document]
haxxmaxx: We are not using it but it should still be there. Why are you omitting it in the first place?

src/ext/conversion/importProperties.ts
haxxmaxx: well that makes less sense to me, if you want to have logic for it, fine. But at least pass only one index. So leave it as it is if you want

src/types.ts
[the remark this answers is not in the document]
LiKang6688: qDef is not optional there. If we do not omit it, tsc complains.

src/ext/conversion/__test__/exportProperties.spec.ts
haxxmaxx: wouldn't that be just be setting qProperty to something like this:
```
export interface PropTree {
  qChildren: unknown[];
  qProperty: { qHyperCubeDef: QHyperCubeDef }
}
```

in types.ts?

haxxmaxx: the ExportFormat one should this be the generic from engine API, since that is not the sn-table props, I think

LiKang6688: good idea, I gonna do that.

src/types.ts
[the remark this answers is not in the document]
haxxmaxx: where?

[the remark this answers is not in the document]
LiKang6688: https://github.com/qlik-oss/sn-table/pull/974/files#diff-f48e6e66c8fc7fa397b3358fda246cf5b160b01e53853d22fbb000691a387fd2R56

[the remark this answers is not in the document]
haxxmaxx: Then just add an arbitrary expression when you are mocking. 
```
        qDef: {
          qDef: 'someExpression',
          ...
```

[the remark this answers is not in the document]

src/ext/conversion/importProperties.ts
LiKang6688: Fixed

src/ext/conversion/importProperties.ts
LiKang6688: Fixed

src/ext/conversion/__test__/exportProperties.spec.ts
LiKang6688: Fixed

src/ext/conversion/exportProperties.ts
LiKang6688: Fixed

src/ext/conversion/import-properties.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const getColumnInfo = (
+  columnInfo: DimensionProperties[] | MeasureProperties[],
+  colIdx: number,
+  columnWidths?: number[],
+  numDims?: number
+) => {
+  let index = colIdx;
+  if (numDims) index = colIdx - numDims;
+  const column = JSON.parse(JSON.stringify(columnInfo[index]));
```
haxxmaxx, line 16: why not { ...columnInfo[index] } ?

src/ext/conversion/__test__/import-properties.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+        },
+        columnWidth: {
+          type: 'auto' as ColumnWidthType,
+          pixels: 200,
+          percentage: 20,
+        },
+      },
+      qAttributeExpressions: [],
+    },
+  ];
```
haxxmaxx, line 37: the importProperties runs when you convert TO a sn-table right? then why would there be a `columnWidth` property on the qMeasures/qDimensions?

src/ext/conversion/__test__/import-properties.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    it('should get the correct columnInfo given columnWidths with -1 and pixels', () => {
+      const columnWidths = [400, -1];
+      let newColumnInfo = getColumnInfo(qDimensions, index, columnWidths);
+
+      expect(newColumnInfo.qDef.columnWidth).toEqual({ pixels: 400, type: 'pixels' });
+
+      newColumnInfo = getColumnInfo(qMeasures, 1, columnWidths, 1);
+      expect(newColumnInfo.qDef.columnWidth).toEqual({ type: 'fitToContent' });
+    });
```
haxxmaxx, line 68: I don't see the point of having three tests here. You are basically doing the first two in the third test. Which you might as well test in `getMultiColumnInfo`

src/ext/conversion/__test__/import-properties.spec.ts
haxxmaxx: It would just leave the column with no width info. Which is fine, since it is optional

haxxmaxx (commented):
General comment on the tests:
I think we should tests `getColumnWidths` and `getMultiColumnInfo`. then you automatically get the helper functions. Plus we need to test with a specified qColumnOrder and an undefined one.

src/ext/conversion/import-properties.ts
LiKang6688: { ...columnInfo[index] } is for shallow copy, and we need deep copy here
https://makimo.pl/blog/shallow-and-deep-copies-in-javascript/#:~:text=We%20have%20two%20kinds%20of,that%20contain%20only%20primitive%20values.&text=Nested%20objects%20mean%20objects%20that%20contain%20non%2Dprimitive%20values.

haxxmaxx: huh, I thought spreading was doing a deep copy. Really good to know. Then this is fine

PR #977, opened 2023-04-06
fix: adjuster keyboard issues

haxxmaxx:
Found two small issues with the column adjust when using the keyboard.
- lost focus after pressing arrow once, need to stop propagation
- focus was set to chart, not ... menu. Same there, needed to stop propagation

Also, learned a bit about focus-visible a while ago, so we can use that pseudo-class instead of manually blurring when using the mouse.

### DIFF
--- a/src/table/components/head/ColumnAdjuster.tsx
+++ b/src/table/components/head/ColumnAdjuster.tsx
@@ -17,7 +17,7 @@ const ColumnAdjuster = ({ column, isLastColumn, onColumnResize }: AdjusterProps)
   const { applyColumnWidths, constraints } = useContextSelector(TableContext, (value) => value.baseProps);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
   const setColumnWidths = useContextSelector(TableContext, (value) => value.setColumnWidths);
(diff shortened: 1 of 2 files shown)


src/table/components/head/ColumnAdjuster.tsx
```diff
@@ -77,12 +74,14 @@ const ColumnAdjuster = ({ column, isLastColumn, onColumnResize }: AdjusterProps)
       updateWidth(adjustedWidth);
     } else if (event.key === KeyCodes.SPACE || event.key === KeyCodes.ENTER) {
       preventDefaultBehavior(event);
-      confirmWidth();
       focusHeadMenuButton(event);
+
+      confirmWidth();
     } else if (event.key === KeyCodes.ESC) {
+      preventDefaultBehavior(event);
```
a-m-dev, line 81: what if we call this function before these if ... else cases? 
now you are calling it in all 3 possible cases

haxxmaxx: yes but we don't want to call it for all keyboard events

a-m-dev (approved):

PR #981, opened 2023-04-12
fix: remove sanity check for rect.width === 0

haxxmaxx:
When rect.width becomes 0, we early returned and set `columnWIdths` to `[]`. That means some calculations for the VT broke. There is no real reason to early return. The result will be that all columns get the min size.

### DIFF
--- a/src/table/hooks/use-column-widths.ts
+++ b/src/table/hooks/use-column-widths.ts
@@ -27,7 +27,7 @@ type GetFitToContentWidth = (headLabel: string, totalsLabel: string, glyphCount:
  * The widths are sorted in the order they will be displayed
  */
 export const getColumnWidths = (columns: Column[], tableWidth: number, getFitToContentWidth: GetFitToContentWidth) => {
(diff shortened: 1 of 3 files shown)


src/table/hooks/use-column-widths.ts
```diff
@@ -27,7 +27,7 @@ type GetFitToContentWidth = (headLabel: string, totalsLabel: string, glyphCount:
  * The widths are sorted in the order they will be displayed
  */
 export const getColumnWidths = (columns: Column[], tableWidth: number, getFitToContentWidth: GetFitToContentWidth) => {
-  if (!columns?.length || tableWidth === 0) return [];
+  if (!columns?.length) return [];
```
haxxmaxx, line 30: Still keeping this sanity check. Not sure if it is needed, but not causing any issues either

cbt1 (approved):
LGTM, but it's worth saying that I think this issue will occur anytime the length `columnWidths` is not in sync with the length of columns. That should not happen tho.

PR #984, opened 2023-04-13
Crl/styling key

haxxmaxx:


src/ext/property-panel/settings.js
```diff
@@ -8,7 +8,6 @@ const stylingPanel = {
       translation: 'LayerStyleEditor.component.styling',
       ref: 'components',
       useGeneral: true,
-      key: 'theme',
```
LiKang6688, line 11: no key is needed?

LiKang6688:
```
"components": [
    {
      "key": "general",
      "title": {
        "main": {
          "color": {
            "index": -1,
            "color": "#6bde8a"
          }
        }
      },
      "bgColor": {
        "color": {
          "index": 13,
          "color": "#138185"
        }
      }
    },
    {
      "key": "theme",
      "content": {
        "fontSize": null
      },
      "header": {
        "fontSize": null
      }
    }
  ]
```

is from Properties. 
so `"key": "general"`, should be used here?

LiKang6688 (approved):

PR #987, opened 2023-04-17
fix: default value for pixels

haxxmaxx:


a-m-dev (approved):

src/ext/conversion/export-properties.ts
```diff
@@ -11,7 +12,7 @@ export const getColumnWidth = (colIdx: number, qDimensions: DimensionProperties[
   const { columnWidth } = info.qDef;
 
   // For converting "sn-table to table", columnWidth is -1 unless the width type is pixels.
-  return columnWidth?.type === 'pixels' ? columnWidth.pixels : -1;
+  return columnWidth?.type === 'pixels' ? columnWidth.pixels ?? DEFAULT_COLUMN_PIXEL_WIDTH : -1;
```
LiKang6688, line 15: is it possible that when columnWidth?.type is 'pixels', but columnWidth.pixels has no value?

haxxmaxx: yes. If you convert from old table to sn-table, you only have the type set to fitToContent. If you then change to pixels, there is no pixel value set. It works in the code since we have a fallback in use-column-widths. But if you then convert back to old table, you get null values in the columnWidths array

PR #988, opened 2023-04-17
chore: update playwright test config

LiKang6688:
- update playwright config 
- use toHaveScreenshot instead of toMatchSnapshot


test/rendering/utils/run-rendering-tests.ts
```diff
@@ -67,10 +68,8 @@ const runRenderingTests = (theme: Object | Function, themeType: String, language
         const element = page.locator('text=Zocalo');
         await element.scrollIntoViewIfNeeded();
       }
-      // Playwright captures screenshot
-      const img = await playwright.screenshot();
-      // Compare screenshot with baseline image
-      expect(img).toMatchSnapshot(`${name}.png`);
```
LiKang6688, line 73: toMatchSnapshot() now should be used for comparing text or arbitrary binary data
https://playwright.dev/docs/test-snapshots

playwright.config.ts
```diff
@@ -24,26 +28,26 @@ const config: PlaywrightTestConfig = {
   // Forbid test.only on CI
   forbidOnly: !!process.env.CI,
   // The maximum number of retry attempts per test, two retries on CI
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 1 : 0,
```
LiKang6688, line 31: retries one time is enough

playwright.config.ts
```diff
@@ -24,26 +28,26 @@ const config: PlaywrightTestConfig = {
...
   // Limit the number of workers
-  workers: 2,
+  workers: 6,
```
LiKang6688, line 33: 3 type of browser * 2 test files

test/integration/integration.config.ts
```diff
@@ -18,17 +20,22 @@ const config: PlaywrightTestConfig = {
...
     {
       name: 'chromium',
       use: { ...devices['Desktop Chrome'] },
     },
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+    // {
+    //   name: 'webkit',
```
LiKang6688, line 35: integration testing does not work correctly in Webkit yet

haxxmaxx:
So why do we need 174 snapshots? I think we need a smarter approach to this. Maintaining that many snapshots will be a lot of work

LiKang6688:
> 



> So why do we need 174 snapshots? I think we need a smarter approach to this. Maintaining that many snapshots will be a lot of work

it starts to test against not only one browser but three browsers

cbt1:
> > 
> 
> > So why do we need 174 snapshots? I think we need a smarter approach to this. Maintaining that many snapshots will be a lot of work
> 
> it starts to test against not only one browser but three browsers more workers are used in CI to keep the same running time

I would agree with @haxxmaxx here, it going to be a pain to look at each and every snapshot when they fail.

Do we need to run 3 different browsers?
Do we need to run the same test in 2 different "configurations"? It think it language and/or theme now?

LiKang6688:
> > > 
> > 
> > 
> > > So why do we need 174 snapshots? I think we need a smarter approach to this. Maintaining that many snapshots will be a lot of work
> > 
> > 
> > it starts to test against not only one browser but three browsers more workers are used in CI to keep the same running time
> 
> I would agree with @haxxmaxx here, it going to be a pain to look at each and every snapshot when they fail.
> 
> Do we need to run 3 different browsers? Do we need to run the same test in 2 different "configurations"? It think it language and/or theme now?

it is also painful for us to introduce bugs in the other browsers that we miss.

I would argue that there are 2 different configurations for the same test. One is for rendering tests and the other one is for a11y tests.

haxxmaxx:
> it is also painful for us to introduce bugs in the other browsers that we miss.
> 
> I would argue that there are 2 different configurations for the same test. One is for rendering tests and the other one is for a11y tests.

That is true, but if you have this many saspshots, it is very unlikely you will look through them all properly, so I think it is very likely you assume that the changes are valid and a bug slips through anyway.

I think we need to be smarter about the fixtures. Right now it looks like there are 15 * 2.  One with pagination, one without. I think we can get that number down a lot. 15 in itself is high, and then only run some of them with pagination (or without, depending what we think is most important). Then we have two themes. Same thing here, I pretty sure we don't have do everything again with a different them. I'd say this should be couple of extra tests.

Some quick math, let's say we have 12 fixtures, then run maybe 5 of those with pagination. Then run a small number of them for a different theme, say 5. Summing that up and doing all of them for each browser that is 66. Still very high but maybe manageable.

LiKang6688:
> > it is also painful for us to introduce bugs in the other browsers that we miss.
> > I would argue that there are 2 different configurations for the same test. One is for rendering tests and the other one is for a11y tests.
> 
> That is true, but if you have this many saspshots, I suspecting you wont look through them all properly, so I think it is very likely you assume that the changes are valid and a bug slips through anyway.
> 
> I think we need to be smarter about the fixtures. Right now it looks like there are 15 * 2. One with pagination, one without. I think we can get that number down a lot. 15 in itself is high, and then only run some of them with pagination (or without, depending what we think is most important). Then we have two themes. Same thing here, I pretty sure we don't have do everything again with a different them. I'd say this should be couple of extra tests.
> 
> Some quick math, let's say we have 12 fixtures, then run maybe 5 of those with pagination. Then run a small number of them for a different theme, say 5. Summing that up and doing all of them for each browser that is 66. Still very high but maybe manageable.

Sure, I will look into your proposal

PR #990, opened 2023-04-18
test: keyboard navigation using tab

LiKang6688:


test/integration/keyboard.integration.ts
```diff
@@ -17,6 +17,17 @@ test.describe('Tests served by: fixture-file rendering by Nebula', () => {
...
+  const isSVG = async () => {
+    const index = (await page.locator('*:focus').innerHTML()).match(/svg/)?.index;
+    if (index && index > 0) return true;
+    return false;
+  };
+  const isOption = async () => {
+    const index = (await page.locator('*:focus').innerHTML()).match(/option/)?.index;
+    if (index && index > 0) return true;
+    return false;
+  };
```
haxxmaxx, line 30: There aren't really any util functions defined in this file. Do they fit into some other file or maybe create a new one?

haxxmaxx (commented):
I don't mind adding the test. Maybe we need to do some shift+tab as well. But this is deviating from the the test structure in the arrow key test. Would be nice if we can write new tests in a similar way. Doesn't have to be as exactly as the arrow test. But much of the same things are going on: Do something (in this case tab) then check where we are. Should be possible to define the steps and then loop over them

test/integration/keyboard.integration.ts
LiKang6688: Okay, I will create a new file to relocate them.

LiKang6688:
> I don't mind adding the test. Maybe we need to do some shift+tab as well. But this is deviating from the the test structure in the arrow key test. Would be nice if we can write new tests in a similar way. Doesn't have to be as exactly as the arrow test. But much of the same things are going on: Do something (in this case tab) then check where we are. Should be possible to define the steps and then loop over them

I dislike the loop, I see that it causes lots of trouble for me to debug. That is the reason I write steps like it now.

haxxmaxx:
> > I don't mind adding the test. Maybe we need to do some shift+tab as well. But this is deviating from the the test structure in the arrow key test. Would be nice if we can write new tests in a similar way. Doesn't have to be as exactly as the arrow test. But much of the same things are going on: Do something (in this case tab) then check where we are. Should be possible to define the steps and then loop over them
> 
> I dislike the loop, I see that it causes lots of trouble for me to debug. That is the reason I write steps like it now.

Good to know, since this is a new way of writing tests that Csaba set up. Ok then it's fine writing them like this. But then there needs to be at least some comments explaining what is going on. Moving to headers, pagination controls etc

PR #992, opened 2023-04-20
fix: hover colors

haxxmaxx:
making sure the hover color is using the correct reference. Also updating the rendering tests, since this should be closer to what the props are in sense.

### DIFF
--- a/test/rendering/__fixtures__/scenario_10.fix.js
+++ b/test/rendering/__fixtures__/scenario_10.fix.js
@@ -126,6 +126,9 @@ export default () => ({
         showDetailsExpression: false,
         components: [
           {
+            key: 'general',
+          },
+          {
             key: 'theme',
             content: {
               fontSize: 22,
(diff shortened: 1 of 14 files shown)


test/rendering/__fixtures__/scenario_10.fix.js
```diff
@@ -125,6 +125,9 @@ export default () => ({
         showDetails: false,
         showDetailsExpression: false,
         components: [
+          {
+            key: 'general',
```
a-m-dev, line 129: Since you are using `.find`, probably you dont need this!

a-m-dev (approved):
LGTM

test/rendering/__fixtures__/scenario_10.fix.js
haxxmaxx: yeah this should have been here earlier and we would have caught this bug. But I added it anyway, might help us catch bugs in the future

PR #994, opened 2023-04-24
fix: quaratine and unquarantine column order

bao-ho:
## Description

Quarantine and unquarantine column order of sn-table so that the property can be stored and recovered correctly.

### DIFF
--- a/src/ext/conversion/__test__/export-properties.spec.ts
+++ b/src/ext/conversion/__test__/export-properties.spec.ts
@@ -84,7 +84,7 @@ describe('exportProperties', () => {
           qDimensions,
           qMeasures,
           qColumnOrder: [0, 2, 1],
-          columnOrder: [0, 1],
+          columnOrder: [0, 2, 1],
         },
       },
     } as unknown as PropTree;
@@ -102,6 +102,6 @@ describe('exportProperties', () => {
     ).toEqual([0, 2, 1]);
     expect(
       expFormat.properties.qLayoutExclude.quarantine.straightTableColumnOrder['qHyperCubeDef.columnOrder']
-    ).toEqual([0, 1]);
+    ).toEqual([0, 2, 1]);
   });
 });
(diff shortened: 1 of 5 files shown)


src/ext/conversion/export-properties.ts
```diff
@@ -44,6 +44,8 @@ const exportProperties = (propertyTree: PropTree, hyperCubePath?: string): Expor
   setValue(expFormat, 'properties.qHyperCubeDef', qHyperCubeDef);
 
   conversion.quarantineProperty(expFormat.properties, 'qHyperCubeDef.columnWidths', 'straightTableColumnWidths');
+  conversion.quarantineProperty(expFormat.properties, 'qHyperCubeDef.qColumnOrder', 'straightTableColumnOrder');
```
LiKang6688, line 47: qHyperCubeDef.qColumnOrder and qHyperCubeDef.columnOrder are same?

bao-ho: yes, I assume columnOrder is an old property that has been replaced by qColumnOrder. We still need to support columnOrder for the old object to work, I think.

LiKang6688: you are right, sn-table->old-table->sn-table

LiKang6688 (approved):
LGTM

src/ext/conversion/__test__/export-properties.spec.ts
```diff
@@ -83,6 +83,8 @@ describe('exportProperties', () => {
         qHyperCubeDef: {
           qDimensions,
           qMeasures,
+          qColumnOrder: [0, 2, 1],
+          columnOrder: [0, 1],
```
LiKang6688, line 87: we make qColumnOrder and columnOrder same here to migrate confusion?

bao-ho: sure! I will fix that.

haxxmaxx (approved):
Nice that you added the tests :)

PR #998, opened 2023-05-01
chore: clean up duplicated packages in lock file

LiKang6688:

PR #1000, opened 2023-05-06
chore: check duplicated packages in CI

LiKang6688:
- Check duplicated packages in CI
- Check duplicated packages in "lint-staged"
- Add yarn-deduplicat as a devDep and a script for it that fixes the issues

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,9 @@ jobs:
             //npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
             //npm.pkg.github.com/:always-auth=true" >> ~/.npmrc
       - run:
+          name: Check duplicated packages
+          command: npx yarn-deduplicate --list --fail
+      - run:
           name: Install dependencies
           command: yarn install --frozen-lockfile
       - save_cache: *save_yarn_cache
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,11 +34,11 @@ jobs:
             //npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
             //npm.pkg.github.com/:always-auth=true" >> ~/.npmrc
       - run:
-          name: Check duplicated packages
-          command: npx yarn-deduplicate --list --fail
-      - run:
           name: Install dependencies
           command: yarn install --frozen-lockfile
+      - run:
+          name: Check duplicated packages
+          command: yarn deduplicate-packages --list --fail
       - save_cache: *save_yarn_cache
 
       - store_artifacts:
--- a/.github/CONTRIBUTION.md
+++ b/.github/CONTRIBUTION.md
@@ -68,7 +68,7 @@ You can skip git commit hooks with:
 
 Yarn v1 does not support package deduplication natively. yarn-deduplicate is used to cleans up yarn.lock by removing duplicates:
 
-      npx yarn-deduplicate yarn.lock
+      yarn deduplicate-packages
 
 ## Releasing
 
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -33,6 +33,9 @@ jobs:
             echo "@qlik-trial:registry=https://npm.pkg.github.com/
             //npm.pkg.github.com/:_authToken=$GITHUB_TOKEN
             //npm.pkg.github.com/:always-auth=true" >> ~/.npmrc
+      - run:
```
haxxmaxx, line 36: shouldn't you do this after you install? Or am I misunderstanding how deduplicate works?

.github/CONTRIBUTION.md
```diff
@@ -64,6 +64,12 @@ You can skip git commit hooks with:
 
     git commit -m "commit message" --no-verify
 
+## Deduplicate packages
+
+Yarn v1 does not support package deduplication natively. yarn-deduplicate is used to cleans up yarn.lock by removing duplicates:
+
+      npx yarn-deduplicate yarn.lock
+
```
haxxmaxx, line 72: This feels a bit out of context. All the other docs here explain normal things you do when developing. This is more of a maintenance thing. And pretty generic, it just describes what `yarn-deduplicat` does. I think if we want it it in here, then we should at least have a command for it, if that is possible

cbt1:
What happens when the job fails? How would you fix the issue?

You consider adding `yarn-deduplicat` as a devDep and a script for it that fixes the issues. Also consider adding a "lint-staged" check for it on changes to yarn.lock so that you cannot commit a lock file that will fail on CI.

.circleci/config.yml
LiKang6688: It does not matter.  As `npx yarn-deduplicate` check yarn.lock file and `yarn install --frozen-lockfile` does not change yarn.lock file

.github/CONTRIBUTION.md
LiKang6688: `npx yarn-deduplicate yarn.lock` is the command.

LiKang6688:
> What happens when the job fails? How would you fix the issue?
> 
> You consider adding `yarn-deduplicat` as a devDep and a script for it that fixes the issues. Also consider adding a "lint-staged" check for it on changes to yarn.lock so that you cannot commit a lock file that will fail on CI.

Sure

package.json
```diff
@@ -139,7 +141,8 @@
     "stylis-plugin-rtl-sc": "npm:stylis-plugin-rtl@1.1.0",
     "ts-jest": "29.1.0",
     "typescript": "5.0.4",
-    "yargs": "17.7.2"
+    "yargs": "17.7.2",
+    "yarn-deduplicate": "^6.0.1"
```
cbt1, line 145: Pin the version and run `yarn` again
```suggestion
    "yarn-deduplicate": "6.0.1"
```

LiKang6688: Thx

.github/CONTRIBUTION.md
haxxmaxx: ah you changed it, nice

cbt1 (approved):
LGTM but I would have expected the `yarn.lock` file to be updated a bit more. Like do we not actually have that many duplicate packages in there?

LiKang6688:
> LGTM but I would have expected the `yarn.lock` file to be updated a bit more. Like do we not actually have that many duplicate packages in there?


No, now it is fairly good as I deduplicated the package before here https://github.com/qlik-oss/sn-table/pull/998. 
When we bump versions of packages, we need to manually deduplicate packages like here https://github.com/qlik-oss/sn-table/pull/999

PR #1001, opened 2023-05-08
feat: support taking a snapshot

quanho:
This PR aims at supporting taking a snapshot for pagination table.


cbt1:
There are 2 different tables in sn-table and they work quite differently and provides different feature sets. That is something that needs to be considering when add support for snapshots, as they likely requires 2 different solutions.

We should have a chat how they work and what makes them different. The proposed solution here won't really work as soon as the user have scrolled.

src/services/view-service.ts
```diff
@@ -0,0 +1,17 @@
...
+  const { snapshotData } = layout;
+  const isSnapshot = !!snapshotData;
+
+  return {
+    qLeft: isSnapshot ? snapshotData?.content?.qLeft || 0 : 0,
+    qTop: isSnapshot ? snapshotData?.content?.qTop || 0 : 0,
+    qWidth: isSnapshot ? snapshotData?.content?.qWidth || 0 : 0,
+    qHeight: isSnapshot ? snapshotData?.content?.qHeight || 0 : 0,
+    scrollLeft: isSnapshot ? snapshotData?.content?.scrollLeft || 0 : 0,
+    scrollTop: isSnapshot ? snapshotData?.content?.scrollTop || 0 : 0,
```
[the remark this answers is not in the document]
quanho: Right!

src/ext/index.js
```diff
@@ -31,6 +31,7 @@ export default function ext(env) {
       snapshot: true,
       viewData: false,
       exploration: true,
+      cssScaling: true,
```
LiKang6688, line 34: What is cssScaling?

src/table/pagination-table/components/TableWrapper.tsx
```diff
@@ -78,8 +78,10 @@ function TableWrapper(props: TableWrapperProps) {
   useKeyboardActiveListener();
 
   useEffect(() => {
-    tableContainerRef.current?.scrollTo(0, 0);
-  }, [pageInfo, totalRowCount]);
+    const left = layoutService.isSnapshot ? layoutService.layout.snapshotData?.content?.scrollLeft || 0 : 0;
+    const top = 0;
+    tableContainerRef.current?.scrollTo(left, top);
+  }, [pageInfo, totalRowCount, layoutService]);
```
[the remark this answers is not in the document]
quanho: Updated!

src/ext/index.js
quanho: For resize a snapshot! If the table size when we captured a snapshot is 200 x 300 and then in story we resize the table to 100 x 180 then we use css scaling for this. So the transform will scale (0.5)

src/table/hooks/use-layout-service.ts
```diff
@@ -0,0 +1,7 @@
+import { useMemo } from '@nebula.js/stardust';
+import createLayoutService from '../../services/layout-service';
+import { TableLayout, LayoutService } from '../../types';
+
+const useLayoutService = (layout: TableLayout): LayoutService => useMemo(() => createLayoutService(layout), [layout]);
```
haxxmaxx, line 5: custom hooks that we do using startust hooks, we normally store in src/nebula-hooks

src/table/hooks/use-view-service.ts
```diff
@@ -0,0 +1,7 @@
+import { useMemo } from '@nebula.js/stardust';
+import createViewService from '../../services/view-service';
+import type { ViewService } from '../../types';
+
+const useViewService = (): ViewService => useMemo(() => createViewService(), []);
```
haxxmaxx, line 5: I think `createViewService` can live in this file as well.

src/table/pagination-table/components/TableWrapper.tsx
```diff
@@ -122,6 +128,7 @@ function TableWrapper(props: TableWrapperProps) {
         tabIndex={-1}
         role="application"
         data-testid="table-container"
+        onScroll={onScroll}
```
haxxmaxx, line 131: This is a bit messy because we manually add a listener to this element, in use-scroll-listener. For some reason we are listening to `wheel` events, which is not the same as `scroll`. So this should actually be the better way of doing it, and consider adding listener here as well

src/table/utils/find-visible-rows.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const findEndIndex = (rows: HTMLCollectionOf<Element>, min: number, max: number): number => {
+  for (let i = rows.length - 1; i > -1; i--) {
+    const rowRect = rows[i].getBoundingClientRect();
+    const center = rowRect.y + rowRect.height / 2;
+    if (center >= min && center <= max) {
+      return i;
+    }
+  }
+  return -2;
```
haxxmaxx, line 22: why -2?

src/types.ts
```diff
@@ -3,6 +3,26 @@ import { stardust } from '@nebula.js/stardust';
...
+  qTop: number;
+  qWidth: number;
+  qHeight: number;
+  scrollLeft: number;
+}
+export interface SnapshotData {
+  content?: {
+    scrollLeft: number;
+  };
+  object: {
```
LiKang6688, line 21: when object.size is assigned to snapshotData?

src/table/utils/find-visible-rows.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const findVisibleRows = (rootElement: HTMLElement, totalsPosition: TotalsPosition) => {
```
haxxmaxx, line 25: Are you adding tests for this in anyway?

src/types.ts
```diff
@@ -90,6 +110,22 @@ export interface TableLayout extends Omit<EngineAPI.IGenericHyperCubeLayout, 'qH
...
+  y: number;
+}
+export interface LayoutService {
+  layout: TableLayout;
+  isSnapshot: boolean;
+  size: Point;
+}
+
+export interface SnapshotLayout extends EngineAPI.IGenericObjectLayout {
+  title?: string;
```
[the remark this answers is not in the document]
quanho: Updated!

src/types.ts
quanho: Good point! 
It is from sense client https://github.com/qlik-trial/sense-client/blob/master/web/assets/objects/snapshot/snapshot-helper.js#L81 but we don't need it in sn-table.

src/table/hooks/use-snapshot.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    if (!snapshotLayout.snapshotData.content) {
+      if ((model as EngineAPI.IGenericObject)?.getHyperCubeData) {
+        if (!snapshotLayout.qHyperCube) {
+          snapshotLayout.qHyperCube = {} as HyperCube;
+        }
+        const totalsPosition = getTotalPosition(layoutService.layout);
+        const { visibleRowStartIndex = -1, visibleRowEndIndex = -2 } = findVisibleRows(rootElement, totalsPosition);
+        snapshotLayout.qHyperCube.qDataPages = await (model as EngineAPI.IGenericObject).getHyperCubeData(
```
LiKang6688, line 30: when taking a snapshot using nebula, model is undefined. So it does not work.

LiKang6688: Therefore, we need to handle qDataPages from layout.qHyperCube like https://github.com/qlik-oss/sn-org-chart/blob/master/src/snapshot.js#L13

src/services/layout-service.ts
```diff
@@ -0,0 +1,21 @@
+import { TableLayout, LayoutService } from '../types';
+
+const MAX_COLUMN_COUNT = 1000;
+
+const createLayoutService = (layout: TableLayout): LayoutService => {
```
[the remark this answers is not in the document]
quanho: We can use layout for now!

src/table/hooks/use-snapshot.ts
LiKang6688: Screenshot 2023-05-15 at 15 19 26

I had a try

src/table/utils/find-visible-rows.ts
quanho: To make sure visibleRowEndIndex - visibleRowStartIndex + 1 is not bigger than 0. But I will refactor this to avoid confusion.

src/table/hooks/use-snapshot.ts
```diff
@@ -0,0 +1,52 @@
...
+  onTakeSnapshot(async (snapshotLayout: SnapshotLayout) => {
+    if (!snapshotLayout.snapshotData) {
+      return snapshotLayout;
+    }
+
+    if (!model) {
+      return snapshotLayout;
+    }
+
+    if (!snapshotLayout.snapshotData.content) {
```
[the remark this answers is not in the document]
quanho: Good point! Updated!

src/table/hooks/use-layout-service.ts
quanho: Removed it!

src/table/pagination-table/components/TableWrapper.tsx
haxxmaxx: In other words, ignore this comment

src/table/hooks/use-scroll-listener.ts
```diff
@@ -1,22 +1,31 @@
...
+    const scrollCallback = (evt: Event) => {
+      viewService.scrollLeft = (evt.currentTarget as Element)?.scrollLeft || 0;
+    };
+
     memoedContainer.addEventListener('wheel', horizontalScrollCallback);
+    memoedContainer.addEventListener('scroll', scrollCallback);
 
     return () => {
       memoedContainer.removeEventListener('wheel', horizontalScrollCallback);
+      memoedContainer.removeEventListener('scroll', scrollCallback);
```
haxxmaxx, line 26: ah very nice, then we get all the listener in one place

haxxmaxx:
Going on vacation now, so I'd say it looks good except that I would like to see some tests if possible on find-visible-rows. I'll let the others do the rest of the review

src/table/hooks/use-snapshot.ts
LiKang6688: I was wrong. Never mind.

LiKang6688 (approved):
Awesome!

src/table/utils/find-visible-rows.ts
quanho: Added!

PR #1002, opened 2023-05-08
fix: selectionsAPI type

LiKang6688:
- fix: selectionsAPI type
when taking and rendering snapshots as well as exporting those snapshots to images, selectionsAPI (useSelections()) is undefined

- fix: embed is not optional
- refactor: use undefined or optional properly https://tkdodo.eu/blog/optional-vs-undefined
- refactor: remove unneeded manageData in RenderProps

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,16 +37,16 @@ export default function supernova(env: Galaxy) {
       const rootElement = useElement();
       const reactRoot = useReactRoot(rootElement);
       const layout = useStaleLayout() as TableLayout;
-      const app = useApp();
-      const model = useModel();
+      const app = useApp(); // undefined when taking snapshot
+      const model = useModel(); // undefined when taking snapshot
       const constraints = useConstraints();
       const translator = useTranslator() as ExtendedTranslator;
-      const selectionsAPI = useSelections() as ExtendedSelectionAPI | undefined;
+      const selectionsAPI = useSelections() as ExtendedSelectionAPI | undefined; // undefined when taking snapshot
       const keyboard = useKeyboard();
       const rect = useRect();
       const embed = useEmbed();
       const theme = useExtendedTheme(rootElement);
-      const changeSortOrder = useSorting(layout.qHyperCube, model);
+      const changeSortOrder = useSorting(layout.qHyperCube, model); // undefined when taking snapshot
       const applyColumnWidths = useApplyColumnWidths(layout.qHyperCube, model);
       const isFontLoaded = useWaitForFonts(theme, layout);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,7 @@ export default function supernova(env: Galaxy) {
       const model = useModel();
       const constraints = useConstraints();
       const translator = useTranslator() as ExtendedTranslator;
-      const selectionsAPI = useSelections() as ExtendedSelectionAPI;
+      const selectionsAPI = useSelections() as ExtendedSelectionAPI | undefined;
       const keyboard = useKeyboard();
       const rect = useRect();
       const embed = useEmbed();
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,8 +46,8 @@ export default function supernova(env: Galaxy) {
       const rect = useRect();
       const embed = useEmbed();
       const theme = useExtendedTheme(rootElement);
-      const changeSortOrder = useSorting(model, layout.qHyperCube);
-      const applyColumnWidths = useApplyColumnWidths(model, layout.qHyperCube);
+      const changeSortOrder = useSorting(layout.qHyperCube, model);
+      const applyColumnWidths = useApplyColumnWidths(layout.qHyperCube, model);
       const isFontLoaded = useWaitForFonts(theme, layout);
 
       extendContextMenu();
--- a/src/table/context/index.ts
+++ b/src/table/context/index.ts
@@ -1,2 +1,2 @@
 export { useContextSelector } from './useContextSelector';
-export { TableContext, TableContextProvider } from './TableContext';
+export { TableContext, TableContextProvider, EMPTY_TABLE_DATA } from './TableContext';
--- a/src/table/pagination-table/components/head/TableHeadWrapper.tsx
+++ b/src/table/pagination-table/components/head/TableHeadWrapper.tsx
@@ -14,16 +14,17 @@ import { handleMouseDownToFocusHead } from '../../../utils/handle-mouse';
 
 function TableHeadWrapper() {
   const { columns } = useContextSelector(TableContext, (value) => value.tableData);
-  const { layout, selectionsAPI, styling, constraints, rootElement, keyboard } = useContextSelector(
+  const { layout, styling, constraints, rootElement, keyboard } = useContextSelector(
     TableContext,
     (value) => value.baseProps
   );
   const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
-  const isInSelectionMode = selectionsAPI?.isModal();
+  const isSelectionMode = useContextSelector(TableContext, (value) => value.baseProps.selectionsAPI?.isModal());
+
   const headRowRef = useRef<HTMLTableRowElement>(null);
-  const isInteractionEnabled = !constraints.active && !isInSelectionMode;
+  const isInteractionEnabled = !constraints.active && !isSelectionMode;
 
   useEffect(() => {
     if (headRowRef.current) {
--- a/src/table/pagination-table/components/head/TableHeadWrapper.tsx
+++ b/src/table/pagination-table/components/head/TableHeadWrapper.tsx
@@ -14,14 +14,14 @@ import { handleMouseDownToFocusHead } from '../../../utils/handle-mouse';
 
 function TableHeadWrapper() {
   const { columns } = useContextSelector(TableContext, (value) => value.tableData);
-  const { layout, styling, constraints, rootElement, keyboard } = useContextSelector(
+  const { layout, selectionsAPI, styling, constraints, rootElement, keyboard } = useContextSelector(
     TableContext,
     (value) => value.baseProps
   );
   const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
-  const isInSelectionMode = useContextSelector(TableContext, (value) => value.baseProps.selectionsAPI.isModal());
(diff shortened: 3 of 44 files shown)


src/table/virtualized-table/HeaderCell.tsx
```diff
@@ -23,9 +23,9 @@ const HeaderCell = ({ index, style, data }: HeaderCellProps) => {
     columResizeHandler,
   } = data;
 
-  const { layout, constraints } = useContextSelector(TableContext, (value) => value.baseProps);
+  const { layout, selectionsAPI, constraints } = useContextSelector(TableContext, (value) => value.baseProps);
   const showRightBorder = useContextSelector(TableContext, (value) => value.showRightBorder);
-  const isInSelectionMode = useContextSelector(TableContext, (value) => value.baseProps.selectionsAPI.isModal());
+  const isInSelectionMode = selectionsAPI?.isModal();
```
cbt1, line 28: A recall there being good reason for the "selector" to return the value of isModal. Maybe @haxxmaxx remembers?

haxxmaxx: at least for the pagination table, not exactly sure here. The reason is that we want to trigger a reload when the return of isModal changes (e.g. going in or out of selection mode). But maybe that happens anyway for the virtualized table? we still use staleLayout right?

src/table/pagination-table/components/head/TableHeadWrapper.tsx
```diff
@@ -14,14 +14,14 @@ import { handleMouseDownToFocusHead } from '../../../utils/handle-mouse';
...
-  const { layout, styling, constraints, rootElement, keyboard } = useContextSelector(
+  const { layout, selectionsAPI, styling, constraints, rootElement, keyboard } = useContextSelector(
     TableContext,
     (value) => value.baseProps
   );
   const setHeadRowHeight = useContextSelector(TableContext, (value) => value.setHeadRowHeight);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
-  const isInSelectionMode = useContextSelector(TableContext, (value) => value.baseProps.selectionsAPI.isModal());
+  const isInSelectionMode = selectionsAPI?.isModal();
```
haxxmaxx, line 24: so this would introduce a bug at least, we need to turn off interactions when you go into selectionmode

src/table/utils/__tests__/selections-utils.spec.ts
```diff
@@ -70,18 +70,18 @@ describe('selections-utils', () => {
 
         const newState = reducer(state, action);
         expect(newState).toEqual({ ...state, rows: { [cell.qElemNumber]: cell.rowIdx }, colIdx: cell.colIdx });
-        expect(state.api.begin).toHaveBeenCalledTimes(1);
-        expect(state.api.select).toHaveBeenCalledWith({ method: 'selectHyperCubeCells', params });
-        expect(state.api.cancel).not.toHaveBeenCalled();
+        expect(state.api?.begin).toHaveBeenCalledTimes(1);
```
haxxmaxx, line 73: Here you have defined state.api so it should always be defined. Does lint complain or something?

src/table/virtualized-table/HeaderCell.tsx
cbt1: I think it's safest to just leave the selector as it was before with optional chaining to fix the type error.

src/table/utils/accessibility-utils.ts
```diff
@@ -201,7 +201,7 @@ export const focusSelectionToolbar = (element: HTMLElement, keyboard: stardust.K
 export const announceSelectionState = (
   announce: Announce,
   nextCell: HTMLTableCellElement,
-  isSelectionMode: boolean
+  isSelectionMode?: boolean
```
haxxmaxx, line 204: a default value here would be clean IMO

src/table/utils/handle-keyboard.ts
```diff
@@ -54,7 +54,7 @@ export const handleBodyKeyUp = (evt: React.KeyboardEvent, selectionDispatch: Sel
 /**
  * Manually focuses the selection toolbar if tabbing from the last focusable element
  */
-export const handleLastTab = (evt: React.KeyboardEvent, isSelectionMode: boolean, keyboard: stardust.Keyboard) => {
+export const handleLastTab = (evt: React.KeyboardEvent, keyboard: stardust.Keyboard, isSelectionMode?: boolean) => {
```
haxxmaxx, line 57: a default value here and it doesn't have to be optional when you pass it on

haxxmaxx (commented):
So would be nice to settle this ambiguity once and for, mostly for my own sake 🙃 

Let's say you have a function call something like this 
```
const { arg1, arg2 } = props;
someFunc(arg1, arg2)
```
now let's say that arg2 **can have the value** undefined. So in the code above you always pass both arguments, meaning it isn't really optional. And I would then type it like this:
`const someFunc = (arg1: string, arg2: string | undefined) => {...`
Mostly to indicate that you should always give it both arguments, but then it could be the case that arg2 is undefined.

The other case is that you actually have an optional parameter. `shouldBubbleEarly` is a good example of that. You can pass a second argument to indicate the status of selection mode, but you don't have to pass it, since it is not applicable for all uses. And then the typing makes sense like this:
`const someFunc = (arg1: string, arg2?: string) => {...`

So, then my question is. Are they equivalent? Any reason that it is better practice to have all "optional/possibly undefined" arguments typed as optional and not `| undefined`?

LiKang6688:
> So would be nice to settle this ambiguity once and for, mostly for my own sake 🙃
> 
> Let's say you have a function call something like this
> 
> ```
> const { arg1, arg2 } = props;
> someFunc(arg1, arg2)
> ```
> 
> now let's say that arg2 **can have the value** undefined. So in the code above you always pass both arguments, meaning it isn't really optional. And I would then type it like this: `const someFunc = (arg1: string, arg2: string | undefined) => {...` Mostly to indicate that you should always give it both arguments, but then it could be the case that arg2 is undefined.
> 
> The other case is that you actually have an optional parameter. `shouldBubbleEarly` is a good example of that. You can pass a second argument to indicate the status of selection mode, but you don't have to pass it, since it is not applicable for all uses. And then the typing makes sense like this: `const someFunc = (arg1: string, arg2?: string) => {...`
> 
> So, then my question is. Are they equivalent? Any reason that it is better practice to have all "optional/possibly undefined" arguments typed as optional and not `| undefined`?

No, they are not equivalent.
?: This parameter is optional
| undefined: This parameter is required, but can be undefined.

When I choose which one to use, I check if it makes sense to call the function without that parameter. If it is not required, I would go for ?.

haxxmaxx:
> > > No, they are not equivalent. ?: This parameter is optional | undefined: This parameter is required, but can be undefined.
> > > When I choose which one to use, I check if it makes sense to call the function without that parameter. If it is not required, I would go for ?.
> > 
> > 
> > Ok that makes sense, but then I am not sure I agree with which of these args that have been changed. `model` for example, is it undefined when rendering a snapshot? If not, I would say it is not optional. It just happens to be undefined on the first render (or very early, before we get it from nebula). So it CAN be undefined, but it will become defined in comming renders. I feel like that is a common react pattern, and it doesn't make arguments optional, but rather not defined yet
> 
> model does not happen to be undefined on the first render when rendering a snapshot. It is undefined all the way.

ok then it makes sense 👍 So the selection api and model are undefined fro snapshots? anything else? could be good to clarify in the PR description for back-tracking

LiKang6688:
> > > > No, they are not equivalent. ?: This parameter is optional | undefined: This parameter is required, but can be undefined.
> > > > When I choose which one to use, I check if it makes sense to call the function without that parameter. If it is not required, I would go for ?.
> > > 
> > > 
> > > Ok that makes sense, but then I am not sure I agree with which of these args that have been changed. `model` for example, is it undefined when rendering a snapshot? If not, I would say it is not optional. It just happens to be undefined on the first render (or very early, before we get it from nebula). So it CAN be undefined, but it will become defined in comming renders. I feel like that is a common react pattern, and it doesn't make arguments optional, but rather not defined yet
> > 
> > 
> > model does not happen to be undefined on the first render when rendering a snapshot. It is undefined all the way.
> 
> ok then it makes sense 👍 So the selection api and model are undefined fro snapshots? anything else? could be good to clarify in the PR description for back-tracking

Good point. I will add those.

src/table/utils/__tests__/selections-utils.spec.ts
LiKang6688: As SelectionState.api is optional

src/table/utils/__tests__/selections-utils.spec.ts
```diff
@@ -253,7 +253,7 @@ describe('selections-utils', () => {
 
       it('should return state unchanged when the app is in selection modal state and action.type is reset', () => {
         const action = { type: SelectionActions.RESET } as ResetAction;
-        state.api.isModal = () => true;
+        (state.api as ExtendedSelectionAPI).isModal = () => true;
```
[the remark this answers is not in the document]
LiKang6688: as it is cased as `as unknown as ExtendedSelectionAPI`
state.api is optional in first place

src/table/utils/accessibility-utils.ts
LiKang6688: Agree. Done

src/index.ts
```diff
@@ -37,17 +37,17 @@ export default function supernova(env: Galaxy) {
       const rootElement = useElement();
       const reactRoot = useReactRoot(rootElement);
       const layout = useStaleLayout() as TableLayout;
-      const app = useApp();
-      const model = useModel();
+      const app = useApp(); // undefined when taking snapshot
```
cbt1, line 40: These are all undefined when taking snapshot in nebula dev tool right? So not in Sense?

LiKang6688: that's correct

cbt1: Consider stating that then

cbt1 (approved):
LGTM, but cannot speak for what you and Max ended up deciding regarding optional props

src/index.ts
LiKang6688: These are undefined all the way for in snapshot, so i do not think they should be stating

src/table/context/index.ts
```diff
@@ -1,2 +1,2 @@
 export { useContextSelector } from './useContextSelector';
-export { TableContext, TableContextProvider } from './TableContext';
+export { TableContext, TableContextProvider, EMPTY_TABLE_DATA } from './TableContext';
```
haxxmaxx, line 2: is this used somewhere?

LiKang6688: Yes, https://github.com/qlik-oss/sn-table/pull/1002/files#diff-ae735badc3af4f2cafbd6d1d059127069d4c4447e56a042cb4f618049493b39eL62

LiKang6688: as https://github.com/qlik-oss/sn-table/blob/9023f518201fbc3b9bc9db515b87ace64e7be0d8/src/table/context/TableContext.tsx#L24 tableData is set with EMPTY_TABLE_DATA

haxxmaxx (approved):

PR #1008, opened 2023-05-17
fix: QB-19034 Straight table - horizontal borders disappear for certain font sizes

sharleneaqlik:
By setting the font size in the styling panel, some values make the totals bottom border disappear. Seems to have to do with incorrect rounding. All affected font sizes result in cells with a height that ends with .6667. It is every third font size starting at five, so 5, 8, 11, 14, 17 and so on. Seems like both header and content need these sizes for the bug to appear. Also only a problem when pagination is toggled off

Actual behavior:
The botton border for totals disappears when the font size for header and content is a 3n + 2 px

Expected behavior:
the border are present independent of font size

Step by step instructions:
Create a Straight Table from the viz bundle
set the font size for the header and content to e.g. 14px.
make sure the use pagination is toggled off
the border disappears


src/table/virtualized-table/Body.tsx
```diff
@@ -135,7 +135,7 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
       data-key="body"
       ref={gridRef}
       innerRef={innerForwardRef}
-      style={{ overflow: 'hidden' }}
+      style={{ overflow: 'hidden', willChange: 'auto' }}
```
Donya-qlik, line 138: May I ask what is WillChange?

Donya-qlik: Oh interesting! I didn't know that! tnx

LiKang6688 (changes requested):
TD

src/table/virtualized-table/Body.tsx
LiKang6688: Are you sure that VariableSizeGrid also needs to have WillChange?

src/table/virtualized-table/Body.tsx
```diff
@@ -135,7 +135,7 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
       data-key="body"
```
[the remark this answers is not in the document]
cbt1: Would guess that it was to "tag" the element so that it could be identified in the browser when inspecting. But it does not work. `VariableSizeGrid` just ignores it.

cbt1:
Overall this looks good to me. I do recommend that you do as Li suggested regarding styling, ofc also fixing the rendering tests. Just make sure that the border renders as expected when running this on your machine. The rendering test does seem to change quite a bit. I did some testing locally and it looked good in that one chart at atleast.

src/table/virtualized-table/Body.tsx
sharleneaqlik: @LiKang6688 : this was specifically mentioned by @cbt1 , Is there any specific reason you think this may not be needed in `VariableSizeGrid `?

sharleneaqlik:
> screenshots for rendering tests needs to be updated

Yes, saw the tests failing. On it to update them.

src/table/virtualized-table/Body.tsx
LiKang6688: as I see in the sn-pivot-table, VariableSizeGrid's WillChange is not changed. 
And it is called listStyle, it is confusing for me that it is also applied to VariableSizeGrid

src/table/virtualized-table/Body.tsx
```diff
@@ -135,7 +149,7 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
...
       ref={gridRef}
       innerRef={innerForwardRef}
-      style={{ overflow: 'hidden' }}
+      style={listStyle}
```
LiKang6688, line 152: I do not think it is good to use the name listStyle for VariableSizeGrid. Also, VariableSizeGrid does not need boxSizing: 'border-box' as far as i see.

sharleneaqlik: listStyle for VariableSizeGrid was changed. Strange that you didn't see the change. But I didn't see your review comment 2 hours ago either. This is strange. By the way, where do you see `boxSizing: 'border-box'` property? I don't see that in Body.tsx

src/table/virtualized-table/Header.tsx
```diff
@@ -5,6 +5,21 @@ import HeaderCell from './HeaderCell';
 import { useContextSelector, TableContext } from '../context';
 import useResetHeader from './hooks/use-reset-header';
 
+const listStyle: React.CSSProperties = {
```
LiKang6688, line 8: There are two same listStyle, i think it is good to remove the duplication.

LiKang6688 (approved):

PR #1014, opened 2023-05-29
feat: support snapshoting for virtualized table

quanho:


src/table/Root.tsx
```diff
@@ -102,7 +103,7 @@ export function renderVirtualizedTable(props: VirtualTableRenderProps, reactRoot
           pageInfo={pageInfo}
           initialDataPages={initialDataPages}
         >
-          <VirtualizedTable />
+          <VirtualizedTable viewService={viewService} />
```
haxxmaxx, line 106: Not sure if we want to put it here or on the context. Especially since there are no other props passed to `VirtualizedTable`

quanho: I updated it by moving viewService to the context.

src/table/virtualized-table/Body.tsx
```diff
@@ -144,6 +157,13 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
     [innerForwardRef, bodyHeight, rowMeta, rowCount, resizeVisibleCells, rowsInPage]
   );
 
+  const scrollLeft = layout.snapshotData ? viewService.scrollLeft : 0;
+  const scrollTopRatio = layout.snapshotData ? viewService.scrollTopRatio || 0 : 0;
+
+  useEffect(() => {
+    interpolatedScrollTo(scrollTopRatio, scrollLeft);
+  }, [scrollTopRatio, scrollLeft]);
```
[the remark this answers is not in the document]
quanho: This is similar to https://github.com/qlik-oss/sn-table/blob/main/src/table/pagination-table/components/TableWrapper.tsx#L84

src/table/Root.tsx
```diff
@@ -50,8 +51,9 @@ export function renderPaginationTable(props: RenderProps, reactRoot?: ReactDom.R
           changeSortOrder={changeSortOrder}
           applyColumnWidths={applyColumnWidths}
           rect={rect}
+          viewService={viewService}
         >
-          <TableWrapper {...wrapperProps} direction={direction} />
+          <TableWrapper {...wrapperProps} direction={direction} viewService={viewService} />
```
cbt1, line 56: `viewService` is passed to the context provider and the `TableWrapper` component?

src/table/virtualized-table/Body.tsx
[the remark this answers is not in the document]
cbt1: @haxxmaxx `scrollTopRatio, scrollLeft` are not state variables. So changing them won't trigger a re-render thankfully. But on initial/first render it will tell `react-window` to scroll to a position, that might trigger a internal re-render of the `react-window` components. I do not think that is an issue, either the value is `0` and nothing should happen or the value is greater than `0` and it should scroll to the correct position in the snapshot. Given that all states are correct.

However not including `interpolatedScrollTo` in the dependency array might cause bugs. So I recommend finding a solution that allows you to include it. (wrap `interpolatedScrollTo` in a `useCallback` might to it..)

src/table/virtualized-table/Body.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { memo, useMemo, useLayoutEffect, useRef, useImperativeHandle, forwardRef } from 'react';
+/* eslint-disable react-hooks/exhaustive-deps */
```
cbt1, line 1: Remove this and find a better solution. See other comment about a possible solution.

src/table/Root.tsx
quanho: Updated!

quanho:
> Just did some quick testing and a few things that did not work for me:
> 
> * Create a virtual scroll table with a few 100 rows. Scroll to the middle. Take snapshot. Add snapshot to storytelling -> Empty cells are rendered.
> * Create a virtual scroll table with more than 250k rows. Go to page 2. Take snapshot. Add snapshot to storytelling -> Empty cells are rendered.

Fixed!

quanho:
Change to another PR https://github.com/qlik-oss/sn-table/pull/1022

PR #1016, opened 2023-05-30
fix: chart exploration should not be in the property panel when straight table is inside container

LiKang6688:
straight table that has chart exploration enabled and is placed inside a container should not show Chart exploration in the property panel


linhnguyen-qlik (approved):

haxxmaxx (approved):
No need to care about the test coverage

PR #1017, opened 2023-05-30
refactor: switch to useInteractionState

haxxmaxx:
Switching to `useInteractionState` since `useConstraints` has been depricated. Trying to use as few `!` as possible. Unfortunately, the type for the props is still optional so you have to do `!!` to get a boolean


mashup-example/using-build-tool/qlik-sense-desktop/configure.js
```diff
@@ -25,15 +25,6 @@ const nuked = embed.createConfiguration({
     theme: 'myTheme',
     language: 'en-US',
     keyboardNavigation: true,
-    constraints: {
```
haxxmaxx, line 28: These were the same as the default values, so don't think it need to be specified

mashup-example/using-build-tool/qlik-sense-saas-platform-sdk/configure.js
```diff
@@ -5,7 +5,7 @@ const n = embed.createConfiguration({
   context: {
     theme: 'light',
     language: 'en-US',
-    constraints: {
+    interactions: {
       active: false, // do not allow interactions
```
haxxmaxx, line 9: according to the comment this should have been set to true. With inverting the booleans it should actually be correct.

src/table/pagination-table/components/body/TableBodyWrapper.tsx
```diff
@@ -22,17 +22,15 @@ function TableBodyWrapper({ setShouldRefocus, tableWrapperRef, announce }: Table
...
     },
   } = useContextSelector(TableContext, (value) => value.baseProps);
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
   const selectionDispatch = useContextSelector(TableContext, (value) => value.selectionDispatch);
-  // constraints.active: true - turn off interactions that affect the state of the visual
-  // representation including selection, zoom, scroll, etc.
-  // constraints.select: true - turn off selections.
-  const isSelectionsEnabled = !constraints.active && !constraints.select;
+  // Both active and select conditions need to be true to make selections. See stardust API for more info
+  const isSelectionsEnabled = !!interactions.active && !!interactions.select;
```
haxxmaxx, line 33: I think we shouldn't describe how a documented API works, where it is consumed.

PR #1018, opened 2023-05-31
perf: throttle page fetch

cbt1:
This PR changes two things:
1. qDataPage fetches are throttled by 75 ms, instead of "as fast as possible".
2. Pages in are merged so that the GetHyperDataCalls to Engine contains less "pages" with qWidth and qHeight === 1

The intention is to reduce the burden on Engine and in that way also improve the performance. Some smaller tables with a simple data structure are likely getting a small performances hit because of the throttles fetches.


haxxmaxx (approved):
Nice, never really understood queueMicrotask anyway

quanho (approved):

PR #1019, opened 2023-06-01
perf: stale page check

cbt1:
When the user scrolls, a scroll event listener starts a process to check if cells that are scrolled into view are missing data. When this is the case a request to fetch data for those missing cells is sent to Engine. But sometimes the user scrolls rather quickly, so quickly that some of the cells that where empty and needed data, are no longer in the visible view of the user.

This PR attempts to optimize that process, so that when a cell is no longer visible to the user and the data has not already been fetch. It's considered stale and not included in the next data fetch.


src/table/virtualized-table/utils/merge-pages.ts
```diff
@@ -1,31 +1,69 @@
...
+  prevPage.qHeight === qPage.qHeight;
+
+const isPageStale = (gridState: React.MutableRefObject<GridState>, pageInfo: PageInfo, qPage: EngineAPI.INxPage) => {
+  const pageRowStart = pageInfo.page * pageInfo.rowsPerPage;
+  const colStartIndex = gridState.current.overscanColumnStartIndex - COLUMN_DATA_BUFFER_SIZE;
+  const colEndIndex = gridState.current.overscanColumnStopIndex + COLUMN_DATA_BUFFER_SIZE;
+  const rowStartIndex = pageRowStart + gridState.current.overscanRowStartIndex - ROW_DATA_BUFFER_SIZE;
+  const rowEndIndex = pageRowStart + gridState.current.overscanRowStopIndex + ROW_DATA_BUFFER_SIZE;
+
+  // qWidth and qHeight is assumed to always be 1. So that is ignored in this check.
```
cbt1, line 24: Fix: only qHeight is 1

src/table/virtualized-table/utils/merge-pages.ts
```diff
@@ -1,31 +1,67 @@
...
-const canMergeColumns = (prevPage: EngineAPI.INxPage, page: EngineAPI.INxPage) =>
-  prevPage.qLeft + prevPage.qWidth === page.qLeft && prevPage.qTop === page.qTop && prevPage.qHeight === page.qHeight;
+const isPageStale = (gridState: React.MutableRefObject<GridState>, pageInfo: PageInfo, qPage: EngineAPI.INxPage) => {
+  const pageRowStart = pageInfo.page * pageInfo.rowsPerPage;
+  const colStartIndex = gridState.current.overscanColumnStartIndex - COLUMN_DATA_BUFFER_SIZE;
+  const colEndIndex = gridState.current.overscanColumnStopIndex + COLUMN_DATA_BUFFER_SIZE;
+  const rowStartIndex = pageRowStart + gridState.current.overscanRowStartIndex - ROW_DATA_BUFFER_SIZE;
+  const rowEndIndex = pageRowStart + gridState.current.overscanRowStopIndex + ROW_DATA_BUFFER_SIZE;
+
+  return (
```
cbt1, line 12: I think this should be improved. There are really two points here `{ x: qPage.qLeft, y: qPage.qTop }` and `{ x: qPage.qLeft + qPage.qWidth, y: qPage.qTop + qPage.qTop + qPage.qHeight }`. If any of those points are inside the "rectangle" that is the rows and columns that are visible to the user. The page should not be considered stale. This is currently not how this logic work.

src/table/virtualized-table/utils/merge-pages.ts
```diff
@@ -1,31 +1,70 @@
...
+  const { qLeft, qTop, qWidth, qHeight } = qPage;
+
+  // The page is considered stale if both "points" from by the qPage, are outside the rectangle
+  // formed by the gridState + buffer.
+  return !(
+    qLeft >= colStartIndex &&
+    qLeft <= colEndIndex &&
+    qTop >= rowStartIndex &&
+    qTop <= rowEndIndex &&
+    qLeft + qWidth >= colStartIndex &&
```
cbt1, line 20: Note to self: Should be an OR not an AND operator between the points

haxxmaxx:
Looking through my review requests and found this. Should I take a look?

cbt1:
> Looking through my review requests and found this. Should I take a look?

I think you can wait. It will take a look at the "note to self" when I get the time. It should still be valid.

cbt1:
Closing, lets do something similar to PVT where fetching is throttled/debounced and uses two buffers

PR #1024, opened 2023-06-05
test: optimize rendering tests

haxxmaxx:
A rework of https://github.com/qlik-oss/sn-table/pull/1011/files to reduce the number of fixture. More or less covering the same amount of features, except some hover cases. I also rewrote the documentation a bit, trying to clarify it a bit.

Next I will add two more browsers, and a couple more tests with styling coming from the theme.


niekvanstaveren (approved):
Looks okay I think. We can probably do some more around this. 
Would suggest moving the fixtures into the default folder.

haxxmaxx:
> Looks okay I think. We can probably do some more around this. Would suggest moving the fixtures into the default folder.

What do you mean by do some more? Reduce the number of fixtures more? I listed some more things I will do next after this PR

niekvanstaveren:
> > Looks okay I think. We can probably do some more around this. Would suggest moving the fixtures into the default folder.
> 
> What do you mean by do some more? Reduce the number of fixtures more? I listed some more things I will do next after this PR

That indeed 😄

PR #1026, opened 2023-06-07
refactor: add feature flag for download

quanho:


src/ext/index.js
```diff
@@ -26,9 +26,9 @@ export default function ext(env) {
     ...getExploration(env),
     data: getData(env),
     support: {
-      export: true,
+      export: env.flags.isEnabled('PS-20907_TABLE_DOWNLOAD'),
```
haxxmaxx, line 29: could we follow the same pattern as we used in this repo before? Then this would be `PS_20907_SN_TABLE_DOWNLOAD`

quanho: I could only find PS_18291_TABLE_EXPLORATION so I use the same parttern.

haxxmaxx (approved):

PR #1028, opened 2023-06-20
feat: show a snapshot with the exact scroll position

Donya-qlik:
**The issue:**
The current sn-table snapshot calculations are not accurate. If the table row is half-visible, it will take it as a totally visible row and if the row is less than 50% visible, it will not consider it as part of the snapshot.

**Solution for pagination mode**
The improvement calculates the exact top row visible height (if there is any) and adjusts the final table scrollTop based on the new calculations to reach to the exact visible row height. The result is a snapshot that fulfills `what you get is what you see`

image
image

### DIFF
--- a/src/table/pagination-table/components/head/TableHeadWrapper.tsx
+++ b/src/table/pagination-table/components/head/TableHeadWrapper.tsx
@@ -28,7 +28,7 @@ function TableHeadWrapper() {
 
   useEffect(() => {
     if (headRowRef.current) {
-      setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
+      setHeadRowHeight(headRowRef.current.clientHeight);
     }
   }, [styling.head.fontSize, columnWidths, setHeadRowHeight]);
 
--- a/src/table/utils/find-visible-rows.ts
+++ b/src/table/utils/find-visible-rows.ts
@@ -2,24 +2,38 @@ import { TotalsPosition, ViewService } from '../../types';
 
 const EPSILON = 0.001;
 
-const findStartIndex = (rows: HTMLCollectionOf<Element>, min: number, max: number): number => {
+export const getPartialTopScrollHeight = (
+  rows: HTMLCollectionOf<Element> | undefined,
+  tableBodyRect: { top: number; bottom: number },
+  index: number
+): number => {
+  if (!rows || rows.length === 0 || index < 0) return 0;
+  const rowRect = rows[index].getBoundingClientRect();
+  if (rowRect.top === tableBodyRect.top) return 0;
+  const partialHeight = tableBodyRect.top - rowRect.top;
+  return partialHeight > 0 ? partialHeight : 0;
+};
+
+const findStartIndex = (
+  rows: HTMLCollectionOf<Element> | undefined,
+  tableBodyRect: { top: number; bottom: number }
+): number => {
+  if (!rows) return -1;
   for (let i = 0; i < rows.length; i++) {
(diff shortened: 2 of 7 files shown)


src/table/utils/find-visible-rows.ts
```diff
@@ -2,24 +2,38 @@ import { TotalsPosition, ViewService } from '../../types';
 
 const EPSILON = 0.001;
 
-const findStartIndex = (rows: HTMLCollectionOf<Element>, min: number, max: number): number => {
+export const getPartialTopScrollHeight = (
```
cbt1, line 5: I think you can simplify this into:

```js
export const getPartialTopScrollHeight = (
  rows: HTMLCollectionOf<Element> | undefined,
  tableBodyRect: { top: number; bottom: number },
  index: number
): number => {
  if (!rows || rows.length === 0 || index < 0) return 0;
  const rowRect = rows[index].getBoundingClientRect();
  return Math.max(tableBodyRect.top - rowRect.top, 0);
};
```

src/table/utils/find-visible-rows.ts
```diff
@@ -64,13 +84,14 @@ export function findVirtualizedVisibleRows(rootElement: HTMLElement, viewService
...
   const visibleBottomInPage = visibleTopInPage + visibleHeight - 1;
-  const topLeftCell = getFristCellOfRow(visibleTopInPage, cells);
+  const topLeftCell = getFirstCellOfRow(visibleTopInPage, cells);
   const topLeftCellRect = topLeftCell?.getBoundingClientRect();
   if (!topLeftCell || !topLeftCellRect) return {};
   const bodyYMin = bodyRect.y;
   if (1 - scrollTopRatio < EPSILON) {
     // The last data row is visible, then the priority start from the bottom row
     const shouldIncludeTopRow = topLeftCellRect.y >= bodyYMin - 4;
+    console.log('value: ', visibleTopInPage + (shouldIncludeTopRow ? 0 : 1) + offset);
```
cbt1, line 94: Remove console.log

src/table/utils/find-visible-rows.ts
Donya-qlik: Yes! Exactly! Thanks

src/table/pagination-table/components/head/TableHeadWrapper.tsx
```diff
@@ -28,7 +28,7 @@ function TableHeadWrapper() {
 
   useEffect(() => {
     if (headRowRef.current) {
-      setHeadRowHeight(headRowRef.current.getBoundingClientRect().height);
+      setHeadRowHeight(headRowRef.current.clientHeight);
```
cbt1, line 31: Why change to `clientHeight`? Seems like the `getBoundingClientRect` returns a more suitable value as you get a fractional value. Which may fit better with "WYSIWYG"

Donya-qlik: Actually I got a bit help from Max here. 
The issue happened when Totals row was set. If I remember it correctly, getBoundingClientRect returns rendering sizes which, beside having a Totals row, a kind of size ratio miscalculation occurred when I added scrollTop adjustment So the adjustment was applied on Totals row as the first visible row! Also resizing the table when snapshot was used ended a big gap between header and Totals rows. Using clientHeight solved all these.

cbt1: Ok then I won't look into that anymore 👍

cbt1 (approved):
LGTM

PR #1032, opened 2023-06-22
fix: fix chart monitoring

quanho:
For sn-table chart monitoring, there is a scenario where we have a virtualized table and we capture a snapshot of it on "Monitor hub" action, but then printing service tries to render this snapshot data (which we use pagination table to render it) and capture another snapshot after the pagination table is rendered. In this case we don't need to find visible rows anymore, since they are already stored in viewState.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -60,7 +60,9 @@ export default function supernova(env: Galaxy) {
 
       extendContextMenu();
 
-      useSnapshot({ layout, viewService, model, rootElement, contentRect });
+      if (env.flags.isEnabled('PS_20907_TABLE_DOWNLOAD')) {
+        useSnapshot({ layout, viewService, model, rootElement, contentRect });
+      }
 
       useVirtualizedTable({
         app,


Caele:
You need to make sure you capture the case I used as well, where the viewstate isn't set but interaction is disabled, which I think means the table reverts to a paginated setup but without a viewstate.

quanho:
> You need to make sure you capture the case I used as well, where the viewstate isn't set but interaction is disabled, which I think means the table reverts to a paginated setup but without a viewstate.

We only use pagination table if either the table is a pagination table or layout.snapshotData || viewService.viewState?.visibleHeight is truthy. So you mean the case where viewService.viewState is falsy but therelayout.snapshotData is truthy?

src/types.ts
```diff
@@ -235,6 +235,8 @@ export interface ViewState {
   scrollTopRatio?: number;
   visibleTop: number;
   visibleHeight: number;
+  visibleLeft?: number;
+  visibleWidth?: number;
```
haxxmaxx, line 239: how come these are added in this PR?

haxxmaxx (approved):
LGTM. I just merged new rendering test snapshots, so you shouldn't have to update the snapshots in this PR

PR #1035, opened 2023-07-04
feat: VIZ-2624-set-size-to-table-snapshot

sharleneaqlik:
A table can be seen as 5 components:
1) the tilte
2) the footer
3) the data area
4) the vertical scroll
5) the horizontal scroll
The main area is the table but not include the title

Currently we use the main area to render data area in snapshot or download mode.
Expected: We should deduct the footer and scrolls from the main area when render data area in snapshot mode. Both the Pagination and Virtual Table should show content like below. 

**Decision**: Develop the feature by filling up with more rows and remove the footer when total number of rows displayed on the screen is more than 100. In the case of rows 10, 25 and 100 show the exact rows if available. 

image

### DIFF
--- a/src/table/hooks/use-snapshot.ts
+++ b/src/table/hooks/use-snapshot.ts
@@ -3,6 +3,7 @@ import { onTakeSnapshot, useImperativeHandle, stardust } from '@nebula.js/stardu
 import type { TableLayout, ViewService, SnapshotLayout, HyperCube } from '../../types';
 import { findPaginationVisibleRows, findVirtualizedVisibleRows } from '../utils/find-visible-rows';
 import { getTotalPosition } from '../../handle-data';
+import { initialPageInfo } from '../../nebula-hooks/use-pagination-table';
 
 interface UseSnapshotProps {
   layout: TableLayout;
@@ -21,10 +22,21 @@ const useSnapshot = ({ layout, viewService, model, rootElement, contentRect }: U
         rootElement,
         totalsPosition
       );
+
+      const totalRowCount = layout.qHyperCube.qSize.qcy;
+      const paginationNeeded = totalRowCount > 10;
+      const rowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
+      const rowsRendered = visibleRowEndIndex - visibleRowStartIndex + 1;
+      const getVisibleHeight = () => {
+        return paginationNeeded && rowsRendered < rowsPerPage
+          ? visibleRowEndIndex - visibleRowStartIndex + 4 // Temp:is considered to avoid ratio change to remove empty space at the bottom
+          : rowsRendered;
+      };
+
       return {
         scrollLeft: viewService.scrollLeft,
         visibleTop: viewService.qTop + visibleRowStartIndex,
-        visibleHeight: visibleRowEndIndex < 0 ? 0 : visibleRowEndIndex - visibleRowStartIndex + 1,
+        visibleHeight: visibleRowEndIndex < 0 ? 0 : getVisibleHeight(),
         rowsPerPage: viewService.rowsPerPage,
         page: viewService.page,
       };
--- a/src/nebula-hooks/use-pagination-table.ts
+++ b/src/nebula-hooks/use-pagination-table.ts
@@ -36,7 +36,7 @@ interface UsePaginationTable {
   viewService: ViewService;
 }
 
-const initialPageInfo = {
+export const initialPageInfo = {
   page: 0,
   rowsPerPage: 100,
   rowsPerPageOptions: [10, 25, 100],


src/table/hooks/use-snapshot.ts
```diff
@@ -13,6 +14,12 @@ interface UseSnapshotProps {
 }
 
 const useSnapshot = ({ layout, viewService, model, rootElement, contentRect }: UseSnapshotProps) => {
+  const totalRowCount = layout.qHyperCube.qSize.qcy;
+  const visualRowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
+  const getVisibleHeight = (visibleRowEndIndex: number, visibleRowStartIndex: number) => {
+    return Math.min(totalRowCount, visualRowsPerPage, visibleRowEndIndex - visibleRowStartIndex + 4);
+  };
```
haxxmaxx, line 21: no real point of doing this as a function, since you will always use a `visibleHeight` value. I'd suggest something like this, baking in the `visibleRowEndIndex < 0`
```suggestion
  let visibleHeight = 0
  if (visibleRowEndIndex >= 0) {
    const totalRowCount = layout.qHyperCube.qSize.qcy;
    const rowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
    // Adding three extra rows to make sure will fill the space from the pagination
    const extendedVisibleHeight = visibleRowEndIndex - visibleRowStartIndex + 4;
    visibleHeight = Math.min(totalRowCount, visualRowsPerPage, extendedVisibleHeight)
  }
  }
```

sharleneaqlik: Fixed using [this](https://github.com/qlik-oss/sn-table/pull/1035/commits/35a0a95f0285cc1df753900875b1077f1df07dc1).

src/table/hooks/use-snapshot.ts
```diff
@@ -21,10 +28,11 @@ const useSnapshot = ({ layout, viewService, model, rootElement, contentRect }: U
         rootElement,
         totalsPosition
       );
+
       return {
         scrollLeft: viewService.scrollLeft,
         visibleTop: viewService.qTop + visibleRowStartIndex,
-        visibleHeight: visibleRowEndIndex < 0 ? 0 : visibleRowEndIndex - visibleRowStartIndex + 1,
+        visibleHeight: visibleRowEndIndex < 0 ? 0 : getVisibleHeight(visibleRowEndIndex, visibleRowStartIndex),
```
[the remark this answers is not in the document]
sharleneaqlik: Fixed using [this](https://github.com/qlik-oss/sn-table/pull/1035/commits/35a0a95f0285cc1df753900875b1077f1df07dc1).

sharleneaqlik:
@haxxmaxx , you can do the final review since this your last day today. I can add the unit tests separately.

sharleneaqlik:
@haxxmaxx I have added three test cases where the lowest number of rows should be picked up. `ViewState` function's test case can be done separately in a different PR.

src/table/hooks/__tests__/use-snapshot.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+describe('use-snapshot', () => {
+  let visibleRowEndIndex: number;
+  let visibleRowStartIndex: number;
+  let layout: TableLayout;
+  let viewService: ViewService;
+  const dimensionCount = 2;
+  const measureCount = 1;
+  let rowCount = 100;
+
+  it('should return the minimum value for the row number', async () => {
```
haxxmaxx, line 14: These descriptions are a bit vague and I'm not sure you are testing all three cases properly. 

1. test if the value coming from `visibleRowEndIndex` and `visibleRowStartIndex` is returned. Make sure it is different than rowsPerPage and rowCount (rowCount becomes qcy)
1. test if `rowsPerPage` is used, again should differ from the other two
1. test if `qcy` is used, which is the third argument to generate layout

src/table/hooks/__tests__/use-snapshot.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { generateLayout } from '../../../__test__/generate-test-data';
+import { TableLayout, ViewService } from '../../../types';
+import { getVisibleHeight } from '../use-snapshot';
+
+describe('use-snapshot', () => {
```
haxxmaxx, line 5: You also need to test the early return, when visibleRowEndIndex < 0

src/table/hooks/use-snapshot.ts
```diff
@@ -13,6 +14,14 @@ interface UseSnapshotProps {
 }
 
 const useSnapshot = ({ layout, viewService, model, rootElement, contentRect }: UseSnapshotProps) => {
+  const getVisibleHeight = (visibleRowEndIndex: number, visibleRowStartIndex: number) => {
+    if (visibleRowEndIndex < 0) return 0;
+
+    const totalRowCount = layout.qHyperCube.qSize.qcy;
+    const visualRowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
+    return Math.min(totalRowCount, visualRowsPerPage, visibleRowEndIndex - visibleRowStartIndex + 4);
```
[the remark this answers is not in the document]
sharleneaqlik: fixed

src/table/hooks/__tests__/use-snapshot.spec.ts
sharleneaqlik: test case added.

src/table/hooks/use-snapshot.ts
```diff
@@ -12,37 +13,54 @@ interface UseSnapshotProps {
...
+  if (visibleRowEndIndex < 0) return 0;
+
+  const totalRowCount = layout.qHyperCube.qSize.qcy;
+  const visualRowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
+  // EXTRA_ROWS will be added to the visualHeight when the pagination footer is displayed and rows per page is heighr
+  return Math.min(totalRowCount, visualRowsPerPage, visibleRowEndIndex - visibleRowStartIndex + 1 + EXTRA_ROWS);
+};
+
+export const getViewState = (layout: TableLayout, viewService: ViewService, rootElement: HTMLElement) => {
+  if (viewService.viewState) return viewService.viewState;
```
Donya-qlik, line 33: Do we have the case that `viewState` does not exist ? If so we should set the default values instead.
The condition doesn't sound proper! you should define `else` to handle the undefined case otherwise` it won't make sense. It could be something like this:
```suggestion
  viewService.viewState ?? { scrollLeft: 0, visibleLeft: 0 .... and others with their default values};
```

Donya-qlik: I'm actually thinking the function is there to define the `viewState` as it returns it's properties. so I think if you want to have such a sanity check you should start it with `if(!vieService.viewState) { the rest of the code }` otherwise take the default values.

src/table/hooks/use-snapshot.ts
```diff
@@ -12,37 +13,54 @@ interface UseSnapshotProps {
...
+    visibleTop: visibleRowStartIndex,
+    visibleHeight: getVisibleHeight(visibleRowEndIndex, visibleRowStartIndex, layout, viewService),
+    scrollTopRatio: viewService.scrollTopRatio,
+    rowsPerPage: viewService.rowsPerPage,
+    page: viewService.page,
   };
+};
 
+const useSnapshot = ({ layout, viewService, model, rootElement, contentRect }: UseSnapshotProps) => {
   onTakeSnapshot(async (snapshotLayout: SnapshotLayout) => {
```
Donya-qlik, line 64: Is there a reason you haven't moved this function out of here?

src/table/hooks/use-snapshot.ts
```diff
@@ -12,37 +13,54 @@ interface UseSnapshotProps {
...
+  const visualRowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
+  // EXTRA_ROWS will be added to the visualHeight when the pagination footer is displayed and rows per page is heighr
+  return Math.min(totalRowCount, visualRowsPerPage, visibleRowEndIndex - visibleRowStartIndex + 1 + EXTRA_ROWS);
+};
+
+export const getViewState = (layout: TableLayout, viewService: ViewService, rootElement: HTMLElement) => {
+  if (viewService.viewState) return viewService.viewState;
+  if (layout.usePagination) {
+    const totalsPosition = getTotalPosition(layout);
+    const { visibleRowStartIndex = -1, visibleRowEndIndex = -1 } = findPaginationVisibleRows(
```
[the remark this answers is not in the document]
Donya-qlik: **Update**: I tried and it didn't work! it seems it takes them as new variables and destructuring them even with the same name doesn't change the concept

src/table/hooks/use-snapshot.ts
haxxmaxx: A good rule of thumb is to early-return early in the code. If viewState exists, just return it and don't bother with the rest of the code. A bit easier to read. Also if you do
```
if (!viewService.viewState) {
  ...
  ...
}
```
the code gets an extra level of indentation which you don't really need.

src/table/hooks/use-snapshot.ts
```diff
@@ -79,6 +99,7 @@ const useSnapshot = ({ layout, viewService, model, rootElement, contentRect }: U
     }
     return snapshotLayout;
   });
+
   useImperativeHandle(
     () => ({
       getViewState,
```
quanho, line 105: getViewState: () => getViewState(layout, viewService, rootElement)

PR #1047, opened 2023-08-07
fix: introduce alternate state in listbox popup

niekvanstaveren:


src/table/components/head/HeadCellMenu.tsx
```diff
@@ -149,10 +152,9 @@ export default function HeadCellMenu({ column, tabIndex }: HeadCellMenuProps) {
     [translator, isDim, fieldInstance, selectionActionsEnabledStatus, embedListbox]
   );
 
-  const handleOpenDropdown = async () => {
+  const handleOpenDropdown = () => {
     if (!openMenuDropdown && model) {
-      const layout = await model.getLayout();
```
a-m-dev, line 154: I think there were a reason why we specifically wanted to fetch layout before calling `updateSelectionActionsEnabledStatus` because the latest layout had some more information that was stale in `layout` object from `useContextSelector` in `use-field-selection` module -> when we tried to use it. (as far as i remember, selection menu items were not synched with the most updated layout state from engine and that is why we needed to fetch layout here and pass it to being used)

niekvanstaveren: I can revert the changes for updateSelectionActionsEnabledStatus just to be on the safe side.

a-m-dev: So after trying to mount my build of sn-table into QMFE, here is the result:

https://github.com/qlik-oss/sn-table/assets/29652890/ac64f7cc-ebd4-4848-bcba-bdba1a8c8106

what i did was: 
1. upload app from ticket into our tenant
2. build SNT (with your changes and my console.log to have a valid check that table contains the changes)
3. copy extension into `sense-client/node_modules/..../viz-bundle/sn-table` folder
4. restart my local sense-client to pick up with latest SNT changes

and then, still going to selections -> select all on mentioned master dim from ticket is not performing selection action. 
Is that correct or maybe I'm missing some steps to validate your changes?

src/table/hooks/use-field-selection.ts
```diff
@@ -22,23 +22,23 @@ export const checkStateCountByKey = <T>(keys: (keyof T)[], obj: T): boolean => {
...
 const useFieldSelection = (column: Column): UseFieldSelectionOutput => {
-  const { app } = useContextSelector(TableContext, (value) => value.baseProps);
+  const { app, layout } = useContextSelector(TableContext, (value) => value.baseProps);
   const [fieldInstance, setFieldInstance] = useState<EngineAPI.IField | null>(null);
   const [selectionActionsEnabledStatus, setSelectionActionsEnabledStatus] = useState(
     SELECTION_ACTIONS_ENABLED_DEFAULT_STATUS
   );
 
   useEffect(() => {
     if (!app || !app.getField || !column || !column.isDim) return;
```
a-m-dev, line 32: Maybe a check for layout here?

a-m-dev (approved):
LGTM, 
Just a minor concern about layout check

src/table/hooks/use-field-selection.ts
niekvanstaveren: How would this even render without a layout? I guess the same can be said for column 🤷

PR #1049, opened 2023-08-08
fix: make selection in master dimension

niekvanstaveren:


niekvanstaveren:
Will fix the unit tests, but would like to check this solution as well.

src/table/hooks/use-field-selection.ts
```diff
@@ -30,7 +40,18 @@ const useFieldSelection = (column: Column): UseFieldSelectionOutput => {
 
   useEffect(() => {
     if (!app || !app.getField || !column || !column.isDim) return;
-    app.getField(column.fieldId, layout.qStateName).then(setFieldInstance);
+    const { qLibraryId, fieldId } = column;
+    qLibraryId
```
cbt1, line 44: This would probably be easier to read if it was a "if else statement" rather then a ternary operator

src/table/hooks/use-field-selection.ts
```diff
@@ -30,7 +40,18 @@ const useFieldSelection = (column: Column): UseFieldSelectionOutput => {
...
+      ? app.createSessionObject(getListBoxSessionObject(qLibraryId, layout.qStateName)).then((listboxSessionObject) =>
+          setFieldInstance({
+            selectAll: () => listboxSessionObject.selectListObjectAll('/qListObjectDef'),
```
a-m-dev, line 47: Just a random thought, can we use model here instead of creating a session object or they are totally used for different purposes? [(This here from nebula is why i ask this question)](https://github.com/qlik-oss/nebula.js/blob/280f587b60698bd76919a0b72a249c7e747a1749/apis/nucleus/src/components/listbox/interactions/listbox-selection-toolbar.js#L22-L32)

Plus, we have direct access to model in line 35 through `useContextSelector(...)` call

a-m-dev: So you mean the model that we have access within table ([which comes form using `useModel` hook](https://github.com/qlik-oss/sn-table/blob/bbedb322dd622ba725d72d688c4866e964a26d40/src/index.ts#L43C14-L43C14)) and the model that nebula uses (useSessionModel -> [which ends up in here](https://github.com/qlik-oss/nebula.js/blob/280f587b60698bd76919a0b72a249c7e747a1749/apis/nucleus/src/hooks/useSessionModel.js#L23)) are different? 

The confusing part for me is they both are model, but how they (may) act differently in this case that would make us to create a new session object from scratch (for master dims -> which is distinguishable from a normal dim by `qLibraryId`)?

niekvanstaveren: The model that comes from the useModel hook is the model for the table. It is completely different from the model that is returned for a single field (or in this case the on the fly session object for a listbox)

src/table/hooks/use-model-store.ts
```diff
@@ -0,0 +1,17 @@
+const keyStore = (initialState: { [key: string]: EngineAPI.IField } = {}) => {
```
cbt1, line 1: I don't think this is a hook as it does not have to obey the rules of hooks

a-m-dev (approved):
LGTM, 
I tried it with both normal and Drill down master dim and it works fine.

src/table/hooks/use-model-store.ts
niekvanstaveren: Indeed

src/table/utils/model-store.ts
```diff
@@ -0,0 +1,17 @@
+const keyStore = (initialState: { [key: string]: EngineAPI.IField } = {}) => {
```
cbt1, line 1: I wonder if this can not be re-written to simply:
```js
const modelStore = new Map<string, EngineAPI.IField>();

export default modelStore;

```

src/table/utils/model-store.ts
```diff
@@ -0,0 +1,17 @@
...
+      sharedState[key] = value;
+    },
+  };
+
+  return store;
+};
+
+const modelStore = keyStore({});
+
+export default modelStore;
```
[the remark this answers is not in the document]
niekvanstaveren: I can test that. Ideally it would I think, that way we create only a minimal set of objects.
Think however that every object has its own object store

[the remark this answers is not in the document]
cbt1: Yeah I would agree with that, I think it does. Just not sure.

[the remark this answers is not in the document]
cbt1: One thing that comes to mind is that it might "never" be clean-up. If a user adds a field and then remove it. That field will remain in the store.

Another thing is that I wonder what happens when a user is on a sheet with a sn-table, moves to another sheet and than back to the sheet with the table again. Is that store than re-created?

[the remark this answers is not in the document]
niekvanstaveren: I'll have to do some testing around this. I assume the store gets recreated every time. 
Only way around that would probably be if we can have a store somewhere else (re-use the nebula setup?). Also not sure how much effort we should put in to that. Let me get @Caele take on this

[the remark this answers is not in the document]
cbt1: My guess is that it's fine. The same approach is used in picasso.js to store result of measured text.

If you're uncertain about this, it's possible create a singleton per component instance in the supernova definition, using the useMemo hook.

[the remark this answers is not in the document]
niekvanstaveren: Okay so it seems that the store is shared for all tables on the sheet. The store even not re-created when moving away from the sheet. I guess it is created when the table code is loaded for the first time.

cbt1 (approved):

PR #1050, opened 2023-08-11
chore: use qlik eslint configuration

niekvanstaveren:


src/handle-data.ts
```diff
@@ -95,9 +95,7 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
   const { qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIdx < numDims;
-  const info = (isDim ? qDimensionInfo[colIdx] : qMeasureInfo[colIdx - numDims]) as
-    | ExtendedNxMeasureInfo
-    | ExtendedNxDimensionInfo;
```
Donya-qlik, line 100: Are they supposed to be removed ?

src/table/components/head/__tests__/HeadCellMenu.spec.tsx
```diff
@@ -264,36 +264,41 @@ describe('<HeadCellMenu />', () => {
       fireEvent.click(screen.getByText(`SNTable.MenuItem.${targetAction}`));
     };
 
-    afterEach(async () => {
```
Donya-qlik, line 267: what was wrong with this afterEach ?

src/handle-data.ts
niekvanstaveren: The typecasting was apparently not needed.

src/table/components/head/__tests__/HeadCellMenu.spec.tsx
niekvanstaveren: You cannot have assertions outside of test cases

Donya-qlik (approved):

jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   testEnvironment: 'jest-environment-jsdom',
   testRegex: 'src/.+\\.(test|spec)\\.[jt]sx?$',
-  setupFilesAfterEnv: ['./jest.setup.js'],
+  setupFilesAfterEnv: ['./jest.setup.spec.js'],
```
cbt1, line 4: Why was the name change required?

package.json
```diff
@@ -79,20 +119,14 @@
...
-    "@mui/material": "5.13.2",
-    "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest",
     "@nebula.js/cli": "4.1.0",
     "@nebula.js/cli-build": "4.1.0",
     "@nebula.js/cli-sense": "4.1.0",
     "@nebula.js/cli-serve": "4.1.0",
     "@playwright/test": "1.34.3",
     "@qlik-oss/nebula-table-utils": "1.1.0",
     "@qlik-trial/sprout": "2.45.6",
+    "@qlik/eslint-config": "^0.4.12",
```
cbt1, line 129: Should pin the version

package.json
```diff
@@ -105,16 +139,7 @@
...
-    "eslint-config-airbnb-typescript": "17.0.0",
-    "eslint-config-prettier": "8.8.0",
-    "eslint-plugin-import": "2.27.5",
-    "eslint-plugin-jest": "27.2.1",
-    "eslint-plugin-jsx-a11y": "6.7.1",
-    "eslint-plugin-prettier": "4.2.1",
-    "eslint-plugin-react": "7.32.2",
-    "eslint-plugin-react-hooks": "4.6.0",
-    "eslint-plugin-testing-library": "5.11.0",
+    "eslint-plugin-jest": "^27.2.3",
```
cbt1, line 142: Should pin the version

package.json
```diff
@@ -71,6 +71,46 @@
     ],
     "yarn.lock": "yarn deduplicate-packages --list --fail"
   },
+  "eslintConfig": {
+    "extends": [
```
cbt1, line 75: Do we not need the base config `@qlik/eslint-config`?

Also what about `@qlik/eslint-config/playwright`?

package.json
```diff
@@ -71,6 +71,46 @@
...
+          "allowTernary": true
+        }
+      ],
+      "@typescript-eslint/unbound-method": 0,
+      "@typescript-eslint/restrict-template-expressions": 0
+    },
+    "root": true
+  },
+  "dependencies": {
+    "@material-ui/core": "4.12.4",
```
cbt1, line 103: Why are these moved to `dependencies`?

jest.config.js
niekvanstaveren: The file contains jest specifics so it needs to be recognised as a test file (beforeEach and jest.spyOn)

package.json
niekvanstaveren: All packages also include the default @qlik/eslint-config. Should probably remove it from other repos instead. 
https://github.com/qlik-oss/dev-tools-js/blob/main/packages/eslint-config/react.js

Not so sure about playwright. I can check what happens when added

package.json
niekvanstaveren: Interesting. It was complaining before that these should be real dependencies, but not anymore it seems. I'll move them back

cbt1 (approved):
LGTM

PR #1051, opened 2023-08-11
fix: fix missing parameter

quanho:
To fix a mistake from https://github.com/qlik-oss/sn-table/pull/1035 which broke sn-table down load on qcs.


Donya-qlik (approved):

sharleneaqlik (approved):

sharleneaqlik:
You would need test coverage for the fix.

quanho:
> You would need test coverage for the fix.

Added!

PR #1055, opened 2023-08-15
feat: VIZ-2626-renders virtual table for snapshot

sharleneaqlik:
Currently Snapshot creation for Virtualized table uses pagination table to create a snapshot. This PR includes the feature implementation to change that. Now this will use Virtualized table for the virtualized table snapshot.   

This PR is only related to the usage of Virtualised Table for Virtualised Snapshot (**Not for snapshot resizing**). 

Ongoing tasks: 
1. Testing
2. Writing test cases

**_Before this PR is merged the following tests must be conducted and passed._** 

**Mandatory Manual Test**: 
Taking/Creating Snapshot(uses snapshotData):
1. Taking a snapshot and use the snapshot in storytelling. (test locally and on qcs )
3. Download on windows - this will use snapshot from the table. 
4. Download on qcs - this will use a capture of live table data

**How to test  Download on QCS:**
Scenario:
1. Locally: can not be tested since client is not the same on tenant. Right click -> Developer -> Test Live Mashup -> Object (this will use printing live) (this is to test if it works on qcs or not).
2. Test on tenant by creating an extension and upload it 

**How to test Download on Windows:** 
1. Right click -> Developer -> Test Exporting (this will run printing snapshot)

**Test free resize on Storytelling:**
1. Select the snapshot -> click the lock sign on the top right corner (lock & unlock)
2. Unlocking allows free re size which shows more rows, but locked option shows resizable table data and not more rows

Flags to use for testing:
!PS_18291_SN_TABLE_BASIC_FEATURES 
EXPORT_3RD_PARTY_EXTENSIONS_ENABLED


src/nebula-hooks/use-virtualized-table.ts
```diff
@@ -53,7 +53,7 @@ const useVirtualizedTable = ({
   viewService,
 }: UseVirtualizedTable) => {
   const isPrintingMode = isPrinting(layout, viewService);
-  const shouldRender = layout.usePagination === false && !isPrintingMode;
+  const shouldRender = layout.usePagination === false || isPrintingMode;
```
quanho, line 56: Should use only layout.usePagination to test:
const shouldRender = layout.usePagination === false;

sharleneaqlik: @quanho do you mean remove `isPrintingMode;` check completely?

cbt1:
Does this work when a user re-sizes a snapshot? Without scaling that, so that new data might be visible or a column migh become wider/smaller

sharleneaqlik:
I am testing it currently. Will update asap.  

> Does this work when a user re-sizes a snapshot? Without scaling that, so that new data might be visible or a column migh become wider/smaller

src/nebula-hooks/use-virtualized-table.ts
quanho: Yes, I used isPrintingMode to do the trick of use pagination table to render virtualized table snapshot.
Now if we always use virtualized table to render virtualized table snapshot and pagination table to render pagination table snapshot then we don't need to check if it is in printing mode or not.
You should remove checking isPrintingMode everywhere.

cbt1:
> @cbt1 When the snapshot has unlock selected then the snapshot can be scaled and all hidden columns are displayed, but not the rows. More rows are not displayed.
> 
> But when the snapshot is locked then the snapshot size gets bigger and no extra columns are displayed.
> 
> I can confirm that, this feature works exactly same as current table for snapshot lock unlock and resizing feature.
> 
> Are these expected behaviours?
> 
> > Does this work when a user re-sizes a snapshot? Without scaling that, so that new data might be visible or a column migh become wider/smaller

We you re-size so that more rows becomes visible they too should have data. To fix that I think you can simple store some more rows and columns in the snapshot data. It won't work for every single case but is likely good enough for most.

IMO you can fix that in another PR.

sharleneaqlik:
Will separate the resizing issue from this PR. SInce this PR is mainly about the usage of Virtualised table for Virtualised snapshot. 


> > @cbt1 When the snapshot has unlock selected then the snapshot can be scaled and all hidden columns are displayed, but not the rows. More rows are not displayed.
> > But when the snapshot is locked then the snapshot size gets bigger and no extra columns are displayed.
> > I can confirm that, this feature works exactly same as current table for snapshot lock unlock and resizing feature.
> > Are these expected behaviours?
> > > Does this work when a user re-sizes a snapshot? Without scaling that, so that new data might be visible or a column migh become wider/smaller
> 
> We you re-size so that more rows becomes visible they too should have data. To fix that I think you can simple store some more rows and columns in the snapshot data. It won't work for every single case but is likely good enough for most.
> 
> IMO you can fix that in another PR.

Donya-qlik (commented):
The code lgtm so I'll focus on testing them locally and will give a feedback soon

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -101,7 +105,7 @@ const useDynamicRowHeight = ({
         rowMeta.current.heights[rowIdx] = height;
       }
 
-      if (!batchStateUpdate) {
+      if (!batchStateUpdate && !isSnapshot) {
```
cbt1, line 108: I think you can move `!isSnapshot` to line 89. So that it will return early and not having to call `getCellSize` which is computationally heavy.

cbt1 (approved):
Have not done a deep dive into the code here but looks like it can work. Just had some minor remarks.

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
sharleneaqlik: Fixed.

src/table/virtualized-table/hooks/use-dynamic-row-height.ts
```diff
@@ -42,6 +44,8 @@ const useDynamicRowHeight = ({
   columns,
   boldText,
   gridState,
+  isSnapshot,
+  viewService,
```
quanho, line 48: We don't need to pass it here. We can get viewService through 
useContextSelector(TableContext, (value) => value.baseProps)

PR #1056, opened 2023-08-15
chore: use qlik prettier

niekvanstaveren:


cbt1:
Lets get this in before more and more changes have come in to `main`?

package.json
```diff
@@ -95,29 +84,29 @@
...
-      "@typescript-eslint/restrict-template-expressions": 0
+      "@typescript-eslint/restrict-template-expressions": 0,
+      "@typescript-eslint/unbound-method": 0
     },
     "root": true
   },
+  "resolutions": {
+    "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest"
+  },
   "devDependencies": {
```
cbt1, line 95: Update `prettier` to v3 as well since qlik prettier has `^3.x` as peer dep.

Donya-qlik (approved):

package.json
niekvanstaveren: Prettier v3 does not really play nice with the latest prettier-config because the trailing comma setting was removed here https://github.com/qlik-oss/dev-tools-js/commit/05a1a0664dc70f02d97d4edf87d3869e841b152d
Had some issues in another repo where it actually broke things.

cbt1: Good to know. We've been using it in the pivot table repo and that has so far been working fine

cbt1 (approved):

PR #1060, opened 2023-09-05
feat: support multi-page pdf by rendering rows as a single-line row

quanho:
The main idea is that:
First client side of the reporting service will send a request to ask sn-table to get number of visible rows for a pdf page. This can be done by using virtualized table (or may be pagination table) to render rows as a single-line row. After reporting service get the answer from the sn-table, it will generate an array of requests, each of which is to ask sn-table to render a number of rows (as single-line row) for a pdf page.


src/table/pagination-table/components/body/TableBodyWrapper.tsx
```diff
@@ -42,10 +43,11 @@ const TableBodyWrapper = ({ setShouldRefocus, tableWrapperRef, announce }: Table
   const hoverEffect = !!getStylingComponent(layout)?.content?.hoverEffect;
 
   useSelectionListener({ keyboard, selectionDispatch, selectionsAPI, setShouldRefocus, tableWrapperRef });
+  const showTotals = (totalsPosition.atTop || totalsPosition.atBottom)  && viewService.viewState?.showTotals !== false;
```
cbt1, line 46: Can you not integrate `viewService.viewState?.showTotals` in to the `getTotalPosition` function instead? Then you only need to resolve it once and both pagination table and virtual scroll table should get it

src/types.ts
```diff
@@ -244,6 +244,11 @@ export interface ViewState {
   visibleWidth?: number;
   rowsPerPage?: number;
   page?: number;
+  isMultiPage?: boolean;
+  maxLineCount?: number;
+  tableType?: number;
```
cbt1, line 249: Is `tableType` something temporary or why is that needed? Is the `usePagination` property on the layout not enough?

src/types.ts
```diff
@@ -244,6 +244,11 @@ export interface ViewState {
   visibleWidth?: number;
   rowsPerPage?: number;
   page?: number;
+  isMultiPage?: boolean;
+  maxLineCount?: number;
```
cbt1, line 248: Is the intention that the user will be able to set the number of lines when printing and/or taking a snapshot?

src/types.ts
quanho: The usePagination property on the layout not enough! Let say we have a pagination table but to estimate the number of rows in a pdf page we need to render it as a virtualized table (we use first data rows to render, so we don't need to fetch data since the data already exist). However, to render multi-page pdf, especially from the second page we have to use pagination table to do the job. Currently, using virtualized table to does not work for other pages than the first page.

src/types.ts
quanho: Yes, in future, when we support different row heights we will send probably 3 as maxLineCount instead of 1 for the first version of multi page pdf.

src/table/pagination-table/components/body/TableBodyWrapper.tsx
quanho: Good suggestion. I will look into this.

quanho: The idea is that we only show totals row on the first pdf page but not from the second pdf page.

src/types.ts
quanho: I just replaced tableType to usePagination to make it consistent with layout.usePagination.

src/handle-data.ts
```diff
@@ -27,7 +27,9 @@ export function getHighestPossibleRpp(width: number, rowsPerPageOptions: number[
 /**
  * Get the position of the totals
  */
-export function getTotalPosition(layout: TableLayout) {
+export function getTotalPosition(layout: TableLayout, viewService?: ViewService) {
+  // For multi-page pdf the totals row may not be needed from the second pdf page
+  if (viewService?.viewState?.skipTotals === true) return { atTop: false, atBottom: false };
```
haxxmaxx, line 32:
```suggestion
  if (viewService?.viewState?.skipTotals) return { atTop: false, atBottom: false };
```

src/render-as-pagination.ts
```diff
@@ -0,0 +1,11 @@
+import { TableLayout, ViewService } from './types';
+import isPrinting from './is-printing';
+
+const renderAsPagination = (layout: TableLayout, viewService: ViewService) => {
+  // For multi-page download, viewService.viewState?.usePagination will override layout.usePagination
+  if (viewService.viewState?.usePagination === true) return true;
+  if (viewService.viewState?.usePagination === false) return false;
```
haxxmaxx, line 7:
```suggestion
  if ((viewService.viewState?.usePagination !== undefined) return viewService.viewState.usePagination
```

src/nebula-hooks/use-virtualized-table.ts
```diff
@@ -53,9 +53,8 @@ const useVirtualizedTable = ({
   isFontLoaded,
   viewService,
 }: UseVirtualizedTable) => {
-  const isPrintingMode = isPrinting(layout, viewService);
-  const shouldRender = layout.usePagination === false && !isPrintingMode;
-  const tableData = useMemo(() => getVirtualScrollTableData(layout, interactions), [layout, interactions]);
+  const shouldRender = !renderAsPagination(layout, viewService);
```
haxxmaxx, line 56: Centralizing this makes sure you render one and one only pagination or virtualized table 👍

src/table/pagination-table/components/TableWrapper.tsx
```diff
@@ -133,7 +133,7 @@ const TableWrapper = (props: TableWrapperProps) => {
         data-testid="table-container"
       >
         <StyledTable styling={styling} showRightBorder={showRightBorder} stickyHeader aria-label={tableAriaLabel}>
-          <TableHeadWrapper />
+          {viewService.viewState?.skipHeader !== true ? <TableHeadWrapper /> : undefined}
```
haxxmaxx, line 136: Cleaner to early return null from `TableHeadWrapper` IMO

src/table/virtualized-table/Table.tsx
```diff
@@ -117,15 +117,15 @@ const Table = (props: TableProps) => {
     <ScrollableContainer ref={ref} width={tableRect.width} height={tableRect.height} onScroll={scrollHandler}>
       <FullSizeContainer width={containerWidth} height={containerHeight}>
         <StickyContainer rect={stickyContainerRect}>
-          <Header
+          {viewService.viewState?.skipHeader === true ? null : <Header
```
haxxmaxx, line 120:
```suggestion
          {viewService.viewState?.skipHeader? null : <Header
```

src/table/virtualized-table/Body.tsx
```diff
@@ -33,6 +33,7 @@ const Body = forwardRef<BodyRef, BodyProps>((props, ref) => {
   const { layout, model, theme, viewService } = useContextSelector(TableContext, (value) => value.baseProps);
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
   const initialDataPages = useContextSelector(TableContext, (value) => value.initialDataPages);
+  const maxNbrLines = viewService.viewState?.maxLineCount || undefined;
```
haxxmaxx, line 36: wont `maxNbrLines` be undefined without `|| undefined`, when `viewState` is undefined?

src/table/utils/get-view-state.ts
```diff
@@ -0,0 +1,41 @@
+import { findPaginationVisibleRows, findVirtualizedVisibleRows } from './find-visible-rows';
+import getVisibleHeight from './get-visible-height';
+import { getTotalPosition } from '../../handle-data';
+import renderAsPagination from '../../render-as-pagination';
+import { TableLayout, ViewService } from '../../types';
+
+const getViewState = (layout: TableLayout, viewService: ViewService, rootElement: HTMLElement) => {
```
haxxmaxx, line 7: Any changes to this function? hard to know when it has been moved

src/__test__/handle-data.spec.ts
```diff
@@ -267,9 +268,27 @@ describe('handle-data', () => {
...
     let expectedPosition: TotalsPosition;
-
+    let viewService: ViewService;
+    
     beforeEach(() => {
       expectedPosition = { atBottom: false, atTop: false };
+      viewService = { viewState: { skipTotals: true } as ViewState } as ViewService;
+    });
+
+    describe('When viewService?.viewState?.skipTotals is true', () => {
```
haxxmaxx, line 278: I would say you only need one test for this.

src/__test__/is-printing.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    layout.snapshotData = {} as SnapshotData;
+    expect(isPrinting(layout, viewService)).toEqual(true);
+  });
+
+  it('should return true if layout.snapshotData is falsy but viewService.viewState.visibleHeight is truthy', () => {
+    if(viewService.viewState) viewService.viewState.visibleHeight = 1;
+    expect(isPrinting(layout, viewService)).toEqual(true);
+  });
+
+  it('should return true if layout.snapshotData is falsy and viewService.viewState.visibleHeight is falsy', () => {
```
haxxmaxx, line 27:
```suggestion
  it('should return false if layout.snapshotData is falsy and viewService.viewState.visibleHeight is falsy', () => {
```

src/__test__/is-printing.spec.ts
```diff
@@ -0,0 +1,30 @@
+import isPrinting from '../is-printing';
+import { SnapshotData, TableLayout, ViewService, ViewState } from '../types';
+
+describe('is-printing', () => {
```
haxxmaxx, line 4: I would say you should move the `isPrinting` function + the test to `renderAsPagination`.

src/table/pagination-table/components/body/TableBodyWrapper.tsx
```diff
@@ -95,7 +96,7 @@ const TableBodyWrapper = ({ setShouldRefocus, tableWrapperRef, announce }: Table
                     handleMouseDownToFocusBody(cell, rootElement, setFocusedCellCoord, keyboard, totalsPosition)
                   }
                 >
-                  <CellText fontSize={cellStyle.fontSize}>{cell.qText}</CellText>
+                  <CellText fontSize={cellStyle.fontSize} lines={viewService.viewState?.maxLineCount || 3}>{cell.qText}</CellText>
```
haxxmaxx, line 99: `CellText` already have a default value of 3, no need to pass it

haxxmaxx:
> The main idea is that: First client side of the reporting service will send a request to ask sn-table to get number of visible rows for a pdf page. This can be done by using virtualized table (or may be pagination table) to render rows as a single-line row.

If you know the height of the PDF and the font size, you can calculate how many rows will fit, as long as you sticking to one line per cell. The paddings, border widths etc are all constant. And the line height is 4/3. So if you take the font size * 4/3 + padding + border height you get the height of a cell

quanho:
> > The main idea is that: First client side of the reporting service will send a request to ask sn-table to get number of visible rows for a pdf page. This can be done by using virtualized table (or may be pagination table) to render rows as a single-line row.
> 
> If you know the height of the PDF and the font size, you can calculate how many rows will fit, as long as you sticking to one line per cell. The paddings, border widths etc are all constant. And the line height is 4/3. So if you take the font size * 4/3 + padding + border height you get the height of a cell
> 
> EDIT: I guess if we want to be able to support more than one line later, I see the point of doing the estimation from the start

So you mean reporting service needs to send a request to sn-table to ask for font family, font size, padding, etc. to calculate. To avoid to ask many things like that and then do calculation, reporting service only need to ask for the row heights and it should don't care how to calculate the row height. Calculating the row height should be in sn-table.

src/table/utils/get-view-state.ts
quanho: There is a small change to this function. I move it to a separated file to make it easier to write unit tests and mock it in other unit files.

src/__test__/is-printing.spec.ts
haxxmaxx: ah missed that one

haxxmaxx:
> So you mean reporting service needs to send a request to sn-table to ask for font family, font size, padding, etc. to calculate. To avoid to ask many things like that and then do calculation, reporting service only need to ask for the row heights and it should don't care how to calculate the row height. Calculating the row height should be in sn-table.

No not really. My point was mainly that you can _calculate_ it exactly, it understood it that as the plan was to _estimate_ it.

src/handle-data.ts
quanho: Fixed!

src/render-as-pagination.ts
quanho: Good point! Fixed!

src/table/virtualized-table/Table.tsx
quanho: Fixed!

src/table/virtualized-table/Body.tsx
quanho: Just in case `maxLineCount` is set to zero, I want to it is set to the default value which is defined in another file.

src/__test__/is-printing.spec.ts
quanho: Good catch! Fixed!

src/table/pagination-table/components/body/TableBodyWrapper.tsx
quanho: I want to avoid the case when `maxLineCount` is set to 0

quanho: But I guess we can ignore that case.

src/table/virtualized-table/Body.tsx
quanho: Probably we can ignore that case.

src/table/pagination-table/components/TableWrapper.tsx
quanho: Good point! Fixed!

src/__test__/handle-data.spec.ts
haxxmaxx: I think these tests should be one where skipTotals is false and one where it is true

haxxmaxx (approved):

src/table/utils/get-visible-height.ts
```diff
@@ -0,0 +1,19 @@
...
+const getVisibleHeight = (
+    visibleRowEndIndex: number,
+    visibleRowStartIndex: number,
+    layout: TableLayout,
+    viewService: ViewService
+  ) => {
+    if (visibleRowEndIndex < 0) return 0;
+  
+    const totalRowCount = layout.qHyperCube.qSize.qcy;
+    const visualRowsPerPage = viewService.rowsPerPage || initialPageInfo.rowsPerPage;
```
cbt1, line 13: I think you mean `visible`RowsPerPage and not `visual`RowsPerPage right?

quanho: This comes from https://github.com/qlik-oss/sn-table/pull/1035. I just moved the logic to another file to make it easier to write unit tests and mocking.

PR #1062, opened 2023-09-13
fix: easy id fixes

haxxmaxx:
all ID's attributes in the DOM should be unique, this is the first batch of fixing those, more to come


cbt1 (approved):
LGTM 🚢

src/table/components/footer/PaginationContent.tsx
```diff
@@ -124,11 +125,12 @@ const PaginationContent = ({
   ) => {
     const label = translator.get(`SNTable.Pagination.${name}`);
     const id = `${name}-dropdown`;
-    const labelId = `${id}-label`;
+    const labelId = `${id}-label-${chartId}`;
```
niekvanstaveren, line 128: It is possible to render the same chart twice on a page. Not sure if we need to cover that case...

haxxmaxx: If you have a master viz or multiple instances in a mashup?

PR #1063, opened 2023-09-13
feat: support showing Totals row in the first page or last page only

quanho:


src/table/utils/get-view-state.ts
```diff
@@ -6,9 +6,9 @@ import { TableLayout, ViewService } from '../../types';
 
 const getViewState = (layout: TableLayout, viewService: ViewService, rootElement: HTMLElement) => {
     if (viewService.viewState && !viewService.viewState.isMultiPage) return viewService.viewState;
-  
+    let totalsPosition;
```
haxxmaxx, line 9: Can't you just do set the totalsPosition here and use that for both cases?
```suggestion
    const totalsPosition = viewService.viewState?.isMultiPage ? totalsPosition : undefined
```

cbt1: Why does it have to be undefined when it's not a multipage print?

Caele: The value is never used by the table itself, so could just as well have it there at all times.

cbt1: Is it just me that is confussed by `viewState` and what is something more of a "rendering option" versus what is a value that some other party will consume? There seem to be no indication of that in this "API"? Or maybe I have completly miss-understood the viewState..

quanho: I tried to avoid update unit tests since there are unit tests to check if getTotalPosition should be run or not.

quanho: > Is it just me that is confussed by `viewState` and what is something more of a "rendering option" versus what is a value that some other party will consume? There seem to be no indication of that in this "API"? Or maybe I have completly miss-understood the viewState..

For multi-page pdf, viewstate is used to comunicate between sn-table and download service. It is more complicated than any other scenarios which use viewstate before.

quanho: > The value is never used by the table itself, so could just as well have it there at all times.

OK, totalsPosition is always there.

src/table/utils/get-view-state.ts
```diff
@@ -2,18 +2,18 @@ import { findPaginationVisibleRows, findVirtualizedVisibleRows } from './find-vi
...
+import { TableLayout, TotalsPosition, ViewService } from '../../types';
 
 const getViewState = (layout: TableLayout, viewService: ViewService, rootElement: HTMLElement) => {
-    if (viewService.viewState && !viewService.viewState.isMultiPage) return viewService.viewState;
-  
-    if (renderAsPagination(layout, viewService)) {
-      const totalsPosition = getTotalPosition(layout, viewService);
+  if (viewService.viewState && !viewService.viewState.isMultiPage) return viewService.viewState;
+    const isPagination = renderAsPagination(layout, viewService);
+    const totalsPosition = (isPagination || viewService.viewState?.isMultiPage) ? getTotalPosition(layout, viewService) : undefined;
```
cbt1, line 10: Why can't you just always resolve the `totalsPosition` object?

quanho: We don't always need totalsPosition.

haxxmaxx: But there is no harm in getting every time, it is not an expensive calculation. That removes all the added logic, and the need to test it

quanho: Again it is no harm but then I need to update a number of unit tests which checking if getTotalPosition is executed or not. I always try to avoid to update old unit tests as much as possible. But if you insist then getTotalPosition is always executed.

cbt1 (approved):

src/table/utils/get-view-state.ts
```diff
@@ -2,40 +2,40 @@ import { findPaginationVisibleRows, findVirtualizedVisibleRows } from './find-vi
...
-    const { visibleRowStartIndex = -1, visibleRowEndIndex = -1 } = findVirtualizedVisibleRows(rootElement, viewService);
-  
+  if (viewService.viewState && !viewService.viewState.isMultiPage) return viewService.viewState;
+  const totalsPosition = getTotalPosition(layout, viewService);
+  if (renderAsPagination(layout, viewService)) {
+    const {
+      visibleRowStartIndex = -1,
+      visibleRowEndIndex = -1,
+      rowPartialHeight,
+    } = findPaginationVisibleRows(rootElement, totalsPosition as TotalsPosition);
```
haxxmaxx, line 15: shouldn't need to cast here, totalsPosition should be always defined

Caele (approved):

src/table/utils/get-view-state.ts
haxxmaxx: Then I see why you went with that approach. But I think that is an approach that builds technical debt. I think spending some extra minutes updating the tests, to write clean, easy to understand code is worth it. Especially when it is such a small change in the tests

PR #1064, opened 2023-09-14
fix: big table dont render properly

cbt1:
The failure to render a cells has to do with the max element dom size being exceeded. This PR just lowers the max allowed size. Where the consequence of that, is that instead of rendering some rows with 3 lines of text, it may now render them with only 1 or 2 lines.

There is a need to take an in-depth look at this and fix the issue in a far more dynamic way. But that is a much much larger and risky change. So this PR takes the easy path that may not work for every browser and machine combination.

big-table-render-error


haxxmaxx (approved):
Since this was internally reported, I think this fix is fine for now.

PR #1065, opened 2023-09-14
fix: use unique ids for header menu and dropdown

haxxmaxx:
Making sure that the header menu button has a unique id. Also forgot to give the select, not just the label, a unique id

### DIFF
--- a/src/table/components/head/HeadCellMenu.tsx
+++ b/src/table/components/head/HeadCellMenu.tsx
@@ -166,8 +166,8 @@ const HeadCellMenu = ({ column, tabIndex }: HeadCellMenuProps) => {
   }, [openMenuDropdown, resetSelectionActionsEnabledStatus]);
 
   const chartId = layout.qInfo.qId as string;
-  const buttonId = `sn-table-head-menu-button-${pageColIdx}-${chartId}`;
   const menuId = `sn-table-head-menu-${pageColIdx}-${chartId}`;
+  const buttonId = `sn-table-head-menu-button-${pageColIdx}-${chartId}`;
 
   return menuItemGroups.length ? (
     <HeadCellMenuWrapper rightAligned={headTextAlign === 'right'}>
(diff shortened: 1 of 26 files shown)


src/table/components/head/HeadCellMenu.tsx
```diff
@@ -165,15 +165,20 @@ const HeadCellMenu = ({ column, tabIndex }: HeadCellMenuProps) => {
...
         ref={anchorRef}
         size="small"
         tabIndex={tabIndex}
-        id="sn-table-head-menu-button"
-        aria-controls={openMenuDropdown ? 'sn-table-head-menu' : undefined}
+        className ='sn-table-head-menu-button'
+        id={buttonId}
+        aria-controls={openMenuDropdown ? menuId : undefined}
         aria-expanded={openMenuDropdown ? 'true' : undefined}
         aria-haspopup="true"
```
haxxmaxx, line 183: I haven't been able to find an example that uses all three of these. Not sure if it is the best practice. But this is the way the examples in MUIs documentation work. So wont touch this.

src/table/components/head/HeadCellMenu.tsx
```diff
@@ -165,15 +165,20 @@ const HeadCellMenu = ({ column, tabIndex }: HeadCellMenuProps) => {
     if (!openMenuDropdown) resetSelectionActionsEnabledStatus();
   }, [openMenuDropdown, resetSelectionActionsEnabledStatus]);
 
+  const chartId = layout.qInfo.qId as string;
+  const buttonId = `sn-table-head-menu-button-${pageColIdx}-${chartId}`;
+  const menuId = `sn-table-head-menu-${pageColIdx}-${chartId}`;
```
haxxmaxx, line 170: I was thinking about making `sn-table-head-menu-` and the other one into constants, but didn't really have a good place to store them, without creating a circular dep

cbt1:
Will have to sync some of these changes with the HeaderMenu in the util repo. I guess the sooner the better.

cbt1 (approved):
LGTM. I will sync the pagination stuff into the my PR in the util repo

cbt1:
> Will have to sync some of these changes with the HeaderMenu in the util repo. I guess the sooner the better.

👉 @a-m-dev

PR #1068, opened 2023-09-20
chore: bump sprout

haxxmaxx:
removing id:s from icons that are used in the table


eliseeborn (approved):

haxxmaxx:
Not that the color of the arrow of in the select component has changed, hence the updated snapshots. This might be changed back later, but we need to get this bug fix released first

cbt1 (approved):

PR #1079, opened 2023-10-17
refactor: color utils

cbt1:
Replaces the color utils functions with `@qlik/nebula-table-utils` functions


src/table/utils/__tests__/styling-utils.spec.ts
```diff
@@ -242,7 +242,7 @@ describe("styling-utils", () => {
 
       const resultStyling = getHeaderStyle(layout, theme, false);
       expect(resultStyling).toEqual({
-        background: "#323232",
+        background: "rgb(50, 50, 50)",
```
cbt1, line 245: `toRGB` always returns a rgb/rgba string

haxxmaxx (approved):

PR #1083, opened 2023-10-26
fix: add touch listeners for ColumnAdjuster

haxxmaxx:
- adding event listeners for `touchstart`, `touchmove,` `touchend`. Using the same logic just slightly different event structure -> we need separate handle functions


cbt1:
Might want to handle `touchcancel` event as well?

cbt1 (approved):
LGTM, just some minor comments.

Did some quick read-up at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events). Looks like the `touchend` event might be triggered every time a "finger" is lifted. So if the user starts dragging with a single finger, while dragging adds another finger and then lift that up. You might end up in a strange scenario.

PR #1084, opened 2023-11-07
feat: `<HeadCellMenu />` from shared lib

a-m-dev:
This feature will be kept behind FF and improved incrementally

Included in this PR: 
- [x] new FF for all this work (`HEAD_CELL_MENU_TEST_FLAG`) 
- [x] integrate new `HeadCellMenu` component from shared lib
- [x] 👆 same for Virtual table
- [x]  create a `featureFlags` Object on context in order to prevent prop drilling through multiple component to access FF
- [x] tests

Not Included in this PR:
- Background color effect (soon)
- Focus on cell, not only on head Cell button (soon) 
- head cell menu button alignment when head cell label is multilined (soon)
- should we keep label of head cell multilined for sn-table? (status: yet is unclear)

### DIFF
--- a/src/__test__/generate-test-data.ts
+++ b/src/__test__/generate-test-data.ts
@@ -21,7 +21,8 @@ export function generateLayout(
   nMeas: number,
   nRows: number,
   qColumnOrder: number[] = [],
-  qGrandTotalRow: Record<string, string>[] = []
+  qGrandTotalRow: Record<string, string>[] = [],
+  qEffectiveInterColumnSortOrder: number[] = [0]
 ): TableLayout {
   const createField = (idx: number) => ({
     qFallbackTitle: `title-${idx}`,
@@ -55,6 +56,7 @@ export function generateLayout(
       qColumnOrder,
       qSize: { qcx: nDims + nMeas, qcy: nRows },
       qDataPages: [],
+      qEffectiveInterColumnSortOrder,
     } as unknown as HyperCube,
     totals,
     qInfo: {
--- a/src/__test__/generate-test-data.ts
+++ b/src/__test__/generate-test-data.ts
@@ -22,7 +22,7 @@ export function generateLayout(
   nRows: number,
   qColumnOrder: number[] = [],
   qGrandTotalRow: Record<string, string>[] = [],
-  qEffectiveInterColumnSortOrder: number[] = [0]
+  qEffectiveInterColumnSortOrder: number[] = Array.from(Array(nDims + nMeas).keys())
 ): TableLayout {
   const createField = (idx: number) => ({
     qFallbackTitle: `title-${idx}`,
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -94,7 +94,7 @@ export const getBodyCellAlign = (cell: EngineAPI.INxCell, textAlign: Align | "au
  * Gets all column info.
  */
 export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: number, visibleColIdx: number): Column {
-  const { qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
+  const { qDimensionInfo, qMeasureInfo, qEffectiveInterColumnSortOrder } = layout.qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIdx < numDims;
   const info = isDim ? qDimensionInfo[colIdx] : qMeasureInfo[colIdx - numDims];
@@ -112,6 +112,10 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
     selectionColIdx = visibleColIdx;
     ({ qDimensionType } = dimInfo);
   }
+
+  // TODO:
+  // check about hidden columns and why we need to get it from properties
+  const isActivelySorted = colIdx === qEffectiveInterColumnSortOrder[0];
 
   const {
     qFallbackTitle,
@@ -135,6 +139,7 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
     qReverseSort,
     columnWidth,
     selectionColIdx,
+    isActivelySorted,
     id: `col-${pageColIdx}`,
     label: qFallbackTitle,
     stylingIDs: qAttrExprInfo.map((expr) => expr.id),
--- a/src/handle-data.ts
+++ b/src/handle-data.ts
@@ -93,11 +93,18 @@ export const getBodyCellAlign = (cell: EngineAPI.INxCell, textAlign: Align | "au
 /**
  * Gets all column info.
  */
-export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: number, visibleColIdx: number): Column {
-  const { qDimensionInfo, qMeasureInfo, qEffectiveInterColumnSortOrder } = layout.qHyperCube;
+export function getColumnInfo(
+  layout: TableLayout,
+  colIdx: number,
+  pageColIdx: number,
+  visibleColIdx: number,
+  qEffectiveInterColumnSortOrder: number[]
+): Column {
+  const { qDimensionInfo, qMeasureInfo } = layout.qHyperCube;
   const numDims = qDimensionInfo.length;
   const isDim = colIdx < numDims;
   const info = isDim ? qDimensionInfo[colIdx] : qMeasureInfo[colIdx - numDims];
+  const isActivelySorted = colIdx === qEffectiveInterColumnSortOrder[0];
 
   let fieldIndex = 0;
   let fieldId = "";
@@ -112,10 +119,6 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
     selectionColIdx = visibleColIdx;
     ({ qDimensionType } = dimInfo);
   }
-
-  // TODO:
-  // check about hidden columns and why we need to get it from properties
-  const isActivelySorted = colIdx === qEffectiveInterColumnSortOrder[0];
 
   const {
     qFallbackTitle,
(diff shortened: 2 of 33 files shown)


src/index.ts
```diff
@@ -84,7 +86,8 @@ export default function supernova(env: Galaxy) {
       });
 
       usePaginationTable({
-        env,
+        isNewHeadCellMenuEnabled,
+        env, // TODO: this is not used!
```
a-m-dev, line 90: should we remove this?

src/table/utils/accessibility-utils.ts
```diff
@@ -26,9 +26,9 @@ export const setFocusOnClosetColumnAdjuster = (anchorRef: React.RefObject<HTMLDi
 export const focusHeadMenuButton = (event: React.KeyboardEvent | React.FocusEvent<HTMLDivElement>) => {
   const target = event.target as HTMLDivElement;
   target.setAttribute("tabIndex", "-1");
-  const headMenuButton = target
-    ?.closest(".sn-table-cell")
-    ?.querySelector(".sn-table-head-menu-button") as HTMLButtonElement;
+  const baseElement = target?.closest(".sn-table-cell");
+  const headMenuButton = (baseElement?.querySelector(".sn-table-head-menu-button") ||
+    baseElement?.querySelector("#nebula-table-utils-head-menu-button")) as HTMLButtonElement;
```
a-m-dev, line 31: good candidate for exporting this id from shared lib, will keep it for the work that i'll do in there.

Also, `.sn-table-head-menu-button` for old head cell and `#nebula-table-utils-head-menu-button` for new head cell, so based on the flag either of those two will be true to focus

src/handle-data.ts
```diff
@@ -113,6 +113,10 @@ export function getColumnInfo(layout: TableLayout, colIdx: number, pageColIdx: n
     ({ qDimensionType } = dimInfo);
   }
 
+  // TODO:
+  // check about hidden columns and why we need to get it from properties
+  const isActivelySorted = colIdx === qEffectiveInterColumnSortOrder[0];
```
haxxmaxx, line 118: I can't find any usage of this prop

src/__test__/generate-test-data.ts
```diff
@@ -21,7 +21,8 @@ export function generateLayout(
   nMeas: number,
   nRows: number,
   qColumnOrder: number[] = [],
-  qGrandTotalRow: Record<string, string>[] = []
+  qGrandTotalRow: Record<string, string>[] = [],
+  qEffectiveInterColumnSortOrder: number[] = [0]
```
haxxmaxx, line 25: the default here is incorrect if you don't have one column, right? I think if we should have a fallback, then it should be correct

src/index.ts
haxxmaxx: did you remove the usage of it in this PR?

src/nebula-hooks/use-sorting.ts
```diff
@@ -23,8 +28,15 @@ export const sortingFactory = (dimensionsLength: number, model: EngineAPI.IGener
...
-      // Reverse
+    }
+
+    // Reverse
+    if (
+      (isNewHeadCellMenuEnabled &&
+        ((newSortDirection && newSortDirection !== sortDirection) || (!newSortDirection && colIdx === topSortIdx))) ||
+      // if flag is not enabled -> we only want to reverse if colIdx is eql to topSortedIdx
+      (!isNewHeadCellMenuEnabled && colIdx === topSortIdx)
+    ) {
```
haxxmaxx, line 39: I agree with codeclimate here. I'd start with creating the variable `isActivelySorted` in here, and use it instead of `colIdx === topSortIdx`, since you check that three times. Then I would probably break out this into two constants, maybe one for when the flag is on and one when the flag is off

src/table/pagination-table/components/head/TableHeadWrapper.tsx
```diff
@@ -70,7 +75,7 @@ const TableHeadWrapper = () => {
           return (
             <StyledHeadCell
               headerStyle={styling.head}
-              style={widthStyle} // add by style to reduce number of classes created by mui
+              style={{ ...widthStyle, ...(isNewHeadCellMenuEnabled ? { cursor: "pointer" } : {}) }} // add by style to reduce number of classes created by mui
```
haxxmaxx, line 78: the width style is added inline since it update frequently. I think you can pass the flag prop to the styled component and resolve it there

src/table/pagination-table/components/head/TableHeadWrapper.tsx
```diff
@@ -80,7 +85,7 @@ const TableHeadWrapper = () => {
               onKeyDown={handleKeyDown}
               onMouseDown={handleMouseDown}
             >
-              <HeadCellContent column={column} isActive={isActive} isInteractionEnabled={isInteractionEnabled}>
+              <HeadCellContent column={column} isSorted={isSorted} isInteractionEnabled={isInteractionEnabled}>
```
haxxmaxx, line 88: I find sorted a bit misleading, all columns are sorted in some way, but not always perfectly ascending/ascending (i.e. index first in `qEffectiveInterColumnSortOrder`). I think `isActivelySorted` was a good name for it

src/table/components/head/HeadCellContent.tsx
```diff
@@ -1,41 +1,72 @@
...
+
+  const [open, setOpen] = useState(false);
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const listboxRef = useRef<HTMLDivElement>(null);
 
   const { startIcon, endIcon, lockIcon } = getHeadIcons(column);
   const tabIndex = isInteractionEnabled && areTabStopsEnabled(keyboard) ? 0 : -1;
 
-  const handleSort = () => isInteractionEnabled && changeSortOrder(column);
+  const handleSort = () => isInteractionEnabled && changeSortOrder(column, column.sortDirection);
```
haxxmaxx, line 37: why are you adding an argument here? I thought with the flag off, you could still use `changeSortOrder` as before

src/table/utils/accessibility-utils.ts
haxxmaxx: yes that should probably be exported.

But we can't use this ID in the sn-table. Since you get one of these buttons per head cell, you get multiple instances of an element with the same ID. Having unique IDs in the DOM is a a11y requirement. So could we switch that to a class instead? Not a hurry now since we have a bunch of other a11y stuff to fix

src/index.ts
a-m-dev: no, i was just asking if we can remove it because seems like it hasn't been used

src/table/utils/accessibility-utils.ts
a-m-dev: Yes, we (probably) can use class instead, but regarding the multiple instances part, i think that is not correct maybe. 
Because we do a querySelector on `baseElement` which is the closest el with `.sn-table-cell` class name on the triggered event location 
So we cant have multiple elements with same id i guess or am i missing something?

src/index.ts
haxxmaxx: then that is a different PR I would say

src/handle-data.ts
a-m-dev: this is being passed to shared lib as field/col/header prop: https://github.com/search?q=repo%3Aqlik-oss%2Fnebula-table-utils%20isActivelySorted&type=code

and it's being used to enable both sort asc and desc button on actively sorted column

src/table/utils/accessibility-utils.ts
haxxmaxx: the code for getting the correct element here will work. The only reason we need unique IDs is because it is a11y requirement. I recently fixed this for sn-table. (Also, I just googled, it says it has to be unique on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id))

src/__test__/generate-test-data.ts
a-m-dev: if we dont have one column, then then table should not be rendered, right? 
So I think having one column means that table is sorted based on that one column and this default is fine? or am i  wrong in this case?

src/nebula-hooks/use-sorting.ts
a-m-dev: fixed

src/handle-data.ts
a-m-dev: Just as a side note: 
fixed the Todo thing in this commit: https://github.com/qlik-oss/sn-table/pull/1084/commits/9cbd5c0d57c3d3fd5cd00ca3cea5cbe76e1d9ddd

src/table/utils/accessibility-utils.ts
a-m-dev: will scope this out to a11y PR, 
I think it will effect PVT as well

src/handle-data.ts
haxxmaxx: I saw that, nice

src/table/utils/accessibility-utils.ts
haxxmaxx: sounds good

src/nebula-hooks/use-virtualized-table.ts
```diff
@@ -45,17 +46,22 @@ const useVirtualizedTable = ({
...
   isFontLoaded,
   viewService,
+  isNewHeadCellMenuEnabled,
 }: UseVirtualizedTable) => {
   const shouldRender = !renderAsPagination(layout, viewService);
-  const tableData = useMemo(
-    () => getVirtualScrollTableData(layout, interactions, viewService),
-    [layout, interactions, viewService]
-  );
+  const [tableData] = usePromise(async () => {
```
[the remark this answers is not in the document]
a-m-dev: reverted based on discussion on slack

src/__test__/generate-test-data.ts
a-m-dev: no the default value is `Array.from(Array(nDims + nMeas).keys())`

src/nebula-hooks/__tests__/use-sorting.spec.ts
```diff
@@ -6,7 +6,7 @@ describe("use-sorting", () => {
   describe("sortingFactory", () => {
```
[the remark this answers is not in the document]
a-m-dev: added two new tests

haxxmaxx (approved):

PR #1085, opened 2023-11-09
refactor: use ColumnAdjuster from utils repo

haxxmaxx:


haxxmaxx:
slight drop in coverage, but there is no logic to test really

src/ext/property-panel/data.js
```diff
@@ -208,12 +208,12 @@ const getColumnResize = {
...
     expression: "optional",
     defaultValue: 20,
-    show: (data) => data.qDef.columnWidth?.type === ColumnWidthTypes.PERCENTAGE,
+    show: (data) => data.qDef.columnWidth?.type === ColumnWidthType.Percentage,
     change(data) {
       data.qDef.columnWidth.percentage =
         data.qDef.columnWidth.percentage === undefined
           ? data.qDef.columnWidth.percentage
-          : Math.max(1, Math.min(MAX_COLUMN_PERCENTAGE_WIDTH, data.qDef.columnWidth.percentage));
+          : Math.max(1, Math.min(ColumnWidthValues.PixelsMax, data.qDef.columnWidth.percentage));
```
cbt1, line 216: `ColumnWidthValues.PixelsMax`?

package.json
```diff
@@ -114,7 +114,7 @@
     "@playwright/test": "1.39.0",
     "@qlik-trial/sprout": "3.7.1",
     "@qlik/eslint-config": "0.4.30",
-    "@qlik/nebula-table-utils": "^2.6.1",
+    "@qlik/nebula-table-utils": "2.6.1",
```
a-m-dev, line 117: I noticed this one as well, and addressed in my PR too

a-m-dev (approved):
🚀

package.json
a-m-dev: this one is in now

src/ext/property-panel/data.js
haxxmaxx: that is indeed wrong, I think I need to test this a bit more in the client setting values using the pp.

PR #1091, opened 2023-11-16
chore: remove redundent deps

cbt1:
The dependecies below where overriding the same deps coming from `@qlik/eslint-config` which are at v6. This PR removes those deps from sn-table and lets `@qlik/eslint-config` decide the version. The result is that `any` type is not allowed any more.

```json
"@typescript-eslint/eslint-plugin": "5.62.0",
"@typescript-eslint/parser": "5.62.0",
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
         }
       ],
       "@typescript-eslint/no-floating-promises": 0,
+      "@typescript-eslint/no-unsafe-enum-comparison": 0,
       "@typescript-eslint/no-unused-expressions": [
         1,
         {
@@ -125,8 +126,6 @@
     "@types/jest": "29.5.8",
     "@types/react": "18.2.37",
     "@types/react-window": "1.8.8",
-    "@typescript-eslint/eslint-plugin": "5.62.0",
-    "@typescript-eslint/parser": "5.62.0",
     "conventional-changelog-cli": "4.1.0",
     "eslint": "8.53.0",
     "eslint-plugin-jest": "27.6.0",
--- a/src/ext/conversion/import-properties.ts
+++ b/src/ext/conversion/import-properties.ts
@@ -57,6 +57,7 @@ export const getMultiColumnInfo = (
 const importProperties = (
   exportFormat: ExportFormat,
   initialProperties: EngineAPI.IGenericHyperCubeProperties,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   extension: any,
   hypercubePath?: string,
 ): PropTree => {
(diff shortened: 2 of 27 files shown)


package.json
```diff
@@ -77,6 +77,7 @@
         }
       ],
       "@typescript-eslint/no-floating-promises": 0,
+      "@typescript-eslint/no-unsafe-enum-comparison": 0,
```
cbt1, line 80: I disable this rule since there is a lot of comparisons with enums `"string" === SomeEnum.Attr` which would have had to be converted into `"string" === SomeEnum.Attr.toString()`.

src/ext/conversion/import-properties.ts
```diff
@@ -57,6 +57,7 @@ export const getMultiColumnInfo = (
 const importProperties = (
   exportFormat: ExportFormat,
   initialProperties: EngineAPI.IGenericHyperCubeProperties,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
cbt1, line 60: Used `any` here for simplicity. As `extension` is coming from Sense.

src/table/context/createSelectorProvider.tsx
```diff
@@ -9,6 +9,7 @@ import React, {
   useRef,
 } from "react";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
cbt1, line 12: Could not come up with a better type solution so had to stick with `any`

haxxmaxx:
So is the intention here to just make eslint work properly? I feel like some types that go from any to unkown could actually be properly typed

cbt1:
> So is the intention here to just make eslint work properly? I feel like some types that go from any to unkown could actually be properly typed

Yes. The "wrong" version of lint rules where used. Which allow `any`. Now that has changed. Fixing all bad types is out of scope for this PR.

haxxmaxx:
> > So is the intention here to just make eslint work properly? I feel like some types that go from any to unkown could actually be properly typed
> 
> Yes. The "wrong" version of lint rules where used. Which allow `any`. Now that has changed. Fixing all bad types is out of scope for this PR.

Ok that makes sense. We should probably have a broader look at some point. The keyboard and mouse events are all over the place. I think if we specify them better we can get rid of all the event target casting (`evt.target as HTMLDivElement` etc)

src/table/context/createSelectorProvider.tsx
haxxmaxx: yupp this is super general, so it is basically any

src/extend-context-menu.ts
```diff
@@ -3,9 +3,9 @@ import { onContextMenu } from "@nebula.js/stardust";
 import copyCellValue from "./table/utils/copy-utils";
 
 export default function extendContextMenu() {
-  onContextMenu?.((menu: any, event: any) => {
+  onContextMenu?.((menu: { addItem: (i: unknown) => void }, event: React.MouseEvent<Element>) => {
```
haxxmaxx, line 6: This can be a keyboard event as well

PR #1092, opened 2023-11-16
feat: new `<HeadCellMenu />` bg color on hover/active state

a-m-dev:
including in this PR: 
- [x] for Pagination
- [x] for VT
- [x] put everything behind flag
- [x] unit tests

**Before:**
Screenshot 2023-11-20 at 16 15 31

**After:**

https://github.com/qlik-oss/sn-table/assets/29652890/8535e476-e816-467e-8d91-7168c4ee9854


src/table/pagination-table/components/head/styles.ts
```diff
@@ -4,14 +4,24 @@ import { COMMON_CELL_STYLING } from "../../../styling-defaults";
...
 // eslint-disable-next-line import/prefer-default-export
 export const StyledHeadCell = styled(TableCell, {
-  shouldForwardProp: (prop: string) => !["headerStyle", "isNewHeadCellMenuEnabled"].includes(prop),
-})(({ headerStyle, isNewHeadCellMenuEnabled }) => ({
+  shouldForwardProp: (prop: string) =>
+    prop !== "headerStyle" &&
+    prop !== "isNewHeadCellMenuEnabled" &&
+    prop !== "interactions" &&
+    prop !== "hoverBackground" &&
+    prop !== "background",
```
cbt1, line 12: Looks like `background` is only used with the feature flag. Was it not used before? So without the flag it should have not background?

src/table/utils/styling-utils.ts
```diff
@@ -1,16 +1,30 @@
...
 import { SELECTION_STYLING } from "../styling-defaults";
-import { CellStyle, GeneratedStyling } from "../types";
+import { CellStyle, FeatureFlags, GeneratedStyling } from "../types";
 
 export const LINE_HEIGHT = 4 / 3;
 export const CELL_PADDING_HEIGHT = 8;
 export const CELL_BORDER_HEIGHT = 1;
 
+// TODO: maybe shared repo?
+const HEADER_MENU_COLOR_MODIFIER = {
```
cbt1, line 15: If they use the same modifiers maybe that should be the default?

a-m-dev: Yes, it can be default, but we also need to have a way to override them, just in case if we want it to

src/table/pagination-table/components/head/styles.ts
a-m-dev: headerStyle has background property on it: https://github.com/qlik-oss/sn-table/blob/d1cf06c376ded408c730b2a05ab99fe4091dc334/src/table/utils/styling-utils.ts#L130

so the active and hover background is derived from that background color ([in styling utils here](https://github.com/qlik-oss/sn-table/blob/44ab786649a287bd3874108a1637dc663b39f9ca/src/table/utils/styling-utils.ts#L145-L156))

src/table/context/TableContext.tsx
```diff
@@ -42,18 +42,18 @@ export const TableContextProvider = ({
...
   const [selectionState, selectionDispatch] = useSelectionReducer(tableData.rows, selectionsAPI);
   const [hoverIndex, setHoverIndex] = useState(-1);
-  const styling = useTableStyling(layout, theme, tableData, rootElement);
+  const styling = useTableStyling(layout, theme, tableData, rootElement, featureFlags);
   const [columnWidths, setColumnWidths, setYScrollbarWidth, showRightBorder] = useColumnWidths(
     tableData.columns,
     tableData.totalsPosition,
     rect.width,
-    styling,
+    styling
```
haxxmaxx, line 55: why the remove of trailing commas? lint rule change?

haxxmaxx (approved):

src/table/components/head/styles.ts
```diff
@@ -17,7 +17,7 @@ export const AbsolutelyStyledRefAnchor = styled("div")({
 
 export const StyledSortButton = styled(Button, {
   shouldForwardProp: (prop: string) =>
-    !["isActivelySorted", "textAlign", "disabled", "isNewHeadCellMenuEnabled"].includes(prop),
+    prop !== "isActivelySorted" && prop !== "textAlign" && prop !== "disabled" && prop !== "isNewHeadCellMenuEnabled",
```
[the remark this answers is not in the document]
a-m-dev: just experimenting something with ts there has been a type issue with `shouldForwardProp` complaining that no overload matching the `props: string` part when running types:check command. but CI is fine with that for now!

will changed back this line to the prev arr version of it.

src/table/context/TableContext.tsx
a-m-dev: I think so, I just pilled the latest main and branched off based on that, my take is that probably when i did a yarn install -> it picked up some new versions of qlik based lint settings and those were updated

cbt1 (approved):
LGTM

src/table/components/head/styles.ts
```diff
@@ -70,7 +70,7 @@ export const StyledHeadCellContent = styled(Box, {
   gap: theme.spacing(0.5),
   fontSize: "inherit",
   fontFamily: "inherit",
-  ...(isNewHeadCellMenuEnabled && { alignItems: "end" }),
+  ...(isNewHeadCellMenuEnabled && { alignItems: "end", overflow: "hidden" }),
```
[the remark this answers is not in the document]
haxxmaxx: great

haxxmaxx (approved):

PR #1095, opened 2023-11-21
feat: import view data properties

niekvanstaveren:


src/qae/index.ts
```diff
@@ -0,0 +1,10 @@
+import { exportProperties, importProperties } from "../conversion";
+import data from "./data";
+import properties from "./object-properties";
+
+export default () => ({
```
haxxmaxx, line 5: does this need to be a function?

src/conversion/import-properties.ts
```diff
@@ -81,6 +89,11 @@ const importProperties = (
     setValue(propertyTree, "qProperty.qHyperCubeDef.qMeasures", measures);
   }
 
+  if (viewDataMode) {
+    setValue(propertyTree, "qProperty.totals.show", false);
```
haxxmaxx, line 93: if `totals: {}` doesn't exist, is it created?

src/ext/index.js
```diff
@@ -34,7 +34,15 @@ export default function ext(env) {
...
       cssScaling: true,
     },
-    importProperties,
-    exportProperties,
+    importProperties(exportFormat, initialProperties, extension, hypercubePath) {
+      const defaultPropertyValues = {
+        defaultDimension: extension.getDefaultDimensionProperties(),
+        defaultMeasure: extension.getDefaultMeasureProperties(),
+      };
+      return importProperties({ exportFormat, initialProperties, extension, hypercubePath, defaultPropertyValues });
```
haxxmaxx, line 42: I'm looking at the `importProperties` call in the `convertTo` method in nebula. The parameters don't really match. Is it a different `importProperties` being called in nebula?

src/conversion/import-properties.ts
niekvanstaveren: I would assume so yes. https://github.com/qlik-trial/conversion/blob/6f6141c69506e2c2afdccf8497471679d686fea4/src/utils/index.js#L43-L64

src/qae/index.ts
niekvanstaveren: Nope :)

src/conversion/import-properties.ts
haxxmaxx: great

haxxmaxx (approved):

PR #1097, opened 2023-11-28
fix: miss while trying to find col adjuster hit area

a-m-dev:
when trying to focus on col adjuster (using keyboard or hitting on the col adjuster menu item) it fails to find the correct element (the adjuster blue needle)

this is due to a class name change we had a while ago, so this PR uses that exact class name and fixes the issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,8 @@
     "root": true
   },
   "resolutions": {
-    "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest"
+    "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest",
+    "rollup": "npm:rollup@3.29.4"
   },
   "devDependencies": {
     "@babel/core": "7.23.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -90,7 +90,8 @@
     "root": true
   },
   "resolutions": {
-    "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest"
+    "@mui/styled-engine": "npm:@mui/styled-engine-sc@latest",
+    "rollup": "npm:rollup@3.29.4"
```
a-m-dev, line 94: Do we need this? (for watcher, the nebula issue that it does not picks changes while developing)
[PVT is currently using it](https://github.com/qlik-oss/sn-pivot-table/blob/177d3991d0e4bcbd1d68d8df596f4f405b757af0/package.json#L83-L85)

haxxmaxx (approved):

PR #1098, opened 2023-11-30
chore: add prettier check

niekvanstaveren:
Found that there was a bunch of files not properly formatted. Every time I did a commit these files were formatted (after the commit was done). 
Added the formatting changes here and also add a formatting check in CI to make sure we do not get a corrupt state again


src/table/components/head/HeadCellContent.tsx
```diff
@@ -19,12 +19,12 @@ import {
 const HeadCellContent = ({ children, column, isInteractionEnabled, open, setOpen }: HeadCellContentProps) => {
   const { keyboard, translator, changeSortOrder, interactions, embed, app, model } = useContextSelector(
     TableContext,
-    (value) => value.baseProps
+    (value) => value.baseProps,
```
cbt1, line 22: This is the opposite of what @a-m-dev have been doing in his PR:s. I think one of you have the wrong version of prettier or the prettier config package

PR #1107, opened 2023-12-08
fix: correct focus on table when keyboard navigation and NewHeadCell is turned on

a-m-dev:
When keyboard navigation is on, specially when the new head cell flag is also true, we had some in correct behaviour when it comes to the focus position on head and body cells. this pr aims to fix it.

This PR testes in scenarios like: 
- [x] selection mode
- [x] exiting/entring table
- [x] pressing esc when focus is on a body cell

and all of them working correctly now and sticking to the rule of only having one tab stop in table when new head cell menu is turned on.


src/table/pagination-table/components/head/HeadCell.tsx
```diff
@@ -74,7 +76,13 @@ const HeadCell = ({ column, columnIndex, columnsLength }: HeadCellProps) => {
...
 
   const newHeadCellRelatedLogic = isNewHeadCellMenuEnabled ? column.colIdx === 0 : false;
-  const tabIndex = newHeadCellRelatedLogic && isInteractionEnabled && areTabStopsEnabled(keyboard) ? 0 : -1;
+  const tabIndex =
+    newHeadCellRelatedLogic &&
+    isFirstHeaderCell(focusedCellCoord) &&
+    isInteractionEnabled &&
+    areTabStopsEnabled(keyboard)
+      ? 0
+      : -1;
```
haxxmaxx, line 85: I thought about this for a bit. I think it should be slightly different. So for the different keyboardNavigation cases:
- false: there should be a tabindex = 0 where `focusedCellCoord` is. So we could basically check if `columnIndex` is the same as in `focusedCellCoord`, then add that tabstop
- true: when you press enter on the chart, you should focus the cell that has the `focusedCellCoord` coordinate. On the first render this is [0, 0]. But in that case, we shouldn't add that tabstop, until `keyboard.active` becomes true

So you should do this instead
```suggestion
  const tabIndex =
    newHeadCellRelatedLogic &&
    !keyboard.enabled
    isThisCellFocused && // see comment above
    isInteractionEnabled &&
      ? 0
      : -1;
```

src/table/pagination-table/components/head/HeadCell.tsx
```diff
@@ -33,9 +34,10 @@ const HeadCell = ({ column, columnIndex, columnsLength }: HeadCellProps) => {
   const columnWidths = useContextSelector(TableContext, (value) => value.columnWidths);
   const isSelectionMode = useContextSelector(TableContext, (value) => value.baseProps.selectionsAPI?.isModal());
   const setFocusedCellCoord = useContextSelector(TableContext, (value) => value.setFocusedCellCoord);
+  const focusedCellCoord = useContextSelector(TableContext, (value) => value.focusedCellCoord);
```
haxxmaxx, line 37: This we don't want to do in this way. `focusedCellCoord` updates eveytime you move the focus -> every head cell will rerender. What you could do is to check if the current cell is focused, so you actually get a value that is a bool:
```suggestion
const isThisCellFocused = useContextSelector(
    TableContext,
    (value) => value.focusedCellCoord[0] === 0 && value.focusedCellCoord[1] === columnIndex,
  );
```
Of course you can break out this logic if you want

src/table/utils/keyboard-utils.ts
```diff
@@ -15,6 +15,9 @@ export const isArrowKey = (key: string) =>
 
 export const isShiftArrow = (evt: React.KeyboardEvent) => evt.shiftKey && isArrowKey(evt.key);
 
+export const isFirstHeaderCell = (coord: [number, number]) =>
+  coord[0] === FIRST_HEADER_CELL_COORD[0] && coord[1] === FIRST_HEADER_CELL_COORD[1];
+
```
haxxmaxx, line 20: This type of function should live in accessibility-utils

src/table/utils/keyboard-utils.ts
```diff
@@ -140,6 +143,7 @@ interface BodyTabHelperProps {
   keyboard?: stardust.Keyboard;
   isSelectionMode?: boolean;
   paginationNeeded?: boolean;
+  isNewHeadCellMenuEnabled?: boolean;
 }
 
 /**
```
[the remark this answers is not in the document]
a-m-dev: fixing it when we remove the flag seems god idea, since everything is hidden for now, I'll add it in my list

src/table/utils/keyboard-utils.ts
a-m-dev: with new changes, this function is not needed

haxxmaxx:
I found another issue. If you don't have totals on top, you can get an extra tab stop on the body. see here https://github.com/qlik-oss/sn-table/blob/htm/fix/new-headcell-keyboard-navigation/src/table/pagination-table/components/body/TableBodyWrapper.tsx#L67 and here https://github.com/qlik-oss/sn-table/blob/htm/fix/new-headcell-keyboard-navigation/src/table/pagination-table/components/body/TableTotals.tsx#L31

a-m-dev:
> I found another issue. If you don't have totals on top, you can get an extra tab stop on the body. see here https://github.com/qlik-oss/sn-table/blob/htm/fix/new-headcell-keyboard-navigation/src/table/pagination-table/components/body/TableBodyWrapper.tsx#L67 and here https://github.com/qlik-oss/sn-table/blob/htm/fix/new-headcell-keyboard-navigation/src/table/pagination-table/components/body/TableTotals.tsx#L31

Based on this ^ I tried to test it with two scenarios: 
1. totals at bottom
2. no totals at all

in both cases, i got only 1 tab stop per table, my emphasis was around the two components you mentioned (TableBodyWrapper and TableTotals) but no double tab stops at all,  can you please attach a screen recording from the case you get two tab stops? because i cant reproduce the bug tbh

haxxmaxx (approved):

PR #1109, opened 2023-12-13
fix: use capital Percentage while accessing enum value

a-m-dev:
when clicking on percentage column size type, table breaks in sence-client, we noticed that the way we access the enum value is incorrect so this pr fixes it.

also while testing it through sence-client, noticed that the `start:mfe` script in pkg.json was not presented, so feel free to complain about it and tell me to remove it if you think it's not needed

Before fix:

https://github.com/qlik-oss/sn-table/assets/29652890/aaa7166f-bee3-4211-b2f6-75e1f3fbe279

After Fix: 

https://github.com/qlik-oss/sn-table/assets/29652890/f458adfe-6f06-441a-9500-f983d0e13634

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "sense": "nebula sense",
     "spec": "npx @scriptappy/cli from-jsdoc -c ./spec-configs/props.conf.js",
     "start": "nebula serve --build false --type sn-table",
+    "start:mfe": "nebula serve --mfe --build false --type sn-table",
     "test:integration": "playwright test --config=test/integration/integration.config.ts",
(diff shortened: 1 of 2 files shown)


cbt1 (approved):

package.json
```diff
@@ -39,6 +39,7 @@
     "sense": "nebula sense",
     "spec": "npx @scriptappy/cli from-jsdoc -c ./spec-configs/props.conf.js",
     "start": "nebula serve --build false --type sn-table",
+    "start:mfe": "nebula serve --mfe --build false --type sn-table",
```
cbt1, line 42: ❤️‍🔥

PR #1114, opened 2024-01-03
chore: add max lint warnings

niekvanstaveren:
Adding max lint warnings in here. Would be nice to keep this to a minimum

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "deduplicate-packages": "yarn-deduplicate yarn.lock",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "deduplicate-packages": "yarn-deduplicate yarn.lock",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
-    "lint": "eslint --cache .",
+    "lint": "eslint --cache --max-warnings=49 .",
```
haxxmaxx, line 34: why 49? Is that the current number?

niekvanstaveren: Yes that is the current indeed

PR #1116, opened 2024-01-05
refactor: align event types

haxxmaxx:
Previously we have been very specific about how keyboard and mouse events are typed. For example, a type could be `React.MouseEvent<HTMLLIElement>`, specifying the target type using generics. This was mainly an attempt to get `target` and `currentTarget` typed properly, so that they have the methods we use (`focus`, `blur` etc). This did not work, since the type the targets get is hardcoded. The definition for the `React.KeyboardEvent` looks like this
```
    KeyboardEvent<T = Element> extends UIEvent<T, NativeKeyboardEvent>
```
which is based on
```
    interface SyntheticEvent<T = Element, E = Event> extends BaseSyntheticEvent<E, EventTarget & T, EventTarget> {}
```

and BaseSyntheticEvent sets the types of its targets like this
``` 
    interface BaseSyntheticEvent<E = object, C = any, T = any> {
        nativeEvent: E;
        currentTarget: C;
        target: T;
```

So `target` would always be `EventTarget` and `currentTarget` will be `EventTarget & T`. `EventTarget` does not have the properties that we need, so it has to be casted anyway. **So now I have changed all event types to just `React.KeyboardEvent` or `React.MouseEvent`**

Additionally, we don't need to be specific with the types of the elements, so I have changed all of them to `HTMLElement`. (With the exception of `ref`s, they seem to need the specific type of element.) This should simplify passing elements around for the a11y logic, since they all should be `HTMLElement`

I was able to type the return of some methods using generics, like `querySelector` and `closest`. But there are still some places where we need to cast to `HTMLElement`.

Also:
- added a type for `FocusedCellCoord` and its setter
- rename `setFocusOnClosetColumnAdjuster` to `setFocusOnColumnAdjuster` (needs new version of table utils package

### DIFF
--- a/src/extend-context-menu.ts
+++ b/src/extend-context-menu.ts
@@ -3,9 +3,9 @@ import { onContextMenu } from "@nebula.js/stardust";
 import copyCellValue from "./table/utils/copy-utils";
 
 export default function extendContextMenu() {
-  onContextMenu?.((menu: any, event: any) => {
-    event.target &&
-      event.target.closest(".sn-table-cell") &&
+  onContextMenu?.((menu: any, event: Event) => {
+    const target = event.target as HTMLElement;
+    target?.closest(".sn-table-cell") &&
       menu.addItem({
         translation: "contextMenu.copyCellValue",
         icon: "lui-icon lui-icon--copy",
--- a/src/extend-context-menu.ts
+++ b/src/extend-context-menu.ts
@@ -4,7 +4,7 @@ import copyCellValue from "./table/utils/copy-utils";
 
 export default function extendContextMenu() {
   onContextMenu?.((menu: any, event: Event) => {
-    const target = event.target as HTMLElement;
+    const target = event.target as HTMLElement | null;
     target?.closest(".sn-table-cell") &&
       menu.addItem({
         translation: "contextMenu.copyCellValue",
--- a/src/table/components/head/HeadCellContent.tsx
+++ b/src/table/components/head/HeadCellContent.tsx
@@ -4,7 +4,7 @@ import HeadCellMenu, { MenuAvailabilityFlags } from "@qlik/nebula-table-utils/li
 import { SortDirection } from "../../../types";
 import { TableContext, useContextSelector } from "../../context";
 import { HeadCellContentProps } from "../../types";
-import { areTabStopsEnabled, setFocusOnColumnAdjuster } from "../../utils/accessibility-utils";
+import { areTabStopsEnabled, setFocusOnClosetColumnAdjuster } from "../../utils/accessibility-utils";
 import getHeadIcons from "../../utils/get-head-icons";
 import { handleHeadCellMenuKeyDown } from "../../utils/handle-keyboard";
 import OldHeadCellMenu from "./HeadCellMenu";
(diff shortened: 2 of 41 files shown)


src/extend-context-menu.ts
```diff
@@ -3,9 +3,9 @@ import { onContextMenu } from "@nebula.js/stardust";
 import copyCellValue from "./table/utils/copy-utils";
 
 export default function extendContextMenu() {
-  onContextMenu?.((menu: any, event: any) => {
-    event.target &&
-      event.target.closest(".sn-table-cell") &&
+  onContextMenu?.((menu: any, event: Event) => {
```
haxxmaxx, line 6: Pretty sure this is some kind of event, could be both keyboard and mouse. So using a general type here

src/table/components/head/HeadCellContent.tsx
```diff
@@ -94,7 +94,7 @@ const HeadCellContent = ({ children, column, isInteractionEnabled, open, setOpen
               sortRelatedArgs={{ sortFromMenu }}
               searchRelatedArgs={{ embed, listboxRef }}
               selectionRelatedArgs={{ app, model }}
-              adjustHeaderSizeRelatedArgs={{ setFocusOnClosetHeaderAdjuster: setFocusOnClosetColumnAdjuster }}
+              adjustHeaderSizeRelatedArgs={{ setFocusOnColumnAdjuster }}
```
haxxmaxx, line 97: This was misspelled and a bit misleading, so renamed

cbt1:
Looks like we got a breaking change from this PR https://github.com/qlik-oss/nebula-table-utils/pull/72 and it for some reason did not get the correct version when making the release

src/extend-context-menu.ts
```diff
@@ -3,9 +3,9 @@ import { onContextMenu } from "@nebula.js/stardust";
...
+    const target = event.target as HTMLElement;
```
cbt1, line 7: This should be casted to `HTMLElement | undefined` since you need to use the optional chain operator below

cbt1 (approved):
LTGM, just a minor comment.

It's probably better to fix the useMeasureText type error in another PR

haxxmaxx:
> Looks like we got a breaking change from this PR [qlik-oss/nebula-table-utils#72](https://github.com/qlik-oss/nebula-table-utils/pull/72) and it for some reason did not get the correct version when making the release
> 
> EDIT: By the looks of it the "Breaking change" annotation was not added when merging the PR

Yeah my commit had a breaking change too. Something must have gone wrong. So might as well fix the usage of useMeasureText in this PR

haxxmaxx:
@cbt1 actually I will break out the utils bump from this PR

cbt1:
You waiting for something? @haxxmaxx

src/extend-context-menu.ts
```diff
@@ -4,7 +4,7 @@ import copyCellValue from "./table/utils/copy-utils";
 
 export default function extendContextMenu() {
   onContextMenu?.((menu: any, event: Event) => {
-    const target = event.target as HTMLElement;
+    const target = event.target as HTMLElement | null;
```
cbt1, line 7: Since this is called from external code (Sense?), it does seem like really good idea to guard against missing target here. Might even want to do `event?.target...`

PR #1118, opened 2024-01-10
chore: use new version of nebula-table-utils

haxxmaxx:
- use new prop for header menu
- new props structure for useMeasreText

### DIFF
--- a/src/table/hooks/__tests__/use-column-widths.spec.ts
+++ b/src/table/hooks/__tests__/use-column-widths.spec.ts
@@ -13,7 +13,9 @@ import useColumnWidths from "../use-column-widths";
 jest.mock("@qlik/nebula-table-utils/lib/hooks");
 
 describe("use-column-widths", () => {
-  let mockedUseMeasureText: jest.MockedFunction<(props: UseMeasureTextProps) => MeasureTextHook>;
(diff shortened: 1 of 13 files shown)


src/table/hooks/__tests__/use-column-widths.spec.ts
```diff
@@ -13,7 +13,9 @@ import useColumnWidths from "../use-column-widths";
 jest.mock("@qlik/nebula-table-utils/lib/hooks");
 
 describe("use-column-widths", () => {
-  let mockedUseMeasureText: jest.MockedFunction<(props: UseMeasureTextProps) => MeasureTextHook>;
+  let mockedUseMeasureText: jest.MockedFunction<
+    (props: UseMeasureTextProps, options?: { maxNbrLinesOfText?: number }) => MeasureTextHook
+  >;
```
haxxmaxx, line 18: @cbt1 I guess we could export smarter types here, I'd say `UseMeasureTextProps` isn't really correct either

cbt1: Hmm not sure, what is wrong with `UseMeasureTextProps`?

cbt1: Ah yes the naming could probably be improved

PR #1122, opened 2024-01-15
fix: column min width

cbt1:
The old column min width was used even when the new header flag was enabled.


a-m-dev (approved):

test/rendering/utils/playwright.ts
```diff
@@ -4,7 +4,7 @@ const playwright = (page: Page) => ({
   async open(url: string) {
     // eslint-disable-next-line playwright/no-networkidle
     await page.goto(url, { waitUntil: "networkidle" });
-    await page.waitForSelector('[data-testid="sn-table"] .sn-table-cell-text');
+    await page.waitForSelector('[data-testid="sn-table"]');
```
cbt1, line 7: The selector return a first element with a width of 0.`waitForSelector` would fail because it's think the element is not visible.

PR #1123, opened 2024-01-15
fix: prevent click when resizing columns

a-m-dev:
This pr prevents dropdown from opening, when user finishes resizing the column.  
Also this pr is based on @haxxmaxx's work in here: https://github.com/qlik-oss/sn-pivot-table/pull/460 (so thank him on that work 👍)

Also possibly it's better to keep these logic in shared repo, but not sure how much work it would've need, so please let me know if we need to move all this into shared repo and then consume it from there. 

### This PR includes:
- [x] changes for pagination table
- [x] changes for Virtual table 
- [x] required unit tests 

### Before: 

https://github.com/qlik-oss/sn-table/assets/29652890/d2c3c900-baf7-44e6-9227-53a5f38ae14e

### After:

https://github.com/qlik-oss/sn-table/assets/29652890/6b96c6b2-56f0-4357-a391-472d6fc9842d


cbt1 (approved):
LGTM. I think we can work on getting some shared logic out of this but not a top prio atm I'd say

haxxmaxx:
There was a corner case that didn't work for the pivot table. I think I noted it in the task, might want to have a look and see if you can test that. I can't remember what it was though

a-m-dev:
The Jira ticket was mentioning about, this behaviour, 
```
Still possible to make trigger event on the cells that can be resezed. When you resize and the mouse up event is on the cell, not the columnadjuster, it is not blocked properly. Especially easy when you hit the min size for the column

```

which is not happening in case of this PR, but there was one comment regarding this behaviour: 
```
There is still a corner case when this happens. If the column has the min size and you try to make it smaller. Then adjuster detects no change, and the state to block other events on the header cell, is reset -> you get a selection. Not that big of a problem since you didn't do any resize.
```


which i tested it, and seems like it happens, but there are 2 scenarios to this: 
1. if user leaves the mouse within the head cell area -> it will open
2. if user leaves the mouse outside of it -> it will not open


https://github.com/qlik-oss/sn-table/assets/29652890/cf66ff95-7ba9-4240-95eb-f9fcf558f76a

So one catch could be that maybe it's not happening because of the jump effect we saw in the screenshot and for that, i tested it with the most left column (because it has not place to go in left) and it was not the case

https://github.com/qlik-oss/sn-table/assets/29652890/c3b7f615-b491-4ee3-bffe-598bfc7eb75a


I would say the behaviour seems to be fine, unless we decide against it.

haxxmaxx (approved):

PR #1124, opened 2024-01-17
chore: use pnpm

cbt1:


.prettierignore
```diff
@@ -1,7 +1,17 @@
 .cache/
-.husky/
 dist/
+core/esm/
 .DS_Store
 *.html
 sn-table-ext
+coverage
+node_modules
+mashup-example
```
cbt1, line 9: This should not be here

jest.config.js
```diff
@@ -1,8 +1,15 @@
+const path = require("path");
+
 module.exports = {
   testEnvironment: "jest-environment-jsdom",
   testRegex: "src/.+\\.(test|spec)\\.[jt]sx?$",
-  setupFilesAfterEnv: ["./jest.setup.spec.js"],
+  setupFilesAfterEnv: ["./jest/setup.ts"],
```
cbt1, line 6: Setup file should be a ts file when running typescript. Otherwise a whole bunch of warnings is logged

package.json
```diff
@@ -127,13 +120,12 @@
     "@testing-library/user-event": "14.5.2",
     "@types/jest": "29.5.11",
     "@types/react": "18.2.47",
+    "@types/react-dom": "18.2.18",
```
cbt1, line 123: Would get warnings when build since it was not available. Since we actually consume the types for it I added the package.

.circleci/config.yml
```diff
@@ -75,9 +72,12 @@ jobs:
       - run:
           name: Install playwright
           command: npx playwright install
+      - run:
+          name: Install pnpm
+          command: npm install -g pnpm
```
a-m-dev, line 77: Just as a suggestion since we are using this in multiple places, probably we can have this and other shared `run` commands within a `common_steps` and extend it in all:

```
common_steps: &common_steps
  - checkout
  - restore_cache: *restore_yarn_cache
  - run:
      name: Install pnpm
      command: npm install -g pnpm


# later in jobs...
integration-tests:
    <<: *defaults
    docker:
      - image: mcr.microsoft.com/playwright:focal
    steps:
      <<: *common_steps
      - attach_workspace:
          at: ~/table
      - run:
          name: Install playwright
          command: npx playwright install

```
but if you think it's unnecessary at this step, feel free to ignore this comment 🙂

cbt1: We are moving to github actions in #1125 so I will not do any work to improve this. Just wanted it to sort of work

package.json
```diff
@@ -48,18 +46,17 @@
...
-    "preversion": "yarn build",
-    "version": "conventional-changelog -p angular -i CHANGELOG.md -s && yarn spec && git add api-specifications CHANGELOG.md"
+    "preversion": "pnpm build",
+    "version": "conventional-changelog -p angular -i CHANGELOG.md -s && pnpm spec && git add api-specifications CHANGELOG.md"
   },
   "lint-staged": {
-    "*.{js,jsx,ts,tsx}": [
-      "bash -c 'yarn lint'",
-      "bash -c 'yarn types:check'"
+    "*.{js,jsx,ts,tsx, yaml}": [
```
a-m-dev, line 53: I think we only have `.yml` files

cbt1: It's meant to catch changes in `pnpm-lock.yaml`. Not ideal but since the pattern here was only using file ends I ended up doing like this

a-m-dev (approved):
LGTM

PR #1125, opened 2024-01-17
chore: pnpm and sn-gh-workflows

niekvanstaveren:


cbt1:
If you're doing this you should take a look at #1124, that also fixes some build errors/warnings and removes redundent packages

cbt1:
I'd even suggest we get #1124 in first. Then you deal with the workflow change?

cbt1:
PTAL #1124

PR #1127, opened 2024-08-07
fix: sanity check column width properly when converting

haxxmaxx:
Makes sure we get valid column width values when converting to a sn-table. There could be invalid values on other charts before


niekvanstaveren (approved):

cbt1:
Wrong repo?

